PR #3, opened 2015-08-13
Implemented TimeZones Interface

omus:
Implemented the full TimeZones interface. There are allot of changes so I'll point form the important changes:

* TimeZones now requires Base.Dates which requires Julia 0.4 pre-release and above
* TZ data compiling reworked due to bugs
* Added tests
* Built TimeZone types are serialized and saved to disk for fast loading
* Implemented `ZonedDateTime` which is the timezone-aware version of `DateTime`
* Implemented calendrical arithmetic for `ZonedDateTime` (see README)
* Updated Travis script and added Coveralls and Codecov (91% coverage)


omus:
Important note: building the package will download TZ data files from the IANA website. The FTP source we get these files is a little flaky and ends up failing downloads once in a while. As a fix I have the build script automatically attempt to retry downloading a failed file up to 3 times. If by chance we can't download a file the build will fail causing Travis to fail.

REQUIRE
```diff
@@ -1,3 +1 @@
-julia 0.3-
-Dates
-
+julia 0.4-
```
quinnj: I think this just needs to be `julia 0.4`, right @IainNZ?

src/TimeZones.jl
```diff
@@ -0,0 +1,62 @@
+module TimeZones
+
+using Base.Dates
+import Base.Dates: days, hour, minute, second, millisecond
+
+export TimeZone, FixedTimeZone, VariableTimeZone, ZonedDateTime,
+    AmbiguousTimeError, NonExistentTimeError, DateTime,
```
quinnj: Don't need `DateTime` here

REQUIRE
IainNZ: If you'd like it to install it right now, Julia 0.4- is correct ("Julia 0.4 prerelease"). `julia 0.4` means `>=0.4.0`. Not really the clearest thing in the world, is it :D

src/TimeZones.jl
```diff
@@ -0,0 +1,62 @@
...
+    tz_path = joinpath(COMPILED_DIR, split(name, "/")...)
+
+    isfile(tz_path) || error("Unknown timezone $name")
+
+    open(tz_path, "r") do fp
+        return deserialize(fp)
+    end
+end
+
+function timezone_names()
```
quinnj: Maybe just call this `timezones`?

src/timezones/io.jl
```diff
@@ -0,0 +1,98 @@
...
+    return "$local_dt$hh:$mm$(ss)"
+end
+Base.show(io::IO,dt::ZonedDateTime) = print(io,string(dt))
+
+
+# DateTime Parsing
+
+# Note: Ideally Base.Dates.Slot definition should just be: "abstract Slot{P<:Any}" which
+# would allow for custom Slots to be of any type.
+immutable TimeZoneSlot <: AbstractTime end
```
quinnj: Yeah, we're really due for another iteration on the Base.Dates io.jl machinery. There are a few open issues about parsing issues and I've just been putting off really digging in (because parsing isn't my favorite activity!). If you have ideas/code here, it'd be much appreciated in Base.

src/timezones/time.jl
```diff
@@ -0,0 +1,76 @@
...
+    return h, m, s
+end
+
+Base.convert(::Type{Second}, t::Time) = Second(toseconds(t))
+Base.convert(::Type{Millisecond}, t::Time) = Millisecond(toseconds(t) * 1000)
+Base.promote_rule{P<:Union{Week,Day,Hour,Minute,Second}}(::Type{P}, ::Type{Time}) = Second
+Base.promote_rule(::Type{Millisecond}, ::Type{Time}) = Millisecond
+
+# Should be defined in Base.Dates
+Base.isless(x::Period, y::Period) = isless(promote(x,y)...)
```
quinnj: Yeah, I think there was trepidation about defining `promote_rule` for Period types because of `Year`/`Month`. I always figured we could define it later with better definition/bounds on what was legal and what wasn't.

src/timezones/types.jl
```diff
@@ -0,0 +1,212 @@
...
+using Base.Dates
+import Base.Dates: value
+
+# Note: The Olsen Database rounds offset precision to the nearest second
+# See "America/New_York" notes for an example.
+
+# Using type Symbol instead of AbstractString for name since it
+# gets us ==, and hash for free.
+immutable FixedTimeZone <: TimeZone
+    name::Symbol
```
quinnj: +1, good use of Symbol

src/timezones/types.jl
```diff
@@ -0,0 +1,212 @@
...
+    utc_offset::Second  # Standard offset from UTC
+    dst_offset::Second  # Addition offset applied to UTC offset
+
+    function FixedTimeZone(name::Symbol, utc_offset::Second, dst_offset::Second=Second(0))
+        new(name, utc_offset, dst_offset)
+    end
+end
+
+function FixedTimeZone(name::String, utc_offset::Int, dst_offset::Int=0)
+    FixedTimeZone(symbol(name), Second(utc_offset), Second(dst_offset))
```
quinnj: Do we want to validate that `name` is a valid TimeZone? Or is the point that you can create custom timezone types?

src/timezones/types.jl
```diff
@@ -0,0 +1,212 @@
...
+    transitions::Vector{Transition}
+end
+
+function VariableTimeZone(name::String, transitions::Vector{Transition})
+    return VariableTimeZone(symbol(name), transitions)
+end
+
+immutable ZonedDateTime <: TimeType
+    utc_datetime::DateTime
+    timezone::TimeZone
```
quinnj: I think we'll want to define `ZoneDateTime{T<:TimeZone}` so `.timezone` can be concrete.

src/TimeZones.jl
omus: Ok. I did create a new method for `DateTime`.

src/timezones/io.jl
omus: I'm starting on some work for a PR to make `Base.Dates` work better for `TimeZones`.

src/timezones/types.jl
```diff
@@ -0,0 +1,212 @@
...
+    zone::FixedTimeZone  # The current zone for the utc_datetime.
+end
+
+"""
+Produces a list of possible UTC DateTimes given a local DateTime
+and a timezone. Results are returned in ascending order.
+"""
+function possible_dates(local_dt::DateTime, tz::VariableTimeZone; from_utc::Bool=false)
```
quinnj: Maybe it's just too early, but can you describe what are some examples here where more than one UTC DateTime would be returned?

src/timezones/types.jl
```diff
@@ -0,0 +1,212 @@
...
+        else
+            error("Multiple timezones found")
+        end
+    end
+
+    dt = DateTime(periods...)
+    return isnull(timezone) ? dt : ZonedDateTime(dt, get(timezone))
+end
+
+type AmbiguousTimeError <: Exception
```
quinnj: I like to put these kinds of errors near the top of the module to define them upfront before any usage.

test/timezone_names.jl
```diff
@@ -0,0 +1,7 @@
+import TimeZones: timezone_names
+
+names = timezone_names()
+
+@test length(names) >= 429
```
quinnj: I guess this is a good way to track the addition of new timezones :) Didn't North Korea just create their own? i wonder when that will be added.

src/timezones/types.jl
omus: I was planning on allowing the creation of custom `FixedTimeZone`. Checking if a TimeZone name exists is done in the `TimeZone(name::String)` function.

test/timezones/Olsen.jl
```diff
@@ -0,0 +1,201 @@
...
+@test test.transitions[6] == Transition(DateTime(1935,9,1,13), zone["TST-5"])
+
+# Note: Due to how I wrote the zones/rules a duplicate transition exists. The TimeZone code
+# should be able to safely handle this but nothing should require duplicates.
+@test test.transitions[6] == test.transitions[7]
+
+
+# Make sure that we can deal with Links. Take note that the current implementation converts
+# links into zones which makes it hard to explicitly test for a link. We expect that the
+# following link exists:
```
quinnj: With links, do you just copy the Zone/rules over with the new name?

src/TimeZones.jl
omus: I was leaning towards calling it `names` and not exporting it. Then the most of calls would be `TimeZones.names()`. Unfortunately `names` lists the functions exported by a module and may be confusing. Calling this function `timezones` could be misleading as someone may expect it to return all the `TimeZone` instances.

test/timezones/accessors.jl
```diff
@@ -0,0 +1,29 @@
...
+@test TimeZones.offset(fixed) == Second(-3600)
+@test_throws MethodError TimeZones.offset(warsaw)
+
+# ZonedDateTime accessors
+zdt = ZonedDateTime(DateTime(2014,6,12,23,59,58,57), fixed)
+@test TimeZones.localtime(zdt) == DateTime(2014,6,12,23,59,58,57)
+@test TimeZones.utc(zdt) == DateTime(2014,6,13,0,59,58,57)
+
+@test TimeZones.days(zdt) == 735396
+@test TimeZones.hour(zdt) == 23
```
quinnj: I'm not sure I quite understand. Wouldn't this return something other than 23? If `zdt` was created with `DateTime(2014,6,12,23,59,58,57)` + `fixed` timezone?

REQUIRE
quinnj: Ah, that makes sense. Thanks.

src/timezones/types.jl
quinnj: Cool, that's what I figured. I also see that `FixedTimeZone`s are meant to be pretty cheap and created/destroyed fairly quickly, so it's good to keep it lightweight.

test/timezone_names.jl
iamed2: It has already been added as of 2015-08-10 :)

src/TimeZones.jl
quinnj: `instances` maybe? I also wonder if this should be a Symbol array that we return, not sure if it makes that big of a difference.

If we want to get really fancy, we could call it `TimeZones.timezones()` and have it return some kind of `TimeZoneList` type that you could `getindex` by numeric index or symbol/name that would deserialize the `TimeZone`.

quinnj:
This is awesome @omus; definitely a lot of great work here; thanks for picking this up and running with it. Had you already started an olsen parser of sorts or did you just have to totally revamp the one I had cobbled together?

Anyway, I'm inclined to merge so that we can keep hacking on things from here, but if you'd rather wait and make tweaks, that's ok too.

src/timezones/types.jl
omus: I'm conflicted on this. First off due to all `TimeZone` types containing a `Symbol` the objects will never be inlined since `isbits` will never be true. Second, the timezone field is only around for when we perform calculations with the `ZonedDateTime` and is only around as a reference so it kinda makes sense to have it as a reference.

Are there any other reasons to parameterize this Type? I'm mostly just curious if I'm not thinking of some feature in Julia.

src/timezones/types.jl
omus: Sure. Typically this occurs when a VariableTimeZone "falls back" creating a duplicate hour:

```julia
julia> TimeZones.possible_dates(DateTime(1916, 10, 1, 0), TimeZone("Europe/Warsaw"))
2-element Array{Tuple{Base.Dates.DateTime,TimeZones.FixedTimeZone},1}:
 (1916-09-30T22:00:00,TimeZones.FixedTimeZone(:CEST,3600 seconds,3600 seconds))
 (1916-09-30T23:00:00,TimeZones.FixedTimeZone(:CET,3600 seconds,0 seconds))
```
When using the `from_utc` keyword you should never get more than one result. Mostly the `from_utc` keyword is here so I didn't have to duplicate this methods work.

omus: Note the the example I gave is from the test cases. I've also included tests that deal with 3 ambiguous hours (I don't believe such an event has occurred in reality)

omus:
@quinnj I reworked you Olsen database parser. Originally I just did enough to get it working in Julia 0.4 but then I found a couple of bugs and realized I needed to do some refactoring to make a solid test suite. 

For example one of the bugs was with the old `HourMin` type which had two fields `hour` and `min`. If you would parse a negative offset as a String such as `-1:20` when you calculated the total seconds it ended up doing `hour * 3600 + min * 60` which ended up resulting in -2400 instead -4800.

You can merge this right away. I do have some tweaks to make but they should be mostly minor at this point.

test/timezone_names.jl
omus: I specifically did `>=` to handle the creation of timezones. I don't expect that the total number of zones will decrease.

test/timezones/Olsen.jl
omus: Links just copy the `Zone` type. The `Rule`s get reused. We could change the code such that `resolve` also takes in links and only performs the copy at the last possible moment. I didn't feel that the minor decrease in memory was worth the restructure.

test/timezones/accessors.jl
omus: `ZonedDateTime` by default expects you are passing in a `DateTime` that is in the given `TimeZone`. Internally the `ZonedDateTime` converts that localtime into utc. Note the difference in the tests that convert the `zdt` into `utc` or `localtime`.

src/timezones/types.jl
omus: Good point. That reminds me I wanted to make the parent of `AmbiguousTimeError` and `NonExistentTimeError` be `TimeError`.

PR #4, opened 2015-08-13
Made *TimeError exceptions subtypes of TimeError.

omus:
Also moved the definitions before their first usage.

### DIFF
--- a/src/timezones/types.jl
+++ b/src/timezones/types.jl
@@ -3,6 +3,21 @@
 using Base.Dates
 import Base.Dates: value
 
+
+abstract TimeError <: Exception
+
+type AmbiguousTimeError <: TimeError
+    dt::DateTime
+    tz::TimeZone
+end
+Base.showerror(io::IO, e::AmbiguousTimeError) = print(io, "Local DateTime $(e.dt) is ambiguious");
+
+type NonExistentTimeError <: TimeError
+    dt::DateTime
+    tz::TimeZone
+end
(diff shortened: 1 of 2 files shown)


omus:
Should I make PRs for minor stuff like this? I don't mind having a second set of eyes on this but for minor changes like this I'm comfortable pushing this directly to master if you're ok with it.

src/timezones/types.jl
```diff
@@ -3,6 +3,21 @@
...
 import Base.Dates: value
 
+
+abstract TimeError <: Exception
+
+type AmbiguousTimeError <: TimeError
+    dt::DateTime
+    tz::TimeZone
+end
+Base.showerror(io::IO, e::AmbiguousTimeError) = print(io, "Local DateTime $(e.dt) is ambiguious");
```
quinnj: Is `e.dt` going to show the local datetime because this is thrown when trying to construct a `ZonedDateTime`? (just making sure there aren't other places where we'd through this where `dt` is UTC)

quinnj:
Definitely ok with you pushing stuff like this.

src/timezones/types.jl
omus: The `DateTime`s in these exceptions is just expected to be in localtime. We may want to extend the exceptions to be aware of what kind of `DateTime`s they contain (local/UTC). Most of the time you only get into trouble using local `DateTime`s which is why that is what is expected. There are some odd interactions within `type.jl` such as:

```julia
throw(NonExistentTimeError(localtime(zdt), tz))
```

omus:
Sounds good. Anything controversial I'll make sure to generate a PR for.

PR #5, opened 2015-08-25
Make use of extensible Base.Dates DateFormat

omus:
The following changes require that https://github.com/JuliaLang/julia/pull/12799 be merged into base Julia first. These changes make use of an extensible version of DateFormat which allows us to remove the overwritten methods.


omus:
Expect the checks to fail until the PR mentioned is merged.

omus:
Coverage decrease is due to one line that will involve a bunch of changes involving formatting. I'll merge this and deal with that after.

PR #7, opened 2015-10-05
Now

Michael-Klassen:
Add now(tz::TimeZone) function, similar to Base.now but returns a ZonedDateTime. Note that we are not using Base.now since that returns a DateTime in the local system time zone and it is preferable to use a UTC DateTime.
Add vectorized versions of accessors.
Create cast from ZonedDateTime to DateTime.


omus:
Build failed due to IANA again. I'll rebuild.

omus:
Changes look good.

PR #8, opened 2015-10-06
Support for reading tzfiles

Michael-Klassen:
Add support for parsing a TimeZone from a tzfile


omus:
We should add support for version 2 and version 3 before this gets merged. Here is a example of a [version 2 tzfile for Warsaw](https://github.com/quinnj/TimeZones.jl/files/9518/Warsaw.Version.2.txt).

omus:
Seems to be something up with 32-bit support. Looking into it.

PR #9, opened 2015-10-06
Added extra FixedTimeZones

Michael-Klassen:
Add support for some common fixed time zones


omus:
Please add some [coverage for fixed.jl](https://codecov.io/github/quinnj/TimeZones.jl/src/timezones/fixed.jl?ref=3fd14f780fbee3da230633a4ad39dd0aecf7a834). Currently the file has no testcases that use the `fixed_timezones` function.

omus:
I've decided that the ["Etc/..." timezones are historical](ftp://ftp.iana.org/tz/data/etcetera) and shouldn't be used. For those who want to use something like `FixedTimeZone("UTC-2", -7200)`. Making the FixedTimeZone yourself should be much safer as the "Etc/GMT±X" timezones may be negated depending on your view.

I've added some notes into build.jl which acknowledges the existence of the "etcetera" file. The only new TimeZones added by this PR are: UTC and GMT.

PR #10, opened 2015-10-13
Initial transition effective from typemin(DateTime)

omus:
VariableTimeZone types are made up of transitions which have a UTC DateTime indicating the instant at when the new rule becomes effective. Previously the first transition started at 1800/1/1 for no real reason except that no zone or rule started that early and it is rather silly to use the concept of time zones prior to their use. After merging in #8 it seems more reasonable to allow users to make ZonedDateTime's prior to 1800 and let them use their own judgement.

As a happy side effect this makes comparing Julia processed timezone rules easier to compare with tzfiles.


quinnj:
+1

PR #12, opened 2015-10-14
Add support for local system TimeZone

omus:
Determines the local systems TimeZone via the `localzone` function. Unfortunately code currently is not well tested due to OS specific code and a number of system calls or reading external files. We should be able to improve coverage when we can utilize some kind of method mocking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 # Ignore dependencies that are build/retrieved
 deps/tzdata
 deps/compiled
+deps/windows
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@
 # Ignore dependencies that are build/retrieved
 deps/tzdata
 deps/compiled
-deps/windows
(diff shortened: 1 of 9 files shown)


omus:
@quinnj do you have an ideas for mocking methods in Julia? The [Fixtures.jl](https://github.com/burrowsa/Fixtures.jl) package looks like it was attempting to do this but it doesn't work with 0.4 and hasn't been updated in 2 years. I've been playing around with adding this functionality to Julia.

omus:
Also, are you comfortable with the current decrease in coverage? I'll try to get the coverage improved in the future.

omus:
I have made a branch called "mocking" that uses a currently unpublished package called [Mocking.jl](https://github.com/invenia/Mocking.jl) which allows me to overwrite methods inside of the localzone function to have more complete test coverage. The [coverage on this branch](https://codecov.io/github/quinnj/TimeZones.jl?branch=mocking) is 90.49% which shows this code works pretty well. I'm going to merge this PR without the additional testing for now as I think it is unwise to force people to use an unpublished package.

omus:
Mocking.jl has now been published so I've merged the additional tests in cbe9b7a

PR #13, opened 2015-11-19
Added TimeZone() constructor that doesn't require arguments

JobJob:
Wasn't really sure what tests to add for this...

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -56,6 +56,7 @@ end
 
 doc"""
 `TimeZone() -> TimeZone`
+
 Construct `TimeZone` information corresponding to the local timezone
 """
 TimeZone() = localzone()
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -15,7 +15,7 @@ export TimeZone, FixedTimeZone, VariableTimeZone, ZonedDateTime, DateTime, timez
     # Re-export from Base.Dates
     yearmonthday, yearmonth, monthday, year, month, week, day, dayofmonth,
     # conversion.jl
-    now,
+    now, nowtz,
     # local.jl
     localzone
 
@@ -53,13 +53,6 @@ function TimeZone(name::AbstractString)
         return deserialize(fp)
     end
 end
-
-doc"""
-`TimeZone() -> TimeZone`
-
-Construct `TimeZone` information corresponding to the local timezone
-"""
-TimeZone() = localzone()
 
 doc"""
 `timezone_names() -> Array{AbstractString}`
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -55,6 +55,12 @@ function TimeZone(name::AbstractString)
 end
 
 doc"""
+`TimeZone() -> TimeZone`
+Construct `TimeZone` information corresponding to the local timezone
+"""
+TimeZone() = localzone()
+
+doc"""
 `timezone_names() -> Array{AbstractString}`
 
 Returns all of the valid names for constructing a `TimeZone`.
--- a/src/timezones/types.jl
+++ b/src/timezones/types.jl
@@ -169,8 +169,10 @@ doc"""
 `ZonedDateTime(local_dt::DateTime, tz::VariableTimeZone, occurrence::Integer=0; from_utc::Bool=false) -> ZonedDateTime`
 
 Constructs a `ZonedDateTime` given a local `DateTime` and a `TimeZone`. If the local
-`DateTime` is ambiguious in the given time zone you can set `occurrence` to a positive
-integer to resolve the ambiuity. When the `from_utc` keyword is true the given `DateTime` is
+`DateTime` is ambiguious in the given time zone, e.g. 02:01 on a date when clocks have moved from
+03:00->02:00 at the end of daylight time, you can set `occurrence` to a positive
+integer to resolve the ambiuity. In the example given, 2 would set it to the second occurrence of 02:01
+i.e. after the clocks had been turned back. When the `from_utc` keyword is true, the given `DateTime` is
 processed as if it is in UTC.
 """
 function ZonedDateTime(local_dt::DateTime, tz::VariableTimeZone, occurrence::Integer=0; from_utc::Bool=false)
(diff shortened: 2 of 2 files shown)


omus:
I'm curious what your particular use case is for this? I was thinking of having localzone() be the default for the ZonedDateTime constructor.

JobJob:
Yeah that's my use case, I wanted a version of `now()` which had milliseconds, I tried `now(Dates.UTC)` but that only gave me UTC time not local time, so in the end I arrived at: `ZonedDateTime(now(Dates.UTC), localzone())` 

I guess my first instinct after pulling the package was to go for `TimeZone()`, had to search a little to find `localzone()` and figured others might be the same, but I didn't really think about it that much. 

Maybe `TimeZone()` is not needed, and instead this https://github.com/quinnj/TimeZones.jl/blob/master/src/timezones/types.jl#L230 could be changed to `function ZonedDateTime(local_dt::DateTime = now(Dates.UTC), tz::FixedTimeZone=localzone(); from_utc::Bool=false)`

I'd personally like it if there was also a `nowtz() = TimeZones()`

Want me to change this PR to that?

omus:
If you want a version of `now()` that does milliseconds you should use `Dates.unix2datetime(time())`. That will give you the current time in milliseconds in you system timezone.

You need to be careful with defaults values in methods. Setting `localzone()` in the `ZonedDateTime` definition like that will call `localzone()` once when the function is compiled and never again. If you were to change your system's timezone after compiling that function it would still default to the timezone from when the method was compiled.

I think the solution is to make additional methods that exclude the timezone argument. That way we can call `localzone()` from within the method body and avoid the issue outlined above. If you want to implement what you're proposing you'd have to at least make two new methods:

```julia
ZonedDateTime(local_dt::DateTime, is_dst::Bool; from_utc::Bool=false)
ZonedDateTime(local_dt::DateTime, occurrence::Integer=0; from_utc::Bool=false)
```

I'm unclear in what you'd want `nowtz()` to do. Is it essentially `now(tz::TimeZone)` where tz is optional?

JobJob:
`Dates.unix2datetime(time())` actually gives UTC time, not local time, `now()` gives me my system's time. You can see in https://github.com/JuliaLang/julia/blob/master/base/dates/conversions.jl#L27 that `now(::Type{UTC})` is defined as `unix2datetime(time())` (n.b. `time()` call in Libc calls `clock_now` in support/timefuncs.c which makes a syscall to gettimeofday() ignoring system tz info)

Re `nowtz()` yes exactly!

Interesting and important point re default arguments and compilation, hopefully what I've done is fine. Sorry about the original hasty pull request before really looking at the code.

Let me know what you think of the updates I've made and any tests I might add.

Nice package btw - great work.

omus:
My bad. `Dates.unix2datetime(time())` is definitely in UTC.

src/TimeZones.jl
```diff
@@ -15,7 +15,7 @@ export TimeZone, FixedTimeZone, VariableTimeZone, ZonedDateTime, DateTime, timez
     # Re-export from Base.Dates
     yearmonthday, yearmonth, monthday, year, month, week, day, dayofmonth,
     # conversion.jl
-    now,
+    now, nowtz,
```
omus: Looks like you forgot to commit the function `nowtz`.

src/timezones/types.jl
```diff
@@ -222,6 +224,57 @@ function ZonedDateTime(local_dt::DateTime, tz::VariableTimeZone, is_dst::Bool; f
 end
 
 doc"""
+`ZonedDateTime(tz::VariableTimeZone, is_dst::Bool; from_utc::Bool=false) -> ZonedDateTime`
+
+Get a `ZonedDateTime` object in the timezone specified, for the current time.
+
+equivalent to `ZonedDateTime(now(Dates.UTC), tz, is_dst; from_utc=true)`
+"""
+ZonedDateTime(local_dt::DateTime, is_dst::Bool) = ZonedDateTime(now(Dates.UTC), tz, is_dst; from_utc=true)
```
omus: You can remove this method. It will be overwritten by: `ZonedDateTime(local_dt::DateTime, is_dst::Bool; from_utc::Bool=false)`. Remember keywords are not included in multiple-dispatch.

src/TimeZones.jl
omus: My bad. I just missed it.

src/timezones/types.jl
```diff
@@ -222,6 +224,57 @@ function ZonedDateTime(local_dt::DateTime, tz::VariableTimeZone, is_dst::Bool; f
...
+
+`ZonedDateTime(occurrence::Integer) -> ZonedDateTime`
+
+ZonedDateTime() will give you a ZonedDateTime for the current time in your local time zone.
+
+equivalent to `ZonedDateTime(now(Dates.UTC), localzone(); from_utc=true)` if occurrence is not set or
+
+equivalent to `ZonedDateTime(now(Dates.UTC), localzone(), occurrence; from_utc=true)` if it is
+"""
+ZonedDateTime(occurrence::Integer=0) = ZonedDateTime(now(Dates.UTC), localzone(); from_utc=true)
```
omus: This should probably be implemented as `now(localzone())`. This does lose millisecond precision though. If you really want millisecond precision:

```julia
function ZonedDateTime(occurrence::Integer=0)
    utc = unix2datetime(time())
    ZonedDateTime(utc, localzone(); from_utc=true)
end
```
I would like to avoid using the `now(Dates.UTC)` method as I'll probably push for its removal from Base.

src/timezones/types.jl
omus: Oh I you're documentation seems to be the implementation you want. You should update the code to reflect the code shown in the documentation.

Michael-Klassen:
`ZonedDateTime()` should probably return `ZonedDateTime(DateTime(), "UTC")`

omus:
After taking a good look over the PR I'm inclined to agree with @Michael-Klassen. The `ZonedDateTime` methods are intended to mimic the `DateTime` constructors.

```julia
julia> DateTime()
0001-01-01T00:00:00

julia> ZonedDateTime(DateTime(), TimeZone("UTC"))
0001-01-01T00:00:00+00:00
```
As for the other constructors I think it they could either default to `localzone()` or `TimeZone("UTC")` with a preference towards UTC. I'm thinking that leaving things as they are forces uses to pass in a `TimeZone` explicitly and avoid issues there.

omus:
As for `nowtz()` I'm also leaning towards just having `now(localzone())`. 

I think your issue that `now()` doesn't provide millisecond precision is reasonable and I think you should [open an issue in Base Julia]((https://github.com/JuliaLang/julia/issues) about it. If you want to fix it yourself you can try removing the [floor call in TmStruct](https://github.com/JuliaLang/julia/blob/master/base/libc.jl#L101) which seems to be the problem. Since I'm emulating the behaviour of `now()` in `now(::TimeZone)` I would update my method to have the same precision.

omus:
@JobJob does this seem reasonable to you? I'll close the PR if you agree. I'll make sure to include the relevant documentation changes you made.

JobJob:
Yes I agree they should be consistent. Though I might add that `DateTime()` defaulting to 0001-01-01T00:00:00+00:00 was not what I expected when I first encountered it, however given that now() exists it seems sensible enough. You guys seem to be much more on top of the landscape and I defer to your judgement. 

Re docs: the main thing that I didn't twig to immediately was the bit about ambiguous times, so keeping a little note about that would be good.

This was just a little bug chase that escalated a little further than it probably should have, but let's keep on fighting the good fight :)

omus:
@JobJob all good. I'm glad you brought this up as it helped me make a decision on the extra `ZonedDateTime` methods. I'll try to make more detailed documentation soon.

PR #14, opened 2015-11-24
Allow user-specified cutoffs for parsing Olson databases.

bcj:
When parsing Olson files, rules without set `to` fields only have transitions calculated up until the end of 2038. This leads to an inconsistency in how transitions are calculated: rules with set `to` fields will have transitions calculated after that cutoff date (even if they start after that cutoff date).

While a better long-term fix would be to re-architect the code so that further transitions could be calculated after the fact, this PR mitigates the issue by: allowing the user to pass in their own cutoff, and by making `order_rules` handle that cutoff in a consistent manner. The cutoff is now strictly enforced and `to` fields are truncated to that date.

Rule|Old Code|This Code (with max_year=2038)
------|-------------|--------------
from <= 2038, to <= 2038|all transitions|all transitions
from <= 2038, to > 2038|all transitions|transitions before 2039, to changed to 2038
from <= 2038, to = Null|transitions before 2039|transitions before 2039
from > 2038, to > 2038|all transitions|no transitions
from > 2038, to = Null|no transitions|no transitions

These two changes ensure that timezones more accurately reflect olson data, and users working with dates far in the future can easily use this library.

### DIFF
--- a/src/timezones/Olson.jl
+++ b/src/timezones/Olson.jl
@@ -47,7 +47,7 @@ typealias OrderedRuleDict Dict{AbstractString,Tuple{Array{Date},Array{Rule}}}
 
 # Min and max years that we create DST transition DateTimes for (inclusive)
 const MIN_YEAR = year(typemin(DateTime))  # Essentially the begining of time
-const MAX_YEAR = 2038                     # year(typemax(Int32))
+const MAX_YEAR = 2037                     # year(unix2datetime(typemax(Int32))) - 1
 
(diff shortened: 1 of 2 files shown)


omus:
@bcj can you look into the 32-bit issues?

bcj:
will do.

bcj:
I'm a little confused about this coverage decreasing. I didn't touch `local.jl` but its coverage went down?

PR #15, opened 2015-12-01
Use millisecond precision with now(::TimeZone)

omus:
The behaviour of `now(::TimeZone)` matches that of `Base.now()`. This change should only be merged if https://github.com/JuliaLang/julia/pull/14170 gets accepted into base Julia.

cc: @JobJob


omus:
Milliseconds change was merged into base Julia.

JobJob:
Good work @omus - cheers.

PR #18, opened 2016-01-11
ZonedDateTime allows creation of dates after max_year

bcj:
When we truncate transitions for timezones, we should not allow dates in those timezones to be created for periods after that truncation.

PR for https://github.com/quinnj/TimeZones.jl/issues/17


src/timezones/Olson.jl
```diff
@@ -332,7 +341,10 @@ function resolve!(zone_name::AbstractString, zoneset::ZoneDict, ruleset::RuleDic
             end
         else
             if !haskey(ordered, rule_name)
-                ordered[rule_name] = order_rules(ruleset[rule_name]; max_year=max_year)
+                dates, ordered_rules, was_truncated = order_rules(ruleset[rule_name]; max_year=max_year)
+
+                ordered[rule_name] = dates, ordered_rules
+                truncated |= was_truncated
```
omus: The truncated information will be lost on future passes where `rule_name` is a key within `ordered`.

src/timezones/Olson.jl
```diff
@@ -432,7 +444,12 @@ function resolve!(zone_name::AbstractString, zoneset::ZoneDict, ruleset::RuleDic
     # Note: Transitions array is expected to be ordered and should be if both
     # zones and rules were ordered.
     if length(transitions) > 1
-        return VariableTimeZone(zone_name, transitions)
+        if truncated
+            maybe_max = Nullable(max_year)
+        else
+            maybe_max = Nullable()
+        end
```
omus: Consider doing `maybe_max = Nullable(truncated ? max_year : nothing)`

src/timezones/types.jl
```diff
@@ -111,17 +111,39 @@ doc"""A `TimeZone` with an offset that changes over time."""
...
+    max_year::Nullable{Int}
+end
+
+function VariableTimeZone(name::AbstractString, transitions::Vector{Transition}, max_year::Nullable{Int})
+    return VariableTimeZone(symbol(name), transitions, max_year)
 end
 
 function VariableTimeZone(name::AbstractString, transitions::Vector{Transition})
-    return VariableTimeZone(symbol(name), transitions)
+    return VariableTimeZone(symbol(name), transitions, Nullable())
```
omus: Switch to `Nullable{Int}()` for type stability

src/timezones/types.jl
```diff
@@ -111,17 +111,39 @@ doc"""A `TimeZone` with an offset that changes over time."""
...
 end
 
+type OutOfRangeTimeError <: TimeError
+    tz::VariableTimeZone
+end
+Base.showerror(io::IO, e::OutOfRangeTimeError) = print(io, "TimeZone $(string(e.tz)) does not handle dates after $(string(e.tz.max_year.value))");
```
omus: `$(string(e.tz.max_year.value))` should probably be changed to `$(string(get(e.tz.max_year, -1)))` or something similar. Otherwise entries where `e.tz.max_year == Nullable{Int}()` will be some random value.

src/timezones/types.jl
```diff
@@ -111,17 +111,39 @@ doc"""A `TimeZone` with an offset that changes over time."""
...
     utc_datetime::DateTime
     timezone::TimeZone
     zone::FixedTimeZone  # The current zone for the utc_datetime.
+
+    function ZonedDateTime(utc_datetime::DateTime, timezone::TimeZone, zone::FixedTimeZone)
+        return new(utc_datetime, timezone, zone)
+    end
+
+    function ZonedDateTime(utc_datetime::DateTime, timezone::VariableTimeZone, zone::FixedTimeZone)
+        if !isnull(timezone.max_year) && year(utc_datetime + Second(zone.offset)) > timezone.max_year.value
```
omus: Seems odd the limit is based upon the year in the given timezone. Maybe instead of using max_year we should set the limit as the next transition instead?

src/timezones/Olson.jl
bcj: I think this code is right. `resolve!` builds a single timezone and we want truncated to be `true` if any of the rules within that timezone are truncated. The `|=` should be enabling that behavior, right?

src/timezones/Olson.jl
bcj: Will do. I forgot about `nothing`

src/timezones/types.jl
bcj: Sure. I hadn't thought about that because the only place _I_ was throwing this error was when max_year exists

src/timezones/types.jl
bcj: as in, instead of max_year have a max_date instead? I would be fine with this. The reason I chose year, was because we always do all transitions until the end of a year.

src/timezones/Olson.jl
omus: The problem comes from ordered being used between calls to `resolve!`. See the [`resolve`](https://github.com/invenia/TimeZones.jl/blob/4650735488cb10734883bfa9182fadaa68533a4b/src/timezones/Olson.jl#L460) functions.

bcj: Alright, I'm not seeing this. I'll be in this afternoon and we can walk through this then

omus:
@bcj can you rebase this on the the current master?

bcj:
tada

bcj:
alright, let's ignore that terrible thing I did that is now gone in the history. My tests pass locally, if CI succeeds I think we're good

omus:
I iterated a bit on the original design and added a bunch of additional tests to ensure that cutoff is working as expected.

PR #22, opened 2016-02-26
WIP: Convenience constructors for ZonedDateTime

omus:
There has been some interest in creating a `ZonedDateTime` without first creating a `DateTime` object. This pull request implements the requested functionality.

I'll note we could also have additional constructors that take in a `AbstractString` timezone name rather than having to create a `TimeZone` object but this leads into issues with knowing if the `TimeZone` is a `VariableTimeZone` or not and whether we need to have the `is_dst` argument supplied.

### DIFF
--- a/docs/arithmetic.md
+++ b/docs/arithmetic.md
@@ -8,7 +8,7 @@ julia> using Dates
 julia> warsaw = TimeZone("Europe/Warsaw")
 Europe/Warsaw
 
-julia> spring = ZonedDateTime(DateTime(2014,3,30), warsaw)
+julia> spring = ZonedDateTime(2014, 3, 30, warsaw)
 2014-03-30T00:00:00+01:00
 
 julia> spring + Day(1)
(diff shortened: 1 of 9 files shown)


omus:
The extra constructors could be simplified if we could intermix optional arguments and arguments:
```julia
ZonedDateTime{I<:Integer}(y::I, [m::I=1, d::I=1, h::I=0, mi::I=0, s::I=0, ms::I=0], tz::TimeZone)
```
See https://github.com/JuliaLang/julia/issues/15254

omus:
@quinnj I'd like to get your feedback on this change

omus:
I'll do a quick summary of this PR and where things are at:

Creating a `ZonedDateTime` by first having to construct a `DateTime` feels clunky. For example
```julia
ZonedDateTime(DateTime(2014), TimeZone("Europe/Warsaw"))
# vs
ZonedDateTime(2014, TimeZone("Europe/Warsaw"))
```
Unfortunately since optional parameters can not precede required parameters it makes it hard for us make constructors like this where the last parameter is the `TimeZone`. Personally I feel that having the `TimeZone` at the end reads best.

I discovered that in this particular use case a macro (`@optional`) could be used to allow optional parameters to precede required parameters. Unfortunately this is problematic to generalize but works well for our particular use case.

omus:
Failures on Julia 0.5 are due to a bug in Mocking.jl Will re-run once the fix is released.

quinnj:
Hmmm......while the whole `@optional` macro feels a little much for what you get here, I do like that it's not something user-facing or that the user is ever aware of. I'm ok adding this.

omus:
Thanks @quinnj. The `@optional` macro definitely isn't required for these additional constructors but it should make maintenance easier. I'll merge this right away.

quinnj:
Definitely and sorry it took so long to review. I'm trying to catch up on a lot of stuff after a pretty busy spell

omus:
No worries. I wasn't in a rush to get this PR in.

PR #23, opened 2016-03-17
Switch to using a HTTP source

### LINKED ISSUE #11: FTP IANA source unreliable
TimeZones.jl retrieves its information directly from [IANA's FTP server](ftp://ftp.iana.org/tz/data/). This server commonly cannot be connected to resulting in the error message from the download function:
```
curl: (56) Recv failure: Connection reset by peer
```

We currently are automatically retrying to download files up to three times which works most of the time but can still cause Travis builds to fail or be annoying for first time uses. We should either find some reliable mirrors or maybe keep a copy of the timezone rules within the repo.

omus:
Previously we were downloading individual uncompressed tzdata files directly from IANA's FTP server. Unfortunately, downloading from this FTP server periodically results in:

    curl: (56) Recv failure: Connection reset by peer

We managed to work around these connection issues by retrying downloads but the FTP server was still problematic for machines behind restricted networks (Note: passive FTP can use a wide range of ports).

This PR switches the build process to download tzdata from an official IANA HTTP server which appears to be much more reliable and won't cause issues with restrictive networks. In order to support this change the we needed to be able to decompress the tar.gz which requires a small additional step in the build process.

Closes #11

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -4,6 +4,13 @@ import TimeZones.Olson: compile
 @windows_only import TimeZones: WIN_TRANSLATION_FILE
 @windows_only using LightXML
 
+# Various sources from which the latest compressed TZ data can be retrieved.
+# Note: HTTP sources are preferable as they tend to work behind firewalls.
+const URLS = (
+    "https://www.iana.org/time-zones/repository/tzdata-latest.tar.gz",
+    "ftp://ftp.iana.org/tz/tzdata-latest.tar.gz",  # Unreliable source
+)
+
 # See "ftp://ftp.iana.org/tz/data/Makefile" PRIMARY_YDATA for listing of
 # regions to include. YDATA includes historical zones which we'll ignore.
 const REGIONS = (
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -21,20 +21,27 @@ isdir(TZDATA_DIR) || mkdir(TZDATA_DIR)
...
     end
 end
-isfile(tzdata) || error("Unable to download TZ data")
+isfile(archive) || error("Unable to download TZ data")
 
 info("Extracting TZ data")
-run(`tar xvf $tzdata -C $TZDATA_DIR $REGIONS`)
-rm(tzdata)
+@unix_only function extract(archive, directory, files)
+    run(`tar xvf $archive --directory=$directory $files`)
```
iamed2: Shouldn't this be `-xvf`? Also, maybe remove the `v`.

iamed2:
Should `$directory` and `$files` be quoted in case they contain spaces?

iamed2:
Has this been tested with GNU and BSD tar? I can't find the `--directory` option in the BSD tar man page.

iamed2:
We should try tar in addition to 7zip on Windows in case they have tar installed.

deps/build.jl
```diff
@@ -15,36 +20,28 @@ const REGIONS = (
...
+info("Extracting TZ data")
+@unix_only function extract(archive, directory, files)
+    run(`tar xvf $archive --directory=$directory $files`)
+end
+@windows_only function extract(archive, directory, files)
+    run(pipeline(`7z x $archive -y -so`, `7z x -si -y -ttar -o$directory $files`))
+end
+
+extract(archive, TZDATA_DIR, REGIONS)
+rm(archive)
```
iamed2: Seems unnecessarily aggressive to remove the archive.

omus:
@iamed2 quoting happens automatically with `Cmd` objects. You can always see how this works by displaying the `Cmd` object.

I haven't personally tested in with BSD tar but I believe `--directory` is safe to use as that is what is BinDeps.jl uses for [`unpack_cmd`](https://github.com/JuliaLang/BinDeps.jl/blob/master/src/BinDeps.jl#L80).

deps/build.jl
omus: We don't have to remove the archive. The file is being downloaded to a temporary file so this is just cleanup. You think it would be useful for debugging to just leave the archive?

iamed2:
`--directory` seems to work in practice for me. Weird that it isn't in the man page.

omus:
I'll note that GNU tar does mention it in the man pages. It just seems to be BSD that doesn't mention it. If you think it would be clearer `-C` is mentioned in the man page for BSD and GNU.

iamed2:
`--directory` is clearer. If for some reason someone runs on a system where that fails, they can file an issue and we'll change it.

deps/build.jl
omus: It's pretty normal for `tar` not to have the hypen:
> The first argument to tar should be a function; either one of the letters Acdrtux, or one of the long function names.  A function letter need not be prefixed with ``-'', and may be combined with other single-letter options

omus: As for the verbose option I think it's rather useful here as it shows which individual tzdata files we're extracting:
```
INFO: Extracting TZ data
x africa
x antarctica
x asia
x australasia
x europe
x northamerica
x southamerica
```

iamed2: TIL about tar

deps/build.jl
iamed2: Yeah. Plus a weird combination situation I came up with in my head.

omus:
The Julia package evaluator found that the test cases for TimeZones.jl [failed for v0.4 starting 2016-03-17](http://pkg.julialang.org/detail/TimeZones.html). The failure was due to the "Connection reset by peer" error we get from the FTP source.

quinnj:
Catching up on old emails; +1 to this!

PR #31, opened 2016-08-12
Improve Range Math Support

omus:
Extends StepRange math to avoid running into issues where the start/stop is either a non-existent or ambiguous DateTime.

Additionally, cleaned up some of the code and refactored `possible_dates` into `interpret`. Added lots of new helper functions for dealing with converting local DateTimes into ZonedDateTime.


src/timezones/interpret.jl
```diff
@@ -0,0 +1,193 @@
...
+end
+
+"""
+    transition_range(dt::DateTime, tz::VariableTimeZone, context::Type{Union{Local,UTC}}) -> UnitRange
+
+Finds the indexes of the `tz` transitions which may be applicable for the `dt`. The given
+DateTime is expected to be local to the time zone or in UTC as specified by `context`. Note
+that UTC context will always return a range of length one.
+"""
+transition_range(dt::DateTime, tz::VariableTimeZone, context::Type{Union{Local,UTC}})
```
iamed2: Are you just calling this function here?

test/timezones/ranges.jl
```diff
@@ -77,3 +77,17 @@ fixed_range = ZonedDateTime(d, fixed):Hour(1):ZonedDateTime(d, fixed) + Day(1)
...
 @test collect(map(TimeZones.utc, no_dst_range)) == raw_range .+ Hour(6)
 @test collect(map(TimeZones.utc, fixed_range)) == raw_range .+ Hour(5)
+
+# recur behaviour with a non-existent hour
+range = ZonedDateTime(2015, 3, 8, dst):Dates.Hour(1):ZonedDateTime(2015, 3, 10, dst)
+@test Dates.recur(dt -> Dates.hour(dt) == 2, range) == [
+    ZonedDateTime(2015, 3, 9, 2, dst)
+]
+
+# recur behaviour with ambiguoous hour
```
omus: Keyboard's fault. I swear

src/timezones/interpret.jl
```diff
@@ -0,0 +1,193 @@
...
+    earliest = local_dt + MIN_OFFSET
+    latest = local_dt + MAX_OFFSET
+
+    # Determine the earliest transition the local DateTime could
+    # occur within.
+    start = searchsortedlast(
+        transitions, earliest,
+        by=el -> isa(el, Transition) ? el.utc_datetime : el,
+    )
+    start = max(start, 1)
```
iamed2: Why max?

src/timezones/interpret.jl
```diff
@@ -0,0 +1,193 @@
...
+            push!(interpretations, ZonedDateTime(utc_dt, tz, t[i].zone))
+        end
+    end
+
+    return interpretations
+end
+
+function interpret(utc_dt::DateTime, tz::VariableTimeZone, ::Type{UTC})
+    range = transition_range(utc_dt, tz, UTC)
+    length(range) == 1 || error("A UTC DateTimes should only have a single interpretation")
```
iamed2: Let the user know that this is a problem with TimeZones, not their code.

Also pick either "A UTC DateTime" or "UTC DateTimes"

src/timezones/interpret.jl
omus: Nope, it's a way of specifying what method the docstring is attached to

```julia
julia> methods(println)
# 3 methods for generic function "println":
println(io::IO) at coreio.jl:6
println(io::IO, xs...) at strings/io.jl:27
println(xs...) at coreio.jl:5

julia> """println rocks!"""
       println(xs::AbstractString)
Base.println

help?> println
search: println print_with_color print print_joined print_escaped

  println(x)

  Print (using print) x followed by a newline.

  println rocks!

julia> methods(println)
# 3 methods for generic function "println":
println(io::IO) at coreio.jl:6
println(io::IO, xs...) at strings/io.jl:27
println(xs...) at coreio.jl:5

help?> println(::AbstractString)
  println rocks!
```

src/timezones/interpret.jl
omus: I'm clamping the `start` to be >= 1. In the case where `earliest` precedes all transitions we can get 0 as the start:
```julia
julia> searchsortedlast([2,3,4], 1)
0
```

src/timezones/interpret.jl
omus: > Let the user know that this is a problem with TimeZones, not their code.

What do you suggest?

> Also pick either "A UTC DateTime" or "UTC DateTimes"

Definitely a mistake.

src/timezones/interpret.jl
iamed2: It really looks like a function call. I'd prefer the arguments go unnamed a la 
```julia
transition_range(::DateTime, ::VariableTimeZone, ::Type{Union{Local,UTC}})
```
which is also valid syntax.

src/timezones/interpret.jl
iamed2: Oh f*** I read it as MATLAB max-across-dimension. Sorry.

src/timezones/interpret.jl
iamed2: > What do you suggest?
Appending "Unexpected TimeZones error:" is probably fine.

src/timezones/interpret.jl
omus: I agree. Leaving out the variable names avoids confusion

src/timezones/interpret.jl
omus: 👍

iamed2:
👍

omus:
Thanks @iamed2. I'll merge this once the checks complete.

Issue #41, opened 2016-12-27
Default ISO Zoned DateTime format only works for multiples of 100 Milliseconds


joshbode:
I had expected that `string` and the `AbstractString` constructor for `ZonedDateTime` would be inverses of one another, e.g. `x == ZonedDateTime(string(x))` as is the case for `Date` and `DateTime`.

However, it only appears to work for `ZonedDateTime`s where the milliseconds are a multiple of 100.

```julia
julia> using TimeZones

julia> tz = TimeZone("Australia/Melbourne")
Australia/Melbourne (UTC+10/UTC+11)

julia> x = ZonedDateTime(now(), tz)
2016-12-28T10:00:50.316+11:00

julia> string(x)
"2016-12-28T10:00:50.316+11:00"

julia> ZonedDateTime(string(x))
ERROR: ArgumentError: Unrecognized time zone: 16+11:00
 in TimeZones.FixedTimeZone(::String) at /Users/josh/.julia/v0.5/TimeZones/src/types.jl:86
 in slotparse(::Base.Dates.DelimitedSlot{Base.Dates.TimeZone}, ::String, ::String) at /Users/josh/.julia/v0.5/TimeZones/src/io.jl:64
 in getslot(::String, ::Base.Dates.DelimitedSlot{Base.Dates.TimeZone}, ::String, ::Int64) at ./dates/io.jl:160
 in parse(::String, ::Base.Dates.DateFormat) at ./dates/io.jl:175
 in TimeZones.ZonedDateTime(::String) at /Users/josh/.julia/v0.5/TimeZones/src/io.jl:94
```

However, if I truncate to multiples of 100 milliseconds, it works:
```julia
julia> y = trunc(x, Second) + Millisecond(trunc(Int64(Millisecond(x)) / 100) * 100)
2016-12-28T10:00:50.3+11:00

julia> ZonedDateTime(string(y))
2016-12-28T10:00:50.3+11:00
```

## Version Info
```julia
julia> versioninfo()
Julia Version 0.5.0
Commit 3c9d753* (2016-09-19 18:14 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin15.6.0)
  CPU: Intel(R) Core(TM)2 Duo CPU     P7550  @ 2.26GHz
  WORD_SIZE: 64
  BLAS: libopenblas (DYNAMIC_ARCH NO_AFFINITY Penryn)
  LAPACK: libopenblas
  LIBM: libopenlibm
[...]

PR #42, opened 2016-12-27
Increase width of Millisecond slot in ISO format from 1 to 3

joshbode:
Change ISO Zoned DateTime format:

    yyyy-mm-ddTHH:MM:SS.szzz -> yyyy-mm-ddTHH:MM:SS.ssszzz

To prevent loss of precision. Addresses issue #41.


omus:
Thanks for the PR. Looks good!

joshbode:
Thanks for that!

PR #46, opened 2017-03-02
Pin version of tzdata to 2016j for tests

omus:
Allows modifications to the tz database to not affect the tests which use `resolve`. Note that users can specify the TZDATA_DIR env variable if they which to run the tests against a different version of the tz database.

Overall the change should avoid having PackageEvaluator.jl fail when just the tzdata has been changed.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,6 +1,6 @@
 import TimeZones: TZDATA_DIR, COMPILED_DIR, extract
 import TimeZones.Olson: compile
-import Compat: @static, is_windows
+import Compat: is_windows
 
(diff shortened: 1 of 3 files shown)


omus:
Currently the CI tests will download the latest version then the 2016j version. This is minor but I'll definitely want to be able to only download the 2016j version when running in the CI.

omus:
I haven't had much time to work on this lately. The current changes do work on 0.4 and 0.5 (0.6 is currently broken with TimeZones.jl) so I'll merge this change and deal with avoiding downloading tzdata twice in another PR.

PR #47, opened 2017-03-10
Add `zdt2unix` and `unix2zdt` for timestamp conversion

spurll:
These functions allow us to easily convert `ZonedDateTime`s to Unix timestamps (in UTC).

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -1,5 +1,8 @@
 import Base.Dates: now, unix2datetime
 
+# UTC is an abstract type defined in Base.Dates, for some reason
+const utc_tz = FixedTimeZone("UTC")
+
 """
     DateTime(::ZonedDateTime) -> DateTime
 
@@ -41,4 +44,16 @@ end
 
 function astimezone(zdt::ZonedDateTime, tz::FixedTimeZone)
     return ZonedDateTime(zdt.utc_datetime, tz, tz)
+end
+
+function zdt2unix(zdt::ZonedDateTime)
+    Dates.datetime2unix(utc(zdt))
+end
+
+function zdt2unix{T<:Number}(t::Type{T}, zdt::ZonedDateTime)
+    floor(t, Dates.datetime2unix(utc(zdt)))
+end
+
+function unix2zdt(seconds::Integer)
+    ZonedDateTime(Dates.unix2datetime(seconds), utc_tz, from_utc=true)
 end
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -50,8 +50,12 @@ function zdt2unix(zdt::ZonedDateTime)
     Dates.datetime2unix(utc(zdt))
 end
 
-function zdt2unix{T<:Number}(t::Type{T}, zdt::ZonedDateTime)
+function zdt2unix{T<:Integer}(t::Type{T}, zdt::ZonedDateTime)
     floor(t, Dates.datetime2unix(utc(zdt)))
+end
+
+function zdt2unix{T<:Number}(t::Type{T}, zdt::ZonedDateTime)
+    convert(t, Dates.datetime2unix(utc(zdt)))
 end
 
 function unix2zdt(seconds::Integer)
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -50,12 +50,12 @@ function zdt2unix(zdt::ZonedDateTime)
     Dates.datetime2unix(utc(zdt))
 end
 
-function zdt2unix{T<:Integer}(t::Type{T}, zdt::ZonedDateTime)
-    floor(t, Dates.datetime2unix(utc(zdt)))
+function zdt2unix{T<:Integer}(::Type{T}, zdt::ZonedDateTime)
+    floor(T, Dates.datetime2unix(utc(zdt)))
 end
 
-function zdt2unix{T<:Number}(t::Type{T}, zdt::ZonedDateTime)
-    convert(t, Dates.datetime2unix(utc(zdt)))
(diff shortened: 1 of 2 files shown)


src/conversions.jl
```diff
@@ -42,3 +45,15 @@ end
 function astimezone(zdt::ZonedDateTime, tz::FixedTimeZone)
     return ZonedDateTime(zdt.utc_datetime, tz, tz)
 end
+
+function zdt2unix(zdt::ZonedDateTime)
+    Dates.datetime2unix(utc(zdt))
+end
+
+function zdt2unix{T<:Number}(t::Type{T}, zdt::ZonedDateTime)
+    floor(t, Dates.datetime2unix(utc(zdt)))
```
omus: Should probably only `floor` when using Integers.

spurll:
Good catch. I'll fix that dispatch on Monday.

src/conversions.jl
spurll: Fixed in 363027a.

test/conversions.jl
```diff
@@ -28,3 +28,27 @@ zdt_warsaw = ZonedDateTime(dt, warsaw; from_utc=true)
...
 @test astimezone(zdt_warsaw, utc) === zdt_utc
+
+# ZonedDateTime to Unix timestamp (and vice versa)
+@test TimeZones.zdt2unix(ZonedDateTime(1970, utc)) == 0
+@test TimeZones.unix2zdt(0) == ZonedDateTime(1970, utc)
+
+no_dst = DateTime(2013, 2, 13)
+no_dst_zdt = ZonedDateTime(no_dst, warsaw)
+no_dst_offset = 3600
+@test TimeZones.zdt2unix(no_dst_zdt) == datetime2unix(no_dst) - no_dst_offset
```
omus: Would it be better to simplify these tests as:
`@test TimeZones.zdt2unix(no_dst_zdt) == datetime2unix(TimeZones.utc(no_dst_zdt))`

test/conversions.jl
```diff
@@ -28,3 +28,27 @@ zdt_warsaw = ZonedDateTime(dt, warsaw; from_utc=true)
...
+
+dst = DateTime(2016, 8, 11)
+dst_zdt = ZonedDateTime(dst, warsaw)
+dst_offset = 7200
+@test TimeZones.zdt2unix(dst_zdt) == datetime2unix(dst) - dst_offset
+
+@test typeof(TimeZones.zdt2unix(ZonedDateTime(1970, utc))) == Float64
+@test typeof(TimeZones.zdt2unix(Float32, ZonedDateTime(1970, utc))) == Float32
+@test typeof(TimeZones.zdt2unix(Int64, ZonedDateTime(1970, utc))) == Int64
+@test typeof(TimeZones.zdt2unix(Int32, ZonedDateTime(1970, utc))) == Int32
```
omus: Maybe change to use `isa`?

src/conversions.jl
```diff
@@ -42,3 +45,19 @@ end
...
     return ZonedDateTime(zdt.utc_datetime, tz, tz)
 end
+
+function zdt2unix(zdt::ZonedDateTime)
+    Dates.datetime2unix(utc(zdt))
+end
+
+function zdt2unix{T<:Integer}(t::Type{T}, zdt::ZonedDateTime)
+    floor(t, Dates.datetime2unix(utc(zdt)))
+end
```
omus: Consider changing to:
```julia
function zdt2unix{T<:Integer}(::Type{T}, zdt::ZonedDateTime)
    floor(T, Dates.datetime2unix(utc(zdt)))
end
```

test/conversions.jl
spurll: I'm torn. I don't like tests that just regurgitate the internals of a function (`datetime2unix(utc(zdt))` is what `zdt2unix` does). I like having a more external/manual check (like defining what the offset is) better. But your call.

omus: Good point. That modification makes the test pretty much exactly the function call. My concern is that doing a manual adjustment like this isn't a great practise. I'll try to think of an alternative way of doing this.

src/conversions.jl
spurll: Good call. I didn't realise that was possible.

test/conversions.jl
omus: What about this? You can drop the `for` loop if you prefer. The advantage of this approach is that we avoid hardcoding the offset.
```julia
warsaw = TimeZone("Europe/Warsaw")
for dt in (DateTime(2013, 2, 13), DateTime(2016, 8, 11))
	zdt = ZonedDateTime(dt, warsaw)
	offset = TimeZones.value(zdt.zone.offset)  # Total offset in seconds
	@test TimeZones.zdt2unix(zdt) == datetime2unix(dt) - offset
end
```

spurll: Sold.

test/conversions.jl
```diff
@@ -28,3 +28,23 @@ zdt_warsaw = ZonedDateTime(dt, warsaw; from_utc=true)
...
+
+# ZonedDateTime to Unix timestamp (and vice versa)
+@test TimeZones.zdt2unix(ZonedDateTime(1970, utc)) == 0
+@test TimeZones.unix2zdt(0) == ZonedDateTime(1970, utc)
+
+for dt in (DateTime(2013, 2, 13), DateTime(2016, 8, 11))
+	zdt = ZonedDateTime(dt, warsaw)
+	offset = TimeZones.value(zdt.zone.offset)   # Total offset in seconds
+	@test TimeZones.zdt2unix(zdt) == datetime2unix(dt) - offset
+end
```
omus: Totally my fault. Can we switch this to use spaces instead of tabs?

Issue #44, opened 2017-02-21
io.slotparse, io.slotformat fail on v0.6-dev


JeffreySarnoff:
ERROR: LoadError: LoadError: TypeError: Type{...} expression: expected UnionAll, got Type{Slot}
Stacktrace:
 [1] include_from_node1(::String) at ./loading.jl:539
 [2] include(::String) at ./sysimg.jl:14
 [3] include_from_node1(::String) at ./loading.jl:539
 [4] eval(::Module, ::Any) at ./boot.jl:235
 [5] require(::Symbol) at ./loading.jl:453
while loading /home/jas/.julia/v0.6/TimeZones/src/io.jl, in expression starting on line 61
while loading /home/jas/.julia/v0.6/TimeZones/src/TimeZones.jl, in expression starting on line 51

omus:
Thanks for reporting the issue. As part of https://github.com/JuliaLang/julia/pull/19545 the internals of the DateTime parsing and formatting changed. I have some work to do before TimeZones.jl will be functional on 0.6 but I could push out a version without formatting and parsing in the mean time if need be.

JeffreySarnoff:
If that would be much work and not move this forward for 0.6 then, don't. OTOH ... (I posted the issue after playing around with timestamp functions for v0.6).

omus:
I've been actively working on this. Basically the hold up with Julia 0.6 is that the fast parsing code no longer supports extending DateFormat. I've made a [PR which addresses this in base Julia](https://github.com/JuliaLang/julia/pull/20952).

JeffreySarnoff:
It looks like the i686 build just ran out of time and did not run into difficulty.

omus:
TimeZones now works with the latest Julia nightly build.

JeffreySarnoff:
I got the new nightly.
Version 0.6.0-pre.alpha.167 (2017-03-17 17:34 UTC) 
Commit 762b943 (0 days old master)
x86_64-pc-linux-gnu

did Pkg.rm("TimeZones") twice, then
Pkg.add("TimeZones")
# ...
WARNING: could not import Dates.slotparse into TimeZones
WARNING: could not import Dates.slotformat into TimeZones
ERROR: LoadError: LoadError: TypeError: Type{...} expression: expected UnionAll, got Type{Slot}
Stacktrace:
 [1] include_from_node1(::String) at ./loading.jl:539
 [2] include(::String) at ./sysimg.jl:14
[...]

PR #50, opened 2017-03-16
Compatibility with Julia 0.6

omus:


omus:
Fixes #44

REQUIRE
```diff
@@ -1,5 +1,5 @@
 julia 0.4
-Compat 0.8
+Compat 0.9.4
```
tkelman: 0.9.5 for `take!`

omus: Thanks

omus:
I have one more commit to push which utilizes the changes in https://github.com/JuliaLang/julia/pull/20952. 

The commit "Temporarily restrict parsing pre-Julia 0.6" should work on 0.6 as a version of TimeZones.jl without parsing support. Currently it's failing because `Dates.parse(::AbstractString, ::DateFormat)` is missing which is required for compiling the tzdata.

omus:
Testing commit which utilizes changes in https://github.com/JuliaLang/julia/pull/20952.

I'm not sure if the latest nightly includes the changes yet.

PR #51, opened 2017-03-17
Explicit Julia 0.6 CI tests

omus:
Currently failing as Julia Version 0.6.0-pre.alpha.0 is missing some required features.


omus:
Looks like the beta version has the update I need. Triggering rebuild.

omus:
Julia 0.6.0-pre.beta.0 works as expected.

PR #53, opened 2017-03-20
Avoid overriding the windowsZone.xml file

omus:
On Windows downloading a new windowsZone.xml file would override the committed the committed file causing the package to be dirty. Code has been updated to download a new windowsZone.xml file to a temporary location and we can fall back to the committed file.

Originally pointed out by @tkelman https://github.com/JuliaLang/METADATA.jl/pull/8378#issuecomment-287478785


tkelman:
This is going to keep downloading new copies every time Pkg.build runs, and never clean them up

PR #58, opened 2017-04-04
Improve build tools

### LINKED ISSUE #55: Decrease load on unicode.org
Last couple of month we (AppVeyor.com) become blocked/throttled by unicode.org. We use single public IP, so this affects all users. We contacted unicode.org people and they told that this is because excessive use of their server from our public IP. Also they said that their source repository is not intended for access by production systems. I searched GitHub and came across [this commit](https://github.com/JuliaTime/TimeZones.jl/commit/8f4b227a168877a70430bdc47c40037ac617a133)

If you could default to cached XML and request repo directly more rarely (or from different machine) just to refresh the cache, this might help us.

omus:
Numerous improvements around the TimeZones.jl build process including:
- Move `build.jl` code into the module
- Allowing users to optionally specify what version of tzdata to use
- Reduce requests to IANA and Unicode servers

Additional work to complete:
- [x] ~Periodically allow downloads of `windowsZones.xml` (currently requires force)~ (#62)
- [x] ~Support retrieval of older versions of `windowsZones.xml`. There may be incompatibilities between old tzdata versions using the latest version~ (#62)
- [x] ~Avoid using `Ref` for keeping track of latest tzdata version and timestamp~
- [x] Docstring for `TimeZones.build`
- [x] Load Windows ~translation~ XML at module load
- [x] Additional tests

Fixes #55


src/tzdata/download.jl
```diff
@@ -0,0 +1,109 @@
...
+
+Downloads a tzdata archive from IANA using the specified `version` to the specified
+directory. See `tzdata_url` for details on tzdata version strings.
+"""
+function tzdata_download(version::AbstractString="latest", dir::AbstractString=tempdir())
+    now_utc = now(Dates.UTC)
+    if version == "latest"
+        v = latest_version(now_utc)
+        if !isnull(v)
+            return joinpath(dir, "tzdata$(unsafe_get(v)).tar.gz")
```
iamed2: I don't think `unsafe_get` exists on 0.5. This code path should probably be tested.

src/tzdata/download.jl
```diff
@@ -0,0 +1,109 @@
...
+        end
+    end
+
+    url = tzdata_url(version)
+    archive = Base.download(url, joinpath(dir, basename(url)))  # Overwrites the local file if any
+
+    # HTTP 404 Not Found can result in a empty file being created
+    if !isarchive(archive)
```
iamed2: Maybe differentiate between an empty file and an invalid archive file? This would also trigger on a corrupt download, right?

src/tzdata/archive.jl
```diff
@@ -0,0 +1,71 @@
+import Compat: @static, is_windows
+
+"""
```
iamed2: Document keyword argument

iamed2:
Also: is Windows CodeCov possible?

src/tzdata/download.jl
```diff
@@ -0,0 +1,109 @@
...
+    tzdata_url(version="latest") -> AbstractString
+
+Generates a HTTPS URL for the specified tzdata version. Typical version strings are
+formatted as 4-digit year followed by a lowercase ASCII letter. Available versions can be
+are listed on "ftp://ftp.iana.org/tz/releases/" which start with "tzdata".
+
+# Examples
+```julia
+julia> tzdata_url("2017a")
+"http://www.iana.org/time-zones/repository/releases/tzdata2017a.tar.gz"
```
omus: Good catch. I need to switch this package over to Documenter yet

src/tzdata/download.jl
omus: The comment needs to be changed. If `download` encounters a 404 we will still generate a file which could be empty or contain the sites 404 page. Corrupt downloads will also raise an exception as the archive shouldn't pass tests.

deps/build.jl
```diff
@@ -1,96 +1,4 @@
...
-    end
-
-    open(WIN_TRANSLATION_FILE, "w") do fp
-        serialize(fp, translation)
-    end
-end
-
-info("Successfully processed TimeZone data")
+# ENV variable allows us to only download a single version during CI jobs
+build(get(ENV, "JULIA_TZ_VERSION", "latest"))
```
tkelman: shouldn't it be the other way around, pinned by default for reproducibility of what gets installed in the usual case, but set latest (maybe as a separate matrix job) on CI to detect problems?

iamed2: I think people want the latest tz files always

tkelman: Then make a new package release when there's a new tzdata file available. This has been the cause of repeated test failures over the last few years. Better to stick to something that's known to work correctly, and only release new versions after they've been tested.

omus: > This has been the cause of repeated test failures over the last few years

The test failures you're referring to were mainly to do with counting the number of time zone abbreviations. That problem was fixed in 2a7a307.

> Better to stick to something that's known to work correctly, and only release new versions after they've been tested.

The tests are meant to ensure that the code is working correctly and not test the the tzdata. The compilation process of converting the tz source files into serialized Julia structs also ensures that the tzdata is parsed correctly. The reason for pinning the version of tzdata for testing was to make sure we were just testing for code changes and not data changes.

End-users will want accurate time zone information which would be the latest version. It is possible that the latest version of the tzdata could include something that causes the tz compilation process to break but this has never happened to date. Additionally, with this PR I've tested the tz compilation process against older versions including 1996n which demonstrates that the tz source format is stable and unlikely to change.

tkelman: > End-users will want accurate time zone information which would be the latest version.

Then release new versions of the package when a new tzdata release comes out. We're talking at most a couple times a month? Users also want to be able to restore past versions of working code and reproduce results, which becomes unnecessarily difficult to do if you're downloading unversioned "latest" content from the internet at install time.

tkelman: It wouldn't even be all that complicated to write a script that runs on Travis on a nightly cron build to check if there's a new version of the tzdata available and automatically make a new tag each time.

iamed2: I think it's way more likely that people will want to pin an old version of the code and get the latest data.

tkelman: What would you want that for? Old code that gives different results than what it provided at the time? At least now that's selectable as an option with this environment variable, but I don't see what the use case is for making non-reproducibility the default.

omus: > I think it's way more likely that people will want to pin an old version of the code and get the latest data.

I think this scenario is likely if I end up dropping support for older versions of Julia

> Users also want to be able to restore past versions of working code and reproduce results, which becomes unnecessarily difficult to do if you're downloading unversioned "latest" content from the internet at install time.

A good point. This PR slightly helps as it reports the tzdata version being used at build time and keeps the tzdata archive around.

It definitely appears there are additional use cases I need to consider.

iamed2: @tkelman Production code that gives correct results without having to update your package dependency tree (or possibly even your Julia version)

This is essentially a dichotomy between correct and reproducible.

tkelman: it's also a question of library code vs application code. a library just needs to be correct at the time it was released, if it stops being correct due to an external resource changing, the solution is use a newer library version. applications or services that don't care about reproducibility or archival can configure the way they use libraries to depend on changing sources of external data, but a library shouldn't be forcing that choice on all users to give different results from running the same source code 6 months later.

iamed2: Hmm. One possible solution is have a separate package which only handles fetching TZ files. That way, people can pin/free that and this package independently.

tkelman: Linux distributions do mostly have the tzdata packaged in a pretty minimal data-only package, from what I've seen

src/tzdata/download.jl
omus: It doesn't exist on 0.5 but it is in Compat.jl. I added an import at the top of the file to make this clear.

src/tzdata/archive.jl
omus: Done

omus:
> Also: is Windows CodeCov possible?

Possible but not currently supported. I opened an [issue](https://github.com/JuliaCI/Coverage.jl/issues/124) for this.

omus:
Window code coverage is too much trouble for this PR. I almost got it working on Julia 0.4 but newer versions of Julia. https://github.com/JuliaLang/julia/issues/21289

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -34,30 +35,67 @@ function compile(xml_file::AbstractString, translation_file::AbstractString)
         end
     end
 
+    return translation
+end
+
+function compile(xml_file::AbstractString, translation_file::AbstractString)
     open(translation_file, "w") do fp
-        serialize(fp, translation)
+        serialize(fp, compile(xml_file))
```
tkelman: what's the motivation for using the Julia serializer here? that isn't really meant for long-term storage

omus: Serialization is used here to avoid having to re-parse the XML during the package loading. I don't need to strictly need to use serialization in this case.

I do use serialization for storing the TimeZone types. Can serialization change between Julia patch releases?

tkelman: It shouldn't change between patch releases, but I'm not sure we have good tests ensuring that. It definitely changes between minor releases, and during the course of nightlies.

deps/build.jl
omus: > One possible solution is have a separate package which only handles fetching TZ files. That way, people can pin/free that and this package independently.

I think this is probably the ideal solution. Unfortunately, I won't be able to do this work right away so this is what I'll do for the moment. For now, I'll update this line from "latest" to "2017b" to deal with the reproducibility problem. Users who want to always use the most up to date version automatically can currently set an env variable `JULIA_TZ_VERSION=latest`. I'll also be keeping a closer eye on tzdata releases so I can keep the default build version up to date.

omus:
Code is finally in a state I am more or less happy with. The code coverage drop is mostly from the introduction of more Windows specific code. PR #60 should help with that.

PR #64, opened 2017-04-10
Add tz string literal

giordano:
This is just a convenient syntax sugar, but I don't know if you planned to use `tz"..."` for something else.

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -3,7 +3,15 @@
 A `TimeZone` is an abstract type that represents information regarding a specific time zone. Typically you will create an instance of a `TimeZone` by passing in a zone name to the convenience constructor `TimeZone`:
 
 ```julia
-TimeZone("Europe/Warsaw")
+julia> TimeZone("Europe/Warsaw")
+Europe/Warsaw (UTC+1/UTC+2)
+```
+
+You can also create a `TimeZone` by using a `tz` string macro:
+
+```julia
+julia> tz"Europe/Warsaw"
+Europe/Warsaw (UTC+1/UTC+2)
 ```
 
 To see all of the [currently available](faq#why-are-the-etc-time-zones-unsupported) time zone names:
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -7,6 +7,7 @@ import Compat: is_windows
 
 export TimeZone, FixedTimeZone, VariableTimeZone, ZonedDateTime, DateTime,
     TimeError, AmbiguousTimeError, NonExistentTimeError, UnhandledTimeError,
+    @tz_str,
     # discovery.jl
     timezone_names, all_timezones, timezones_from_abbr, timezone_abbrs,
     # accessors.jl
@@ -75,6 +76,23 @@ function TimeZone(str::AbstractString)
         end
     end
 end
+
+
+"""
+    @tz_str -> TimeZone
+
+Constructs a `TimeZone` subtype based upon the string `str`.  See docstring of `TimeZone`
+for reference.
+
+```julia
+julia> TimeZone("Africa/Nairobi")
+Africa/Nairobi (UTC+3)
+
+julia> tz"Africa/Nairobi"
+Africa/Nairobi (UTC+3)
+```
+"""
+macro tz_str(str) TimeZone(str) end
 
 """
     build(version="latest", regions=REGIONS; force=false) -> Void
--- a/test/TimeZones.jl
+++ b/test/TimeZones.jl
@@ -1,5 +1,6 @@
 # Deserialization can cause us to have two immutables that are not using the same memory
 @test TimeZone("Europe/Warsaw") === TimeZone("Europe/Warsaw")
+@test tz"Africa/Nairobi" === TimeZone("Africa/Nairobi")
 
 if lowercase(get(ENV, "CI", "false")) == "true"
     info("Testing build process")
@@ -38,13 +39,17 @@ if lowercase(get(ENV, "CI", "false")) == "true"
 
     # Compile tz source files with an extended max_year. An example from the FAQ.
     warsaw = TimeZone("Europe/Warsaw")
+    windhoek = tz"Africa/Windhoek"
     @test last(warsaw.transitions).utc_datetime == DateTime(2037, 10, 25, 1)
     @test get(warsaw.cutoff) == DateTime(2038, 3, 28, 1)
     @test_throws TimeZones.UnhandledTimeError ZonedDateTime(DateTime(2039), warsaw)
 
     TimeZones.TZData.compile(max_year=2200)
     new_warsaw = TimeZone("Europe/Warsaw")
+    new_windhoek = tz"Africa/Windhoek"
 
+    @test warsaw !== new_warsaw
+    @test windhoek === new_windhoek # Instead tz".." doesn't change after re-compilation
     @test last(new_warsaw.transitions).utc_datetime == DateTime(2200, 10, 26, 1)
     @test get(new_warsaw.cutoff) == DateTime(2201, 3, 29, 1)
     ZonedDateTime(2100, new_warsaw)  # Test this doesn't throw an exception


omus:
Thanks for the PR. I assume you want this strictly for convenience?

giordano:
Yes, it's just for convenience (it saves a few key strokes ;-).  Feel free to reject it if you don't think it's useful, at least it was a few minute exercise for me :-)

test/TimeZones.jl
```diff
@@ -1,5 +1,5 @@
 # Deserialization can cause us to have two immutables that are not using the same memory
-@test TimeZone("Europe/Warsaw") === TimeZone("Europe/Warsaw")
+@test tz"Europe/Warsaw" === "Europe/Warsaw"
```
omus: I expect this will fail. I would just leave the rest as it was.

docs/types.md
```diff
@@ -3,7 +3,15 @@
 A `TimeZone` is an abstract type that represents information regarding a specific time zone. Typically you will create an instance of a `TimeZone` by passing in a zone name to the convenience constructor `TimeZone`:
 
 ```julia
-TimeZone("Europe/Warsaw")
+julia> TimeZone("Europe/Warsaw")
+Europe/Warsaw (UTC+1/UTC+2)
+```
+
+You can also create a `TimeZone` by using a `tz` string literal:
```
omus: "using the `tz` string macro"

src/TimeZones.jl
```diff
@@ -76,6 +77,22 @@ function TimeZone(str::AbstractString)
     end
 end
 
+
+"""
+    @tz_str str -> TimeZone
+
+Constructs the `TimeZone` of `str` with `TimeZone` function, which see.
```
omus: "which see"?

test/TimeZones.jl
omus: Maybe add a new test
```julia 
@test tz"Europe/Warsaw" === TimeZone("Europe/Warsaw")
```

giordano: Oops, didn't mean to remove the `TimeZone` on the right side.  Thanks

docs/types.md
giordano: I used the terminology adopted here: https://docs.julialang.org/en/latest/manual/metaprogramming.html#Non-Standard-String-Literals-1 but I don't have a strong opinion, will change following your suggestion.

test/TimeZones.jl
omus: I'd like to keep the original test too since it's directly testing for what is mentioned in the comment.

src/TimeZones.jl
giordano: This was done on purpose
* https://en.wiktionary.org/wiki/q.v.
* https://www.urbandictionary.com/tags.php?tag=which%20see

I saw used often elsewhere, but if you think it's difficult to understand, please suggest me something more common (I'm not a native English speaker).

giordano:
I fixed the test (now it passes on my machine; previously I run the tests on a different version of the patch, not the final one), changed "string literal" to "string macro" and rephrased the docstring of the macro.

giordano:
Probably it doesn't play nicely with `TimeZones.TZData.compile`?

src/TimeZones.jl
```diff
@@ -76,6 +77,25 @@ function TimeZone(str::AbstractString)
     end
 end
 
+
+"""
+    @tz_str str -> TimeZone
```
omus: After looking up the docstring for `@r_str` it appears this should just be `@tz_str -> TimeZone`

src/TimeZones.jl
```diff
@@ -76,6 +77,25 @@ function TimeZone(str::AbstractString)
...
+```julia
+julia> TimeZone("Africa/Nairobi")
+Africa/Nairobi (UTC+3)
+
+julia> tz"Africa/Nairobi"
+Africa/Nairobi (UTC+3)
+```
+"""
+macro tz_str(str)
+    return :( TimeZone($str) )
```
omus: Could you remove the whitespace padding inside the parenthesis? i.e. Change to `:(TimeZone($str))`

test/TimeZones.jl
```diff
@@ -37,13 +37,13 @@ if lowercase(get(ENV, "CI", "false")) == "true"
...
     # Compile tz source files with an extended max_year. An example from the FAQ.
-    warsaw = TimeZone("Europe/Warsaw")
+    warsaw = tz"Europe/Warsaw"
     @test last(warsaw.transitions).utc_datetime == DateTime(2037, 10, 25, 1)
     @test get(warsaw.cutoff) == DateTime(2038, 3, 28, 1)
     @test_throws TimeZones.UnhandledTimeError ZonedDateTime(DateTime(2039), warsaw)
 
     TimeZones.TZData.compile(max_year=2200)
-    new_warsaw = TimeZone("Europe/Warsaw")
+    new_warsaw = tz"Europe/Warsaw"
```
omus: Lets leave this line and the [above line](https://github.com/JuliaTime/TimeZones.jl/pull/64/files#diff-72390e967fd3f6573a6058eba616722aR40) the way the were.

src/TimeZones.jl
giordano: The docstring of `macro` itself puts spaces inside the quoting parentheses, but I will change this.

test/TimeZones.jl
giordano: Actually this test was fundamental to spot that previous macro definition wouldn't work after a `TimeZones.TZData.compile`.

omus:
> Probably it doesn't play nicely with TimeZones.TZData.compile?

Oh, I didn't catch the problem you were running into before. Since macros execute at parse time they typically don't usually take into account runtime state. I think having the `tz_str` macro evaluate the time zone at parse time is probably best since this can help improve performance when using the `tz_str` macro in a loop. This will cause the macro not to work well with the re-compiling the time zone data but I only expect `compile` to be called during: build, interactive sessions, or during tests.

Here's a quick demo that shows the performance gains of evaluating at parse time:

```julia
julia> using TimeZones, BenchmarkTools

julia> macro tza_str(str)
           TimeZone(str)
       end;

julia> macro tzb_str(str)
           :(TimeZone($str))
       end;

julia> tz = TimeZone("UTC");  # Pre-loads the time zone

julia> function demo_a()
           for i = 1:20000
               ZonedDateTime(i, 4, tza"UTC")
           end
       end;

julia> function demo_b()
           for i = 1:20000
               ZonedDateTime(i, 4, tzb"UTC")
           end
       end;

julia> a = @benchmark demo_a()

BenchmarkTools.Trial: 
  samples:          7876
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  625.00 kb
  allocs estimate:  20000
  minimum time:     482.29 μs (0.00% GC)
  median time:      503.79 μs (0.00% GC)
  mean time:        634.22 μs (7.66% GC)
  maximum time:     55.14 ms (0.00% GC)

julia> b = @benchmark demo_b()
BenchmarkTools.Trial: 
  samples:          1102
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  1.22 mb
  allocs estimate:  60000
  minimum time:     3.74 ms (0.00% GC)
  median time:      4.13 ms (0.00% GC)
  mean time:        4.54 ms (2.71% GC)
  maximum time:     13.79 ms (0.00% GC)

julia> judge(median(a), median(b))
BenchmarkTools.TrialJudgement: 
  time:   -87.80% => improvement (5.00% tolerance)
  memory: -50.00% => improvement (1.00% tolerance)
```

giordano:
`tzb_str` macro isn't slower than `TimeZone` (it's probably very slightly faster, but nothing really noticeable).  `tza_str` is faster but doesn't allow for changing state.  Personally I'd go for a solution that works in any case (and isn't bad performance-wise), but the decision is up to you ;-)  Just tell me what to do.

omus:
Lets go with the `tza_str` version. Once that is changed and the tests are updated I think we're good.

giordano:
Ok, macro definition switched back to `macro tz_str(str) TimeZone(str) end`.  Now the test checks that a `TimeZone` constructed with `tz"..."` doesn't change after re-compilation, so you'll be warned if this behavior will change in the future.

PR #69, opened 2017-04-29
explicitly specify 7z

musm:


musm:
https://github.com/JuliaLang/julia/pull/21540

src/tzdata/archive.jl
```diff
@@ -1,5 +1,9 @@
 import Compat: @static, is_windows
 
+if is_windows()
+    const exe7z = joinpath(JULIA_HOME, "7z.exe")
```
omus: Can you rename this to `7Z`?

musm: unfortunately not, can't start with a number `7Z = 7*Z`

omus: Ah, that slipped my mind.

PR #70, opened 2017-05-02
Base supports isless between all Period subtypes

omus:
See: https://github.com/JuliaLang/julia/pull/21378. Special thanks to @tkelman


tkelman:
this could also go in Compat now if you're so inclined

PR #71, opened 2017-05-03
Fix TZ data parsing for Julia 0.6

joshbode:
Julia 0.6 does not provide a function to split a date-time into components.
Parses the string to a `DateTime` to produce:

    [Year, Month, Day, Hour, Minute, Second]

Note: the case of 24:00 is handled prior to parsing to avoid parse failure.


joshbode:
Actually, I think this might actually be fixed in `v0.6.0-pre.beta`
This was developed using `v0.6.0-pre.alpha` on which `Pkg.build("TimeZones")` was failing.

PR #73, opened 2017-05-17
Add TimeError docstrings

omus:


omus:
I made a mistake when separating the exceptions into a new file. Will be sure to squash this PR.

PR #80, opened 2017-07-14
Update docs to use tz string macro

omus:
Most people will probably want to use the `tz` string macro because it's shorter. I've updated the documentation to reflect typical usage.


omus:
cc: @iamed2

PR #85, opened 2017-08-31
fixes v0.7 warnings

JeffreySarnoff:
I may have got to all of them, maybe not -- the edits are unchecked, so there is the possibility of typos.
I understand that you have clients who need the v0.5 compat, so this should be given a separate tag and should go with code that gets a non-v0.5 compat version number

### DIFF
--- a/src/adjusters.jl
+++ b/src/adjusters.jl
@@ -5,7 +5,6 @@ import Base.Dates: firstdayofweek, lastdayofweek, firstdayofmonth, lastdayofmont
 
 # Truncation
 # TODO: Just utilize floor code for truncation?
-
 function trunc(zdt::ZonedDateTime, ::Type{P}) where P<:DatePeriod
     ZonedDateTime(trunc(localtime(zdt), P), timezone(zdt))
 end
--- a/src/adjusters.jl
+++ b/src/adjusters.jl
@@ -5,10 +5,11 @@ import Base.Dates: firstdayofweek, lastdayofweek, firstdayofmonth, lastdayofmont
 
 # Truncation
 # TODO: Just utilize floor code for truncation?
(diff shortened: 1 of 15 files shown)


src/tzdata/archive.jl
```diff
@@ -1,6 +1,6 @@
-import Compat: @static, is_windows
+import Compat: @static
```
omus: You should import `Sys` from Compat for this to work on 0.6

src/adjusters.jl
```diff
@@ -5,10 +5,11 @@ import Base.Dates: firstdayofweek, lastdayofweek, firstdayofmonth, lastdayofmont
 
 # Truncation
 # TODO: Just utilize floor code for truncation?
-function trunc{P<:DatePeriod}(zdt::ZonedDateTime, ::Type{P})
+
```
omus: Extra blank line

JeffreySarnoff:
I made the requested changes.  I do not know how to get them to be part of the active PR.

src/local.jl
```diff
@@ -1,9 +1,9 @@
 # Determine the local system's time zone
 # Based upon Python's tzlocal https://pypi.python.org/pypi/tzlocal
-import Compat: @static, is_apple, is_unix, is_windows, readstring
+import Compat: @static, readstring
```
omus: Need to import `Sys`

JeffreySarnoff: dond

JeffreySarnoff:
added missing () to Sys.iswindows (should fix checks)

JeffreySarnoff:
the current error in testing is on linux -- it is not finding the TZ environment variable set,
the relevant code appears unchanged .. but I cannot test it on linux today
any thoughts?

omus:
I've squashed your changes and included them in #88. Ideally will be merging today if we can get the CI to cooperate.

PR #87, opened 2017-09-03
Add support for TZ format `std offset`

omus:
Adding support for parsing the TZ environmental variable format which includes the standard offset name and value.

The reason for this change is to support `TZ="UTC"` which is the new way that Travis CI indicates the time zone in a Linux environment. Additionally, when a TZ env var cannot be processed it is now included in the error message which should help with debugging unsupported TZ formats.


omus:
Coverage seems messed up.

PR #88, opened 2017-09-06
Update to Julia 0.7 syntax

omus:
Drops support for Julia 0.4 and 0.5. Replaces #85

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ os:
   - osx
   - linux
 julia:
-  - "0.4"
-  - "0.5"
   - "0.6"
   - nightly
 notifications:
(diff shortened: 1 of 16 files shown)


omus:
[AppVeyor failures](https://ci.appveyor.com/project/omus/timezones-jl/build/1.0.347/job/67k9precw7kyjdt4) appear to be https://github.com/JuliaPackaging/WinRPM.jl/issues/57

JeffreySarnoff:
why would that failure occur here and not before these edits?

omus:
> why would that failure occur here and not before these edits?

The issue had to do with a package server being down. Looks like it's working again and re-running the failed jobs now works.

JeffreySarnoff:
yar

PR #89, opened 2017-09-07
Avoid invalid use of SystemError

omus:
Creating a partial `SystemError` struct will result in an `UndefRefError` when attempting to display the exception:

```julia
julia> throw(SystemError("demo"))
ERROR: SYSTEM: show(lasterr) caused an error
UndefRefError()
```

### DIFF
--- a/src/local.jl
+++ b/src/local.jl
@@ -60,7 +60,7 @@ function localzone()
                     end
                 end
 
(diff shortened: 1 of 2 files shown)


omus:
Travis is already passing under Linux. Since these changes only effect Linux I'll merge these right away.

PR #92, opened 2017-10-31
Use 2017c tzdata

omus:


omus:
I'm uncomfortable that this change is completely untested because I'm setting `JULIA_TZ_VERSION` in the CI configuration files. I think I'll introduce an additional test step that disables `JULIA_TZ_VERSION` if the `deps/build.jl` file has been modified.

omus:
My main concern is that if I made a typo, say `2O17c` it wouldn't be caught until a user reported it.

omus:
Note that the windowsZones.xml file had minor changes but I think since it was previously ignored it appears to be an entirely new file.

omus:
Updated the "windowsZones.xml" file with: `download("http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml", Pkg.dir("TimeZones", "deps", "local", "windowsZones.xml"))`

PR #93, opened 2017-11-02
Fix verification of tzdata default

omus:
I messed this up the first time in #92 as the bash command would only have non-zero return code if deps/build.jl was changed. Additionally, I managed to drop the `grep` statement.

I'm going to make two additional commits which should test the desired behaviour but will disappear once this PR is squashed.


omus:
My final test didn't work as intended as diff-wise the deps/build.jl file was no longer changed. Since this feature was manually tested I think this is good enough. I'll try to keep my eyes out on the next tzdata update though.

PR #97, opened 2017-11-12
Update README.md

scls19fr:
Fix RTD link
similar to https://github.com/JuliaTime/TimeZones.jl/pull/96


omus:
Looks good. I found some other RTD links which I'll update in another PR.

PR #98, opened 2017-11-14
Update references to Julia documentation

omus:
Updates old references to the old readthedocs documentation for Julia. A continuation of the work done in #97.

### DIFF
--- a/docs/conversions.md
+++ b/docs/conversions.md
@@ -12,7 +12,7 @@ julia> astimezone(zdt, tz"Asia/Tokyo")
 
 ## Parsing strings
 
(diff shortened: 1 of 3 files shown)


docs/conversions.md
```diff
@@ -45,7 +45,7 @@ julia> ZonedDateTime("2015-03-30+02:00", df)
 
 ## Formatting strings
 
-Formatting a `ZonedDateTime` as a string also extends the functionality provided by `Base.Dates`. The `TimeZones` package adds the new formatting character codes `z` and `Z` to the list of available [formatting character codes](http://julia.readthedocs.io/en/latest/stdlib/dates/#man-date-formatting):
+Formatting a `ZonedDateTime` as a string also extends the functionality provided by `Base.Dates`. The `TimeZones` package adds the new formatting character codes `z` and `Z` to the list of available [formatting character codes](https://docs.julialang.org/en/stable/stdlib/dates/#Base.Dates.format-Tuple{Base.Dates.TimeType,AbstractString}):
```
omus: Change requires that https://github.com/JuliaLang/julia/pull/24607 is backported to Julia 0.6.

omus: I was considering making just this part a separate PR but this link will work as soon as the docs back ported and built. Seems like this is no worse than the current dead link.

PR #99, opened 2017-11-15
Update to Mocking 0.4.1

omus:
Update terminology to match Mocking.


omus:
CI failures from LightXML

PR #100, opened 2017-11-15
Implement methods for `today` and `todayat`.

omus:
Implements `today(::TimeZone)` which provides the current date in the specified `TimeZone`. Additionally, `today_at(::Time, ::TimeZone)` was introduced which works similarly to `::Date + ::Time == ::DateTime` but instead returns a `ZonedDateTime`.

### DIFF
--- a/docs/current.md
+++ b/docs/current.md
@@ -1,6 +1,6 @@
 ## Current Time / System Time Zone
 
-Julia provides the `now()` method to retrieve your current system's time as a `DateTime`. The TimeZones.jl package provides an additional method to the `now` function providing the current time as a `ZonedDateTime`:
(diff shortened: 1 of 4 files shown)


omus:
I need to think about this more yet. I think it may make sense to implement an `at` as proposed in: https://github.com/JuliaLang/julia/issues/24814

omus:
After looking at this again I think `at` gets a little to close to the functionality of the constructors. I do think there should be an easy way create a new `TimeType` by setting the `Time` but I think that may be outside the scope of this PR.

rofinn:
This looks fine for now. I'd suggest removing the '_' in `today_at` as none of the other methods appear to use that convention.

omus:
Thanks @rofinn. I expect `todayat` won't exist for a long time. I think this reveals that the design needs some additional work which I wrote up in #121

PR #106, opened 2017-12-15
Add assistance to no active version error

omus:
Improves on the error message when no active tzdata version is available. Error can occur on a manual installation of the package.


omus:
Coverage was decreased which means this error isn't currently being tested.

omus:
I expect 0.7 to fail but for reasons not due to this change.

PR #107, opened 2017-12-15
Fixes support for `tryparse` with `ZonedDateTime`

omus:
Since `ZonedDateTime` parsing code heavily relies on `Dates` from the standard library I originally did not write tests for this logic. I have now added a few tests to ensure the basics work and additionally have:

* Added `istimezone` function which can determine if a string is valid for `TimeZone`
* Added a `default_format(::Type{ZonedDateTime})` method to extend `Dates.default_format`
* Cleaned up a fallback function for `parse_components`


omus:
cc: @iamed2

test/parse.jl
```diff
@@ -0,0 +1,43 @@
...
+        [
+            Dates.Year(2017),
+            Dates.Month(11),
+            Dates.Day(14),
+            Dates.Hour(11),
+            Dates.Minute(3),
+            Dates.Second(53),
+            tz"UTC+01",
+        ]
+    )
```
iamed2: The formatting of this block makes me sad

omus: lol, that is fair. I don't love it myself. I think I may just save the expected value to a variable first. Any suggestions?

iamed2: That, and not using `==` in function form. That's probably it.

omus:
Should be fine to merge as I just refactored the `==()` test. I expect 0.7 to currently fail.

PR #109, opened 2017-12-21
Allow constructing a FixedTimeZone from Dates.UTC and allow passing it to ZonedDateTime

iamed2:


test/conversions.jl
```diff
@@ -3,6 +3,11 @@ import Compat.Dates
 utc = FixedTimeZone("UTC")
 warsaw = resolve("Europe/Warsaw", tzdata["europe"]...)
 
+# Converting from the stdlib UTC
+@test FixedTimeZone(UTC()) == utc
```
omus: I'm not sure I want to support using an instance of `UTC` as this is never actually utilized in Base. In fact they only ever seem to use the type so maybe the type should be redefined as abstract or rethought.

iamed2: Your interface currently says it accepts any `TimeZone` :P

omus: Fair. This happens to be the only `TimeZone` I don't like 😆

omus:
@iamed2 is this something you actually need? Dates only needs this for `now(::Type{UTC})` and it's never used as an instance.

iamed2:
It's not something I need but I do keep running into this when trying to create a UTC ZDT and using the wrong UTC (there are so many)

PR #110, opened 2017-12-21
Add typemin and typemax

iamed2:
Uses the definitions from stdlib Dates

### DIFF
--- a/test/types.jl
+++ b/test/types.jl
@@ -384,3 +384,7 @@ end
 # Issue #52
 dt = now()
 @test_throws ErrorException ZonedDateTime(dt, warsaw) > dt
+
+# type extrema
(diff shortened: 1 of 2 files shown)


test/types.jl
```diff
@@ -379,3 +379,7 @@ end
 # Issue #52
 dt = now()
 @test_throws ErrorException ZonedDateTime(dt, warsaw) > dt
+
+# type extrema
+@test typemin(ZonedDateTime) <= ZonedDateTime(typemin(DateTime), utc)
+@test typemax(ZonedDateTime) >= ZonedDateTime(typemax(DateTime), utc)
```
omus: Won't these always  be `==`?

iamed2: Given the current definition, yes. But you could conceivably change the definition to restrict the interval to when time zones exist, for example.

omus:
Let's merge this thing and go home!

PR #111, opened 2017-12-21
Allow parsing of two-digit numeric time zone

omus:
Previously was disallowed just when constructing a `FixedTimeZone`. I may want to stop this from being allowed in the constructor but it definitely should be allowed for parsing.


omus:
cc: @iamed2

omus:
Julia 0.7 failures are currently expected. I'll merge this as I'm quite confident that the macOS runners will pass whenever they actually run.

PR #112, opened 2017-12-22
Use the Printf stdlib package on 0.7

ararslan:


omus:
Thanks for the PR

ararslan:
No problem! Looks like there are still some issues with `Nullable`. Now that Compat has `Some` and `Nothing`, perhaps it would be a good time to migrate away from `Nullable`s?

omus:
I am concerned about switching to a `Union` type while 0.6 is currently the release.

Do you happen to know the best practice when the value cannot contain `nothing`. Should `Nullable{DateTime}` be converted to `Union{DateTime,Nothing}` or `Union{Some{DateTime},Nothing}`?

ararslan:
> I am concerned about switching to a `Union` type while 0.6 is currently the release.

That's certainly a fair concern. Perhaps better to figure out the current issues with `Nullable` then.

> Do you happen to know the best practice when the value cannot contain `nothing`.

If I understand correctly, in that case you would use `Union{DateTime,Nothing}`. `Some` is useful when the value it wraps _can_ be `nothing`, in which case `Some(nothing)` indicates that `nothing` was expected.

omus:
Sorry for the hold up on this. There is currently an issue with `__init__` on 0.7 which is stopping TimeZones from extending the parsing code in Dates.

omus:
I was hoping to get everything working before merging stuff but I now thing merging partial fixes is the way to go.

ararslan:
Okay, sounds good. Thanks!

PR #114, opened 2018-01-08
Fix errors for parsing tests on julia 0.7

nicoleepp:
Most of the errors were caused by the expulsion of Nullables from Base (and hence how Base parses TimeZone strings).

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -66,12 +66,7 @@ string is assumed to be a static time zone.
 A list of recognized time zones names is available from `timezone_names()`. Supported static
 time zone string formats can be found in `FixedTimeZone(::AbstractString)`.
 """
-function TimeZone(str::Union{AbstractString, Nullable{String}})
-    if str isa Nullable
-        str = isnull(str) ? throw(ArgumentError("Unknown time zone \"$str\"")) : get(str)
-        str = convert(AbstractString, str)
-    end
-
+function TimeZone(str::AbstractString)
     return get!(TIME_ZONES, str) do
         if ismatch(FIXED_TIME_ZONE_REGEX, str)
             return FixedTimeZone(str)
@@ -84,6 +79,17 @@ function TimeZone(str::Union{AbstractString, Nullable{String}})
             return deserialize(fp)
         end
     end
+end
+
+"""
+    TimeZone(str::Nullable{String}) -> TimeZone
+
+Constructs a `TimeZone` subtype based upon the string. If the string is null throws an
+ArgumentError.
+"""
+function TimeZone(str::Nullable{String})
+    isnull(str) && throw(ArgumentError("Unknown time zone \"$str\""))
+    TimeZone(get(str))
 end
 
 """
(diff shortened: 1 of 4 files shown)


src/TimeZones.jl
```diff
@@ -66,7 +66,12 @@ string is assumed to be a static time zone.
 A list of recognized time zones names is available from `timezone_names()`. Supported static
 time zone string formats can be found in `FixedTimeZone(::AbstractString)`.
 """
-function TimeZone(str::AbstractString)
+function TimeZone(str::Union{AbstractString, Nullable{String}})
+    if str isa Nullable
+        str = isnull(str) ? throw(ArgumentError("Unknown time zone \"$str\"")) : get(str)
+        str = convert(AbstractString, str)
+    end
+
```
omus: Can you handle this with a new method?
```julia
function TimeZone(str::Nullable{String})
    isnull(str) && throw(ArgumentError("Unknown time zone \"$str\"")
    TimeZone(get(str))
end
```

src/parse.jl
```diff
@@ -29,7 +29,7 @@ function tryparsenext_fixedtz(str, i, len, min_width::Int=1, max_width::Int=0)
         return Nullable{String}(), i
     else
         tz = SubString(str, tz_start, tz_end)
-        return Nullable{String}(tz), i
+        return Nullable{String}(convert(String, tz)), i
```
omus: Hmm, this is explicitly required? Can we get away with `Nullable{String}(String(tz))`?

test/parse.jl
```diff
@@ -18,14 +18,25 @@ import Compat.Dates: parse_components, default_format
...
+    else
+        @test isequal(
+            tryparse(ZonedDateTime, "2013-03-20 11:00:00+04:00", dateformat"y-m-d H:M:SSz"),
+            ZonedDateTime(2013, 3, 20, 11, tz"UTC+04"),
+        )
+        @test isequal(
+            tryparse(ZonedDateTime, "2016-04-11 08:00 EST", dateformat"yyyy-mm-dd HH:MM zzz"),
+            nothing,
+        )
+    end
```
omus: There would be less duplicated code if you did something like this:
```julia
zdt = ZonedDateTime(2013, 3, 20, 11, tz"UTC+04")
@test isequal(
    tryparse(ZonedDateTime, "2013-03-20 11:00:00+04:00", dateformat"y-m-d H:M:SSz"),
    VERSION < v"0.7.0-DEV.3017" ? Nullable(zdt) : zdt,
)
```

src/parse.jl
nicoleepp: Yup

omus:
I think this is good. I'll try to get the Mocking 0.7 issue addressed before merging.

omus:
Waiting on https://github.com/invenia/Mocking.jl/pull/34

omus:
@nicoleepp can you rebase these changes? It looks like there are still some problems remaining so I'll probably end up merging this to get some of the issues resolved on 0.7.

nicoleepp:
@omus Will do

omus:
Thank you @nicoleepp. I took a further look into this and on 0.7 `tryparsenext` functions are expected to return `nothing` instead of `Nullable()`. Making this change ended up cleaning up some parts of this PR.

My plan is to merge this PR as is and have a follow up PR which does the cleanup. Thank you very much for your contribution.

PR #115, opened 2018-01-12
Julia 0.7 Date parsing changes

omus:
Refactors some of the changes introduced in https://github.com/JuliaTime/TimeZones.jl/pull/114

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -1,6 +1,13 @@
 import Compat.Dates: DateFormat, DatePart, tryparsenext, format, min_width, max_width,
     default_format
 
+# Handle Nullable deprecation on Julia 0.7
+if VERSION < v"0.7.0-DEV.3017"
+    nullable(T::Type, x) = Nullable{T}(x)
+else
+    nullable(T::Type, x) = x  # Ignore nullable container type T
+end
+
 function tryparsenext_fixedtz(str, i, len, min_width::Int=1, max_width::Int=0)
     tz_start, tz_end = i, 0
     min_pos = min_width <= 0 ? i : i + min_width - 1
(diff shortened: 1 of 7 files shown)


test/ranges.jl
```diff
@@ -97,7 +97,7 @@ range = ZonedDateTime(2015, 11, 1, dst):Dates.Hour(1):ZonedDateTime(2015, 11, 3,
 if VERSION < v"0.7.0-DEV.2778"
     range = ZonedDateTime(2017, 10, 1, 9, utc):ZonedDateTime(2017, 12, 8, 23, utc)
     @test step(range) == Dates.Day(1)
-else
+elseif VERSION < v"0.7-DEV+"  # Currently failing on Julia 0.7-DEV. Disabling for now.
```
omus: I believe this has something to do with the logging changes in Julia. Since this test isn't critical it seems safe to disable it during the dev cycle.

omus:
My local Mac system required https://github.com/JuliaArchive/Nullables.jl/pull/10 which is included in [Nullables.jl v0.0.3](https://github.com/JuliaArchive/Nullables.jl/releases/tag/v0.0.3) which isn't quite released yet: https://github.com/JuliaLang/METADATA.jl/pull/12845

src/parse.jl
```diff
@@ -1,6 +1,13 @@
 import Compat.Dates: DateFormat, DatePart, tryparsenext, format, min_width, max_width,
     default_format
 
+# Handle Nullable deprecation on Julia 0.7
+if VERSION < v"0.7.0-DEV.3017"
+    nullable(T::Type, x) = Nullable{T}(x)
+else
+    nullable(T::Type, x) = x  # Ignore nullable container type T
+end
```
ararslan: I'm confused by this. Could you not use `Nullable` from the Nullables package, which is already a dependency, or switch to `Some`, which is available in Compat?

omus: When `Nullable` was removed from Base the parsing interface in Dates was also changed to no longer use `Nullable`. Specifically the `tryparsenext` interface now expects a `Tuple{Union{T, Nothing}, Int}` where previously a `Tuple{Nullable{T}, Int}`.

Since I am hooking into the code provided by Dates I need to provide exactly what that interface requires.

ararslan: Okay, thanks for the explanation.

omus:
Coverage difference looks wrong to me. I think I'll bump the required version of Nullables.jl to v0.0.3 to be safe and merge this.

PR #119, opened 2018-01-23
Default to using 2018b tzdata

omus:
2018b was released 2018-01-17.


omus:
Note that 2018a tzdata release includes a [negative daylight saving time offset](http://mm.icann.org/pipermail/tz/2018-January/025820.html) which (I believe) is the first time this has occurred. There seems to be [a few time zone libraries which are struggling](http://mm.icann.org/pipermail/tz/2018-January/025908.html) with this.

PR #122, opened 2018-02-09
Default to using 2018c tzdata

omus:
Release notes: http://mm.icann.org/pipermail/tz-announce/2018-January/000048.html


omus:
CI failures are unrelated network issues.

PR #124, opened 2018-02-09
Extend functions using module in declaration instead of import

omus:
I decided to stick with `import` for `+` and `-` as I find this reads nicer.


omus:
Replaced by: https://github.com/JuliaTime/TimeZones.jl/pull/167

PR #126, opened 2018-03-09
Implement show_next_transition

omus:
Add a human friendly function which displays information about the next time zone transition. For example:
```julia
julia> show_next_transition(tz"America/New_York")
Transition Date:   2018-03-11
Local Time Change: 02:00 → 03:00 (Forward)
Offset Change:     UTC-5/+0 → UTC-5/+1
Transition From:   2018-03-11T01:59:59.999-05:00 (EST)
Transition To:     2018-03-11T03:00:00.000-04:00 (EDT)
```


omus:
Note the function doesn't work with `ZonedDateTime`s using`FixedTimeZone`s. Not sure this needs to be addressed.

omus:
Alright, I should be done with this now. I can't think of any other time zone transition information that would be useful to see.

omus:
Working on this has given me some ideas for how to support `LaxZonedDateTime` behaviour with just using the standard `ZonedDateTime`. I'll open an issue about that.

spurll:
This looks great!

PR #136, opened 2018-05-08
Default to using 2018e tzdata

omus:
Release notes: http://mm.icann.org/pipermail/tz-announce/2018-May/000050.html

Updated windowsZones.xml file (Last-Modified: Fri, 04 May 2018 08:13:42 GMT):

```bash
cd deps/local
curl -vR http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml -o windowsZones2018e.xml
cp -p windowsZones2018e.xml windowsZones.xml
```


omus:
Failures only on Julia 0.7 where broadcast syntax needs to be updated.

PR #137, opened 2018-05-17
Make ZonedDateTime subtype AbstractDateTime

ararslan:
On Julia 0.6, `Compat.AbstractDateTime` is an alias for `TimeType`, so the behavior on 0.6 will not change. On 0.7, this is a better choice for the supertype and should fix issues noted in https://github.com/invenia/Intervals.jl/pull/30. As suggested by @rofinn.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -2,6 +2,7 @@
 # import Compat.Dates: UTInstant, DateTime, TimeZone, Millisecond
 using Compat
 using Compat.Dates
+using Compat: AbstractDateTime
 import Compat.Dates: value, argerror, validargs
 import Base: promote_rule, ==, hash, isequal, isless, typemin, typemax
 import Compat: xor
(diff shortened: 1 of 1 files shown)


ararslan:
The test failures on 0.7 are pre-existing.

src/types.jl
```diff
@@ -134,7 +135,7 @@ end
 # A `DateTime` that includes `TimeZone` information.
 # """
 
-struct ZonedDateTime <: TimeType
+struct ZonedDateTime <: Compat.AbstractDateTime
```
omus: I was originally worried about this as I thought `AbstractDateTime` was a `Union` type on 0.6 (it isn't).

omus: A quick style question/comment. I usually would just reference this without `Compat.` so that in the future when I drop that Compat requirement I don't have to change two separate lines. Do you have reason for explicitly using `Compat.` here?

ararslan: Only because it isn't exported from Compat, but I could do an explicit `using Compat: AbstractDateTime` if you'd prefer.

omus: Ah ok. I would prefer `using Compat: AbstractDateTime` if it isn't too much trouble.

ararslan: Not at all!

ararslan:
Once this is merged, could you also make a new tag?

omus:
Will do

PR #139, opened 2018-06-05
Compatibility with Julia 0.7.0-alpha

omus:


ararslan:
To fix the CI failure on 0.7 related to Mocking you might be able to just add a `Pkg.add("Mocking")` in there.

omus:
The issue is actually shown earlier and is from trying to install WinRPM on non-windows systems. This will continue to be a problem while Pkg3 doesn't support OS dependent package requirements.

omus:
I'm considering temporarily dropping LightXML (which uses WinRPM) as a Windows requirement since it's only needed to parse the Windows to tzdata translation file.

ararslan:
Would [EzXML](https://github.com/bicycle1885/EzXML.jl) be a suitable replacement? That uses BinaryProvider rather than BinDeps with WinRPM.

omus:
@ararslan it looks like EzXML is a easy solution here. Thanks for mentioning it.

omus:
😢 Looks like our XML issues aren't over. Seems like EzXML isn't yet compatible with Julia 0.7-alpha: https://github.com/bicycle1885/EzXML.jl/issues/66

omus:
Well EzXML works on 0.6 everywhere and at least can be installed on 0.7 which allows Mac and Linux to succeed. Since this is an overall improvement I'll go ahead with this.

ararslan:
The EzXML issues on 0.7 will be solved shortly: https://github.com/bicycle1885/EzXML.jl/issues/67

omus:
I'll bump up the minimum requirement in another PR. Thanks for taking care of that @ararslan! ❤️

PR #140, opened 2018-06-12
Define iterate for AbstractString in stead of just String in the iterate compatibility code.

rbvermaa:
This allows AbstractString's to be used again, as was possible before commit d061d91a892f5f643733df9c89bf36b04c5b1118.

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -11,7 +11,7 @@ end
 
 # https://github.com/JuliaLang/julia/pull/25261
 if VERSION < v"0.7.0-DEV.5126"
-    iterate(str::String, i::Int) = next(str, i)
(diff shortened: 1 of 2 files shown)


omus:
Thanks, would you mind adding a test?

rbvermaa:
Will do, but will only add it tomorrow.

omus:
Once the CI passes I'll squash and merge. @rbvermaa do you want me to tag a new release right away with this fix?

rbvermaa:
@omus No need for an immediate release, we have a workaround in place (explicitly convert to String before calling parse). It can just go into a next release :)

omus:
Included in release [v0.7.2](https://github.com/JuliaTime/TimeZones.jl/tree/v0.7.2)

PR #145, opened 2018-08-12
Localized

### LINKED ISSUE #121: Replace `ZonedDateTime` with parameterized `Localized` type
When working on #100 there were a couple things I found I would like to adjust. For one there is now:

```julia
now() -> DateTime
now(::TimeZone) -> ZonedDateTime

today() -> Date
today(::TimeZone) -> Date
```

To me the `today(::TimeZone)` method seems out of place that it returns a `Date` instead of a `ZonedDateTime` but it seems odd to return a `ZonedDateTime` set to midnight to specify an entire day. This makes me think we may want to have a "Zoned" date. I was thinking maybe we want something like: `Localized{T<:TimeType}` so we could have both `Localized{Date}` and `Localized{DateTime}` (would replace `ZonedDateTime`).

Such an update would be a pretty big change. The change would mean `today(::TimeZone) -> Localized{Date}` which then we could support `+(::Localized{Date}, ::Time) -> Localized{DateTime}`.

rofinn:
Closes #121 

- [ ] Replaced `ZonedDateTime` w/ `Localized{<:AbstractDateTime}`
- [ ] We didn't need `Localized`/`ZonedDateTime` inner constructors, so turned them to outer constructors.
- [ ] Renamed all variable references of `zdt` with `ldt`
- [ ] Renamed '2' conversion methods to use `localized` rather than `zdt`

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,17 +13,17 @@ TimeZones.jl
 [![Stable Documentation](https://img.shields.io/badge/docs-stable-blue.svg)](http://timezonesjl.readthedocs.io/en/stable/)
 [![Latest Documentation](https://img.shields.io/badge/docs-latest-blue.svg)](http://timezonesjl.readthedocs.io/en/latest/)
 
-[IANA time zone database](http://www.iana.org/time-zones) access for the [Julia](http://julialang.org/) programming language. TimeZones.jl extends the Date/DateTime support for Julia to include a new time zone aware TimeType: ZonedDateTime.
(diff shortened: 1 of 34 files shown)


omus:
Thanks for the PR. I think we should sit on this for a little while as I'd like to experiment with `Localized{Date}` and `Localized{Time}` to avoid too much churn. Having a parameterized `Localized` type will allow us to support higher-accuracy instant types which will be much more compossible with other packages.

PR #147, opened 2018-08-14
Switched `isequal` and `hash` to just use `utc_datetime`

rofinn:
NOTE: This is a breaking change that'll make most of our lives easier
(e.g., joins with `DataFrame`s) :)

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -282,24 +282,10 @@ end
 
 # Equality
 ==(a::ZonedDateTime, b::ZonedDateTime) = a.utc_datetime == b.utc_datetime
+isequal(a::ZonedDateTime, b::ZonedDateTime) = isequal(a.utc_datetime, b.utc_datetime)
+hash(zdt::ZonedDateTime, h::UInt) = hash(zdt.utc_datetime, hash(:ZonedDateTime, h))
 isless(a::ZonedDateTime, b::ZonedDateTime) = isless(a.utc_datetime, b.utc_datetime)
 
-# Note: `hash` and `isequal` assume that the "zone" of a ZonedDateTime is not being set
-# incorrectly.
-
-function hash(zdt::ZonedDateTime, h::UInt)
-    h = hash(zdt.utc_datetime, h)
-    h = hash(zdt.timezone, h)
-    return h
-end
-
-function isequal(a::ZonedDateTime, b::ZonedDateTime)
-    return (
-        isequal(a.utc_datetime, b.utc_datetime) &&
-        isequal(a.timezone, b.timezone) &&
-        isequal(a.zone, b.zone)
-    )
-end
 
 function ==(a::VariableTimeZone, b::VariableTimeZone)
     a.name == b.name && a.transitions == b.transitions
--- a/src/types.jl
+++ b/src/types.jl
@@ -282,10 +282,20 @@ end
 
 # Equality
 ==(a::ZonedDateTime, b::ZonedDateTime) = a.utc_datetime == b.utc_datetime
-isequal(a::ZonedDateTime, b::ZonedDateTime) = isequal(a.utc_datetime, b.utc_datetime)
-hash(zdt::ZonedDateTime, h::UInt) = hash(zdt.utc_datetime, hash(:ZonedDateTime, h))
 isless(a::ZonedDateTime, b::ZonedDateTime) = isless(a.utc_datetime, b.utc_datetime)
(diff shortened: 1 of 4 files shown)


src/types.jl
```diff
@@ -282,24 +282,10 @@ end
 
 # Equality
 ==(a::ZonedDateTime, b::ZonedDateTime) = a.utc_datetime == b.utc_datetime
+isequal(a::ZonedDateTime, b::ZonedDateTime) = isequal(a.utc_datetime, b.utc_datetime)
+hash(zdt::ZonedDateTime, h::UInt) = hash(zdt.utc_datetime, h)
```
iamed2: Do we want `hash(DateTime(...)) == hash(ZonedDateTime(..., tz"UTC"))`?

rofinn: Hmmm, that's a good point. Maybe we should do `hash(zdt.utc_datetime, hash(:ZonedDateTime, h))`?

rofinn: That'll make our `LaxZonedDateTime` code a little awkward as we'll want a `LaxZonedDateTime` to hash to the same value, but that's probably a reasonable compromise.

src/types.jl
```diff
@@ -282,24 +282,10 @@ end
 
 # Equality
 ==(a::ZonedDateTime, b::ZonedDateTime) = a.utc_datetime == b.utc_datetime
+isequal(a::ZonedDateTime, b::ZonedDateTime) = isequal(a.utc_datetime, b.utc_datetime)
+hash(zdt::ZonedDateTime, h::UInt) = hash(zdt.utc_datetime, hash(:ZonedDateTime, h))
```
omus: I think we should just use a hash seed which is stored as a global constant. That way other types which want to hash to the same value can reference the same seed. This would be similar to how `Range` hashing works (`hash(1:2, zero(UInt))`)

```julia
const UTC_INSTANT_HASH_SEED = UInt === UInt64 ? 0xb048e5784b3097fa : 0x2e77bd2b
function hash(zdt::ZonedDateTime, h::UInt)
    h += UTC_INSTANT_HASH_SEED
    h = hash(zdt.utc_datetime, h)  
end
```

omus:
I'll note the reason I'm ok with this change is because this makes `ZonedDateTime` hashing work based upon values much like numbers to. For example `hash(2) == hash(2.0)`. In this case the canonical representation is a UTC instant which would not take the associated time zone offset into account. This has the by product of allowing `DataFrame` joins to work with `ZonedDateTime`s using differing time zones.

src/types.jl
rofinn: I'm not sure what the advantage of that is over just using a hash of the type name? I feel like using the type name is more readable at least.

omus: The main advantage to this approach is that the UTC instant value isn't tied to any specific type so we shouldn't tie the seed to any specific name. Also, since the seed is a global it's easy to reference by other packages.

iamed2: Might as well hash a literal string "tz" or something, rather than a constant. Easier for other packages to use that.

omus: If we want to hash a `Symbol` or `String` as the seed then I would vote for `:utc_instant`

rofinn: > UTC instant value isn't tied to any specific type

Except that it kind of is. It's tied to some kind of localized or zoned DateTime.

src/types.jl
```diff
@@ -283,23 +283,15 @@ end
...
-    return (
-        isequal(a.utc_datetime, b.utc_datetime) &&
-        isequal(a.timezone, b.timezone) &&
-        isequal(a.zone, b.zone)
-    )
-end
+Compute an integer hash code for a ZonedDateTime by hashing the `utc_datetime` field.
+`hash(:utc_datetime, h)` is used to avoid collisions with `DateTime` hashes.
+"""
+hash(zdt::ZonedDateTime, h::UInt) = hash(zdt.utc_datetime, hash(:utc_datetime, h))
```
omus: Can we make this a multiline function? Also, I think `:utc_instant` is a better name

test/types.jl
```diff
@@ -311,6 +311,10 @@ y = deepcopy(x)
 @test isequal(x, y)
 @test hash(x) == hash(y)
 
+# Check that the ZonedDateTime and internal DateTime don't
+# hash to the same value.
```
omus: > plain DateTime

Issue #149, opened 2018-09-03
Pre-compilation failure on Windows


omus:
Reported by [Vadim on Slack](https://julialang.slack.com/archives/C6A044SQH/p1535980554000100):

> I'm trying to compile a sysimage on Windows and fail on TimeZones. I’ve traced it down to Sys.__init__ in Compat.jl - https://github.com/JuliaLang/Compat.jl/blob/master/src/Compat.jl#L439 and the failing line in TimeZones is https://github.com/JuliaTime/TimeZones.jl/blob/master/src/tzdata/archive.jl#L3.
>
>Basically, Compat.Sys doesn’t define BINDIR during precompile, but TimeZones wants Sys.BINDIR during precompile on Windows. I’m on Julia 0.6.4

PR #150, opened 2018-09-03
#149 - Fix precompilation failure on Windows

dm3:
Not sure about the correct Julia placement of the __init__ function. Probably OK to keep at the top of the file.

### DIFF
--- a/src/tzdata/archive.jl
+++ b/src/tzdata/archive.jl
@@ -1,7 +1,9 @@
 import Compat: @static, Sys, devnull
 
-if Sys.iswindows()
-    const exe7z = joinpath(Sys.BINDIR, "7z.exe")
+function __init__()
+    if Sys.iswindows()
+        const exe7z = joinpath(Sys.BINDIR, "7z.exe")
+    end
 end
 
 """


omus:
Thanks for the PR

src/tzdata/archive.jl
```diff
@@ -1,7 +1,9 @@
 import Compat: @static, Sys, devnull
 
-if Sys.iswindows()
-    const exe7z = joinpath(Sys.BINDIR, "7z.exe")
+function __init__()
+    if Sys.iswindows()
+        const exe7z = joinpath(Sys.BINDIR, "7z.exe")
```
omus: `const` only makes sense in the global scope currently. Also we need this variable available as a global so this should just be changed to `global`

src/tzdata/archive.jl
```diff
@@ -1,7 +1,9 @@
 import Compat: @static, Sys, devnull
 
-if Sys.iswindows()
-    const exe7z = joinpath(Sys.BINDIR, "7z.exe")
+function __init__()
```
omus: I suspect multiple `__init__` functions per module won't work

PR #152, opened 2018-09-03
Address EzXML findall deprecation

omus:
Silences a deprecation from EzXML


omus:
Thanks AppVeyor. This will need to wait until AppVeyor decides to start working again

PR #155, opened 2018-10-23
Fix 24-hour parsing with month roll over

omus:
Noticed with new 2018f tzdata release. I was aware that this could be an issue but since there were no examples in the wild until now it was hard to justify supporting this.


omus:
AppVeyor CI failure is unrelated

PR #159, opened 2018-10-30
Use Documenter 0.20 for docs

iamed2:
Uses the new format, adds linkcheck, doctests, and a separate Travis stage for building docs.


docs/src/arithmetic.md
```diff
@@ -1,9 +1,11 @@
...
+
 ## ZonedDateTime-Period Arithmetic
 
-`ZonedDateTime` uses calendrical arithmetic in a [similar manner to `DateTime`](https://docs.julialang.org/en/stable/manual/dates/#TimeType-Period-Arithmetic-1) but with some key differences. Lets look at these differences by adding a day to March 30th 2014 in Europe/Warsaw.
+`ZonedDateTime` uses calendrical arithmetic in a [similar manner to `DateTime`](https://docs.julialang.org/en/stable/stdlib/Dates/#TimeType-Period-Arithmetic-1) but with some key differences. Lets look at these differences by adding a day to March 30th 2014 in Europe/Warsaw.
 
-```julia
-julia> using Base.Dates
+```jldoctest spring
+julia> using TimeZones, Compat.Dates
```
omus: Can we just have the doctests run against Julia 1.0?

docs/src/conversions.md
```diff
@@ -1,8 +1,17 @@
+# Conversions
+
+```@meta
+DocTestSetup = quote
+	using Compat.Dates
+	using TimeZones
```
omus: Use spaces for indentation?

docs/src/arithmetic.md
omus: I think I'd prefer: `using TimeZones, Dates`

docs/src/current.md
```diff
@@ -0,0 +1,49 @@
...
+
+You should be careful not to use `today()` when working with `ZonedDateTime`s as you may end up using the wrong day. For example:
+
+```@repl tz
+midway, apia = tz"Pacific/Midway", tz"Pacific/Apia"
+ZonedDateTime(today() + Time(11), midway)
+ZonedDateTime(today() + Time(11), apia)  # Should be 2018-01-30
+ZonedDateTime(today(midway) + Time(11), midway)
+ZonedDateTime(today(apia) + Time(11), apia)
+```
```
omus: The output was important here. We need to fake the output here so the comment makes sense.

docs/src/arithmetic.md
iamed2: Sure

docs/src/current.md
iamed2: The output is generated for the built product. 

screenshot 2018-10-30 16 54 08

iamed2: A `@repl` block generates the output in repl format when the docs are built

omus: Can we dynamically update that comment?

omus: Originally was:
```
julia> ZonedDateTime(today() + Time(11), apia)  # Should be 2018-01-30
2018-01-29T11:00:00+14:00
```

iamed2: Oh I see, no. I can overwrite `today` for this block so it gives a fixed date

omus: Alternatively we can change the comment to: `# Wrong date. With the current rules apia should be one day ahead of midway`

iamed2: I like that better; done

.travis.yml
```diff
@@ -33,3 +33,12 @@ after_success:
...
       Coveralls.submit(Coveralls.process_folder())'
+jobs:
+  include:
+    - stage: "Documentation"
+      julia: 1.0
+      os: linux
+      script:
+        - julia --project=docs/ -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd())); Pkg.build("TimeZones")'
+        - julia --project=docs/ docs/make.jl
+      after_success: skip
```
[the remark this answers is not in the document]
iamed2: Yeah we could just split the script in two and have two jobs. I'll do that.

PR #161, opened 2018-11-15
Use runtime paths for TimeZones data

green-nsk:
We build an embedded julia image that is distributed across multiple machines. Fixing paths at precompilation time doesn't really work in this scenario and I couldn't find a better or less intrusive solution.

One possible option was to override constant paths in `__init__()`, but julia spits warnings at that and I couldn't really override all constants this way (namely some bits under TzData). Another alternative was to not define constants globally, and only define them in `__init__()`, but that would require even more extensive changes.

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -31,11 +31,11 @@ export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDate
     # TZData
     build
 
-const PKG_DIR = normpath(joinpath(dirname(@__FILE__), ".."))
-const DEPS_DIR = joinpath(PKG_DIR, "deps")
-const ARCHIVE_DIR = joinpath(DEPS_DIR, "tzarchive")
-const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
-const COMPILED_DIR = joinpath(DEPS_DIR, "compiled")
+PKG_DIR() = Pkg.dir("TimeZones")
+DEPS_DIR() = joinpath(PKG_DIR(), "deps")
+ARCHIVE_DIR() = joinpath(DEPS_DIR(), "tzarchive")
+TZ_SOURCE_DIR() = joinpath(DEPS_DIR(), "tzsource")
+COMPILED_DIR() = joinpath(DEPS_DIR(), "compiled")
 const TIME_ZONES = Dict{AbstractString,TimeZone}()
 
 function __init__()
@@ -66,7 +66,7 @@ function TimeZone(str::AbstractString)
             return FixedTimeZone(str)
         end
 
-        tz_path = joinpath(COMPILED_DIR, split(str, "/")...)
+        tz_path = joinpath(COMPILED_DIR(), split(str, "/")...)
         isfile(tz_path) || throw(ArgumentError("Unknown time zone \"$str\""))
 
         open(tz_path, "r") do fp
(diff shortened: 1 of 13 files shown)


omus:
Can you provide an example of the error you get with your embedded Julia image?

src/TimeZones.jl
```diff
@@ -31,11 +31,11 @@ export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDate
     # TZData
     build
 
-const PKG_DIR = normpath(joinpath(dirname(@__FILE__), ".."))
-const DEPS_DIR = joinpath(PKG_DIR, "deps")
-const ARCHIVE_DIR = joinpath(DEPS_DIR, "tzarchive")
-const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
-const COMPILED_DIR = joinpath(DEPS_DIR, "compiled")
+PKG_DIR() = Pkg.dir("TimeZones")
```
omus: It is best to avoid using `Pkg.dir` as we may be working with a copy of TimeZones which isn't installed.

omus:
An alternative to making all of the path constants functions would be to change them to be relative paths and just compute the package root at runtime. It would look something like this:

```julia
pkg_dir() = normpath(joinpath(dirname(@__FILE__), ".."))
const DEPS_DIR = "deps"
const ARCHIVE_DIR = joinpath(DEPS_DIR, "tzarchive")

function foo()
    ...
    joinpath(pkg_dir(), ARCHIVE_DIR)
    ...
end
```
If we go this route possibly having this definition may be handy:
```julia
pkgpath(args...) = normpath(joinpath(dirname(@__FILE__), "..", args...))
```

green-nsk:
I don't know if precise error message is meaningful because we're using an old version of the package. But we're hitting this exception due to PKG_PATH being build server path to julia: https://github.com/JuliaTime/TimeZones.jl/blob/master/src/TimeZones.jl#L70

I'd be glad to replace Pkg.dir() call with something else, but `joinpath(@__DIR__, "..")` doesn't seem to cut it because again value of __DIR__ is picked up at precompile time and not valid at runtime.

green-nsk:
Come to think of it, I like the idea of using `joinpath(pkg_dir(), ARCHIVE_DIR)`. At least for us we'd have to only replace one line in the whole package when building the embedded image, not 60+ lines across dozen files.

It seems many packages have this problem with binary dependencies manifesting one way or another. Most common answer is 'run Pkg.build() when you change julia paths', but this obviously doesn't work for us.

green-nsk:
I tried making the changes, and it seems like replacing `ARCHIVE_DIR()` with `joninpath(pkg_dir(), ARCHIVE_DIR)` only makes code more verbose without adding much value.

Should I just replace `Pkg.dir()` with a `__DIR__` based solution and leave the rest as is?

omus:
I'll note one of the reason I'm trying to find an alternative to `ARCHIVE_DIR()` as all-caps is a convention used for constants.

omus:
Could you elaborate on how your building the embedded image? I'd like to try to figure out how I can test for this so I don't accidentally introduce a regression. I'll note that we (Invenia) have TimeZones.jl integrated into the Julia system image so I do know the package works with pre-compiling and system images.

omus:
@green-nsk?

green-nsk:
Hey, sorry for lapsing. We've now moved to Julia 1.x and as far as I can tell, build process is now much smoother with most recent Julia and TimeZones. Thank you for your help!

PR #163, opened 2019-01-03
Default to using 2018h tzdata

omus:
Release notes: https://mm.icann.org/pipermail/tz-announce/2018-December/000053.html

Note this release used an until rule `2019 Mar Sun>=8 3:00` which is a format I should have been supporting all along. The new `tryparse_dayofmonth` is used for `parse_until` and in computing the `on` string.

I'll also note that [2018i was release 2 days after](https://mm.icann.org/pipermail/tz-announce/2018-December/000054.html) this release but I thought it was worth while making a tag with this revision.

### DIFF
--- a/src/tzdata/compile.jl
+++ b/src/tzdata/compile.jl
@@ -83,44 +83,100 @@ for (abbr, dayofweek) in DAYS
     end
 end
 
+# Generate various DateFormats based the number of periods provided.
+const UNTIL_FORMATS = let
+    parts = split("yyyy uuu dd HH:MM:SS", ' ')
+    map(i -> DateFormat(join(parts[1:i], ' ')), eachindex(parts))
+end
+
+
 isflag(flag::Char) = flag in ('w', 'u', 's')
+
+"""
+    tryparse_dayofmonth(str::AbstractString) -> Union{Function,Nothing}
+
+Parse the various day-of-month formats used within tzdata source files.
+
+```julia
+julia> tryparse_dayofmonth("lastSun")
+last_sunday (generic function with 1 method)
+
+julia> tryparse_dayofmonth("Sun>=8")
+#15 (generic function with 1 method)
+
+julia> TimeZones.TZData.tryparse_dayofmonth("15")
+#16 (generic function with 1 method)
+```
+"""
+function tryparse_dayofmonth(str::AbstractString)
+    if occursin(r"^last\w{3}$", str)
+        # We pre-built these functions above
+        # They follow the format: "lastSun", "lastMon", etc.
+        LAST_DAY_OF_WEEK[str]
+    elseif (m = match(r"^(?<dow>\w{3})(?<op>[<>]=)(?<dom>\d{1,2})$", str)) !== nothing
+        # The first day of the week that occurs before or after a given day of month.
+        # i.e. Sun>=8 refers to the Sunday after the 8th of the month
+        # or in other words, the 2nd Sunday.
+        dow = DAYS[m[:dow]]
+        dom = parse(Int, m[:dom])
+        if m[:op] == "<="
+            dt -> day(dt) <= dom && dayofweek(dt) == dow
+        else
+            dt -> day(dt) >= dom && dayofweek(dt) == dow
+        end
+    elseif occursin(r"^\d{1,2}$", str)
+        # Matches just a plain old day of the month
+        dom = parse(Int, str)
+        dt -> day(dt) == dom
+    else
+        nothing
+    end
+end
 
 # Olson time zone dates can be a single year (1900), yyyy-mm-dd (1900-Jan-01),
 # or minute-precision (1900-Jan-01 2:00).
 # They can also be given in Local Wall Time, UTC time (u), or Local Standard time (s)
-function parsedate(s::AbstractString)
-    s = replace(s, r"\s+" => " ")
-    num_periods = length(split(s, " "))
-    s, flag = num_periods > 3 && isflag(s[end]) ? (s[1:end-1], s[end]) : (s, DEFAULT_FLAG)
-    if occursin("lastSun", s)
-        dt = DateTime(replace(s, "lastSun" => "1", count=1), "yyyy uuu d H:MM:SS")
-        dt = tonext(last_sunday, dt; same=true)
-    elseif occursin("lastSat", s)
-        dt = DateTime(replace(s, "lastSat" => "1", count=1), "yyyy uuu d H:MM:SS")
-        dt = tonext(last_saturday, dt; same=true)
-    elseif occursin("Sun>=1", s)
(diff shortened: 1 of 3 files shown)


iamed2:
"parse_until" sounds like you're parsing until you reach something, rather than parsing an "until rule". Can you think of a better name?

omus:
Hmm, maybe that's why I went with `parsedate` originally. I named it that as the column it is parsing is named "until". I may change it to `parse_date` which is at least less confusing.

src/tzdata/compile.jl
```diff
@@ -69,55 +69,120 @@ const DAYS = Dict(
...
+
 isflag(flag::Char) = flag in ('w', 'u', 's')
 
+
+"""
+    tryparse_dayofmonth(str::AbstractString) -> Union{Function,Nothing}
+
+Parse the various day-of-month formats used within tzdata source files.
+
+```julia
```
ararslan: This should be `julia-repl`, which provides better highlighting for simulated REPL sessions.

src/tzdata/compile.jl
```diff
@@ -69,55 +69,120 @@ const DAYS = Dict(
...
             dayofweekofmonth(dt) == daysofweekinmonth(dt)
         end
-    )
+    end
+end
+
+# Generate various DateFormats based the number of periods provided.
+const UNTIL_FORMATS = let
+    parts = split("yyyy uuu dd HH:MM:SS", ' ')
+    date_formats = Vector{DateFormat}(undef, length(parts))
```
ararslan:
```julia
date_formats = map(i->DateFormat(join(parts[1:i], ' ')), 1:length(parts))
```
?

src/tzdata/compile.jl
```diff
@@ -69,55 +69,120 @@ const DAYS = Dict(
...
 
+
+"""
+    tryparse_dayofmonth(str::AbstractString) -> Union{Function,Nothing}
+
+Parse the various day-of-month formats used within tzdata source files.
+
+```julia
+julia> tryparse_dayofmonth("lastSun")
+lastSun (generic function with 1 method)
```
ararslan: Ew, camel case

src/tzdata/compile.jl
```diff
@@ -69,55 +69,120 @@ const DAYS = Dict(
...
+const LAST_DAY_OF_WEEK = Dict{String, Function}()
+
 # Create adjuster functions such as "lastSun".
 for (abbr, dayofweek) in DAYS
-    sym = Symbol("last" * abbr)
-    @eval (
+    str = "last" * abbr
+    sym = Symbol(str)
+    LAST_DAY_OF_WEEK[str] = @eval begin
         function $sym(dt)
```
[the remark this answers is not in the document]
omus: I could purely use anonymous functions but since these are predictable why generate the same function over again when I can give it a proper name?

src/tzdata/compile.jl
omus: TIL

src/tzdata/compile.jl
iamed2:
```julia
date_formats = map(i->DateFormat(join(parts[1:i], ' ')), 1:length(parts))
```
```julia
date_formats = map(1:length(parts)) do i
    DateFormat(join(parts[1:i], ' '))
end
```
```julia
date_formats = [DateFormat(join(parts[1:i], ' ')) for i in 1:length(parts)]
```

iamed2: Or just remove the `date_formats` variable since it's returned and never used again.

omus: I tend to be biased against `map` for some reason. I do like it better in this case if I used `eachindex`

src/tzdata/compile.jl
omus: Hmmm, good point. I can name the function whatever now that I have a mapping

src/tzdata/compile.jl
ararslan:
```julia
const UNTIL_FORMATS = let parts = split("yyyy uuu dd HH:MM:SS", ' ')
    map(i->DateFormat(join(parts[1:i], ' ')), 1:length(parts))
end
```

src/tzdata/compile.jl
```diff
@@ -69,55 +69,120 @@ const DAYS = Dict(
...
-            periods[4] = Hour(0)
-            shift += Day(1)
-        end
-        dt = DateTime(periods...) + shift
+        flag = DEFAULT_FLAG
     end
 
-    # TODO: I feel like there are issues here.
-    # If the time is UTC, we add back the offset and any saved amount
+    # Save the day of month string and substitute a non-numeric string for parsing.
```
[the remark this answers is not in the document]
omus: > Save the day of month string and replace it with "1" when it is non-numeric.

Does this read better?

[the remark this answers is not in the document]
iamed2: yep!

src/tzdata/compile.jl
ararslan: > biased against `map`

:scream: :sob:

src/tzdata/compile.jl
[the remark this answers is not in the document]
iamed2: I meant you could keep the dict but have it point to anonymous functions

src/tzdata/compile.jl
iamed2: Oddly enough, I learned about `julia-repl` while fixing up docs for this package!

ararslan: I worked with Morten to get it added to the JavaScript library that provides Documenter's highlighting, which is why I know about it. :slightly_smiling_face:

src/tzdata/compile.jl
iamed2: I find inline anonymous function syntax hard to read when there are a few pairs of parentheses going around

omus: What are your feelings on:
```julia
const UNTIL_FORMATS = let
    parts = split("yyyy uuu dd HH:MM:SS", ' ')
    map(i -> DateFormat(join(parts[1:i], ' ')), eachindex(parts))
end
```

src/tzdata/compile.jl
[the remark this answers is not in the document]
omus: @ararslan gave me a good idea as part of: https://github.com/JuliaTime/TimeZones.jl/pull/163#discussion_r245134898. I think I'll keep the names but make them more descriptive.

In the future I wouldn't mind adding them to Dates or making them more public.

src/tzdata/compile.jl
omus: Now is:
```julia
julia> tryparse_dayofmonth("lastSun")
last_sunday (generic function with 1 method)
```

PR #164, opened 2019-01-03
DateTime printing doctest fix

omus:
Address a [doctest failure](https://travis-ci.org/JuliaTime/TimeZones.jl/jobs/475022997) on nightly. Change was introduced by: https://github.com/JuliaLang/julia/pull/30200.

Additionally I escaped the use of '*' as this was messing with my Markdown syntax highlighting.

### DIFF
--- a/docs/src/faq.md
+++ b/docs/src/faq.md
@@ -6,11 +6,11 @@ DocTestSetup = quote
 end
 ```
 
-## [Why are the "Etc/*" time zones unsupported?](@id etc_tzs)
+## [Why are the "Etc/\*" time zones unsupported?](@id etc_tzs)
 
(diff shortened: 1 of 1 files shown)


iamed2:
Did you check that the HTML output of the docs shows `Etc/*` correctly?

docs/src/faq.md
```diff
@@ -32,7 +34,7 @@ julia> last(warsaw.transitions)
 2037-10-25T01:00:00 UTC+1/+0 (CET)
 
 julia> warsaw.cutoff  # DateTime up until the last transition is effective
-Nullable{DateTime}(2038-03-28T01:00:00)
+Nullable{DateTime}(DateTime(2038, 3, 28, 1))
```
ararslan: Doesn't this printing only apply to Julia master?

iamed2: See the filters at the top

iamed2: Both are accepted

omus:
> Did you check that the HTML output of the docs shows Etc/* correctly?

I checked it locally and it worked

docs/src/faq.md
ararslan: Ah, okay, thanks.

omus: The new format is only in the Julia master right now. Because of the regex I could use either string here and I think this is probably better.

PR #166, opened 2019-01-04
Minor refactoring

omus:


omus:
These tests worked locally so I'm rather surprised that to see this CI failures. Will investigate.

omus:
A single test was using `compile` directly which failed to clear the `TIME_ZONES` cache. As I have instructions in the documentation which allows users to re-compile with a different `max_year` I'm going to drop that change for now and come back to it later.

PR #167, opened 2019-01-04
Drop Julia 0.6

omus:
I'm dropping support for Julia 0.6 as I want to change `VariableTimeZone` and `FixedTimeZone` to use `name::String` instead of `name::Symbol`. I originally used symbols here as strings in Julia 0.6 are not immutable but they are in Julia 0.7 and onwards. This PR just deals with cleaning out Julia 0.6 support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ os:
   - osx
   - linux
 julia:
-  - 0.6
   - 0.7
   - 1.0
   - nightly
@@ -25,8 +24,8 @@ script:
   - git fetch origin +:refs/remotes/origin/HEAD; if ! git diff --quiet origin/HEAD HEAD -- deps/build.jl; then unset JULIA_TZ_VERSION; fi
   - |
     julia -e '
-      VERSION >= v"0.7.0-DEV.3656" && using Pkg
-      if VERSION >= v"0.7.0-DEV.5183" && (isfile("Project.toml") || isfile("JuliaProject.toml"))
+      using Pkg
+      if isfile("Project.toml") || isfile("JuliaProject.toml")
           Pkg.build(); Pkg.test(coverage=true)
       else
           Pkg.clone(pwd()); Pkg.build("TimeZones"); Pkg.test("TimeZones"; coverage=true)
@@ -34,8 +33,7 @@ script:
 after_success:
   - |
     julia -e '
-      VERSION >= v"0.7.0-DEV.3656" && using Pkg
-      VERSION >= v"0.7.0-DEV.5183" || cd(Pkg.dir("TimeZones"))
+      using Pkg
       Pkg.add("Coverage"); using Coverage
       Codecov.submit(process_folder())
       Coveralls.submit(Coveralls.process_folder())'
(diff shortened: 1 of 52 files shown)


src/deprecated.jl
```diff
@@ -1,22 +1,9 @@
...
-    # Only remove this method when support for Julia 0.6 is dropped.
-    colon(start::T, stop::T) where {T <: ZonedDateTime} = start:Day(1):stop
-elseif VERSION < v"0.7.0-DEV.4003"  # JuliaLang/julia#26074
-    @deprecate colon(start::T, stop::T) where {T <: ZonedDateTime} start:Day(1):stop  false
-else
-    # Only remove this deprecation when support for Julia 0.7 is dropped.
-    @deprecate (:)(start::T, stop::T) where {T <: ZonedDateTime}   start:Day(1):stop  false
-end
+# Only remove this deprecation when support for Julia 0.7 is dropped (JuliaLang/julia#24258)
+@deprecate ((:)(start::T, stop::T) where {T <: ZonedDateTime}) start:Day(1):stop false
```
omus: I should make this only apply on 0.7

omus: Done

.travis.yml
```diff
@@ -25,17 +24,16 @@ script:
   - git fetch origin +:refs/remotes/origin/HEAD; if ! git diff --quiet origin/HEAD HEAD -- deps/build.jl; then unset JULIA_TZ_VERSION; fi
   - |
     julia -e '
-      VERSION >= v"0.7.0-DEV.3656" && using Pkg
```
ararslan: I think you can ditch all of the custom logic here now and just use the Travis default steps.

REQUIRE
```diff
@@ -1,5 +1,3 @@
-julia 0.6
-Compat 0.66
+julia 0.7.0
 Mocking 0.5.6
-Nullables 0.0.7
```
ararslan: :100:

src/arithmetic.jl
```diff
@@ -1,27 +1,19 @@
-import Base: +, -
-import Compat: @static
-
-if VERSION < v"0.7.0-DEV.4955"
-    import Base: broadcast
-    const broadcasted = broadcast
-else
-    import Base.Broadcast: broadcasted
-end
+import Base.Broadcast: broadcasted
```
ararslan: If this is being extended can you just use the module annotation there rather than `import`ing it?

omus: In this particular case I thought the `import` made things a bit clearer. I'm not a fan of `Base.Broadcast.broadcasted(...) = `

.travis.yml
omus: I may do that as another PR as https://docs.travis-ci.com/user/languages/julia/#coverage has sent me down a rabbit hole of changes.

src/arithmetic.jl
ararslan: That's fair. Though you could also do
```julia
using Base.Broadcast

Broadcast.broadcasted(...) =
```

omus: That's a good point. I do like that but I don't want o wait for the CI again 😆

omus:
Appveyor fail due to download issue.

.travis.yml
omus: Addressed here: https://github.com/JuliaTime/TimeZones.jl/pull/170

PR #168, opened 2019-01-04
Use immutable String instead of Symbol

omus:
Required: https://github.com/JuliaTime/TimeZones.jl/pull/167

The main advantage here is that we were always pretending that the `TimeZone` name was a string. For the "discovery.jl" operations this could end up creating a bunch of unnecessary allocations.


omus:
> Looks good to me. Any idea whether this could break downstream packages that use TimeZones?

Only if they were accessing the field directly and required it to be a `Symbol`

omus:
Since I didn't have an inner constructor I guess there is a small chance that someone was constructing a custom `TimeZone` with a `Symbol` directly but I think that is extremely unlikely.

PR #170, opened 2019-01-04
Refactor Travis CI config

omus:
Useful resources:
- Documentation on Julia's Travis support: https://docs.travis-ci.com/user/languages/julia/#coverage
- Default Travis CI script: https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/julia.rb

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,25 @@ stages:
   - name: Documentation (Deploy)
     if: (branch = master AND type != pull_request) OR tag =~ ^v
 
+.doc_test: &doc_test
+  stage: "Documentation (Test)"
+  os: linux
+  julia: 1.0
+  script:
+    - julia --project=docs/ -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd())); Pkg.build("TimeZones")'
+    - julia --project=docs/ docs/make.jl
+  after_success: skip
+
+.doc_deploy: &doc_deploy
+  stage: "Documentation (Deploy)"
+  os: linux
+  julia: 1.0
+  script:
+    - julia --project=docs/ -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd())); Pkg.build("TimeZones")'
+    - julia --project=docs/ docs/deploy.jl
+  after_success: skip
+
+# Note: At the moment `codecov: true` and `coveralls: true` are not working.
 language: julia
 os:
   - osx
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,6 +4,25 @@ stages:
...
+.doc_deploy: &doc_deploy
+  stage: "Documentation (Deploy)"
+  os: linux
+  julia: 1.0
+  script:
+    - julia --project=docs/ -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd())); Pkg.build("TimeZones")'
+    - julia --project=docs/ docs/deploy.jl
+  after_success: skip
+
+# Note: At the moment `codecov: true` and `coveralls: true` are not working.
```
omus: Were causing CI failure for me so I went back to using a custom `after_success`

.travis.yml
```diff
@@ -13,15 +32,14 @@ julia:
...
-  - JULIA_PROJECT="@."
+  - JULIA_TZ_VERSION="2016j"  # Ideally only set for the "test" stage
 notifications:
   email: false
-env:
-  - JULIA_TZ_VERSION="2016j"
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   # Avoid setting JULIA_TZ_VERSION if the deps/build.jl file has been modified
   - git fetch origin +:refs/remotes/origin/HEAD; if ! git diff --quiet origin/HEAD HEAD -- deps/build.jl; then unset JULIA_TZ_VERSION; fi
```
omus: I could remove the entire `script` section if I would move this part into a `before_script` section. I'll note that the unsetting of the environmental variable does seem to persist between these stages but I'm reluctant to trust that this will continue to be the case.

.travis.yml
```diff
@@ -13,15 +32,14 @@ julia:
...
+  - echo $JULIA_TZ_VERSION  # Display version here as Julia 0.7+ doesn't show the build logs by default
```
omus: Added as it's hard to verify that this variable is correctly being unset now that the `Pkg.build` output is suppressed. @ararslan mentioned with Julia 1.1 you can use `Pkg.build(verbose=true)`

ararslan: That actually happens by default using the default Travis script with 1.1 or later. (Or rather it will once Fredrik's Travis PR is merged and deployed.)

.travis.yml
```diff
@@ -4,6 +4,25 @@ stages:
   - name: Documentation (Deploy)
     if: (branch = master AND type != pull_request) OR tag =~ ^v
 
+.doc_test: &doc_test
```
ararslan: This isn't just a GitLab thing? TIL.

PR #172, opened 2019-01-07
Introduce `TZSource` for managing tzdata source files

omus:
As I work towards parsing both standard time zones and legacy time zones I need a nice way of containing the intermediate representation of the tz source files. The `TZSource` structure allows me to do this by having a Julia representation of the tz source files in a single struct. Additionally, part of this work required me to solve an issue where tz links can reference time zone names in other tz source files. This intermediate representation is also helpful in that scenario.


src/types.jl
```diff
@@ -89,6 +89,9 @@ function FixedTimeZone(s::AbstractString)
     return FixedTimeZone(name, offset)
 end
 
+name(tz::FixedTimeZone) = tz.name
```
iamed2: `Base.nameof` maybe?

src/tzdata/compile.jl
```diff
@@ -508,90 +520,115 @@ function resolve!(zone_name::AbstractString, zoneset::ZoneDict, ruleset::RuleDic
...
 end
 
-function resolve(zoneset::ZoneDict, ruleset::RuleDict; max_year::Integer=MAX_YEAR, debug=false)
-    ordered = OrderedRuleDict()
-    time_zones = Dict{AbstractString,TimeZone}()
 
-    for zone_name in keys(zoneset)
-        tz = resolve!(zone_name, zoneset, ruleset, ordered; max_year=max_year, debug=debug)
-        time_zones[zone_name] = tz
+function TZSource(files::AbstractString...)
```
iamed2: Collection of files instead?

src/tzdata/compile.jl
```diff
@@ -508,90 +520,115 @@ function resolve!(zone_name::AbstractString, zoneset::ZoneDict, ruleset::RuleDic
...
-            elseif kind == "Zone"
-                zone = parse(Zone, line)
-                zones[name] = push!(get(zones, name, Zone[]), zone)
-            elseif kind == "Link"
-                dest = line
-                links[dest] = name
-            else
-                @warn "Unhandled line found with type: $kind"
-            end
+        line = strip(replace(chomp(line), r"#.*$" => ""))
```
iamed2: `chomp` is unnecessary since you're calling `strip` after

src/utils.jl
```diff
@@ -89,3 +89,48 @@ function optional(ex::Expr)
...
+    end
+
+    push!(tail, :(throw(ArgumentError("Unable to replace non-existent field $k within the struct $($T)"))))
+
+    push!(expr.args, Expr(:call, T, vars...))
+    return expr
+end
+
+"""
+    _replace(object, p::Pair{Symbol}...)q
```
iamed2: q

src/utils.jl
```diff
@@ -89,3 +89,48 @@ function optional(ex::Expr)
 
     return funcs
 end
+
+
+function _gen_replace(T::Type, struct_var::Symbol, pair_var::Symbol)
```
iamed2: My hunch is that this is not faster than iterating over the fields in a normal function, which would be much more readable.

src/types.jl
omus: Looks like `nameof` always returns a `Symbol`. I'll note I'm not exporting this accessor.

iamed2: Ah okay then

omus: Would you prefer if we left this accessor out completely for now?

src/tzdata/compile.jl
omus: Ok, I'll end up with this then:
```julia
function TZSource()
    TZSource(
        Dict{String,Vector{Zone}}(),
        Dict{String,Vector{Rule}}(),
        Dict{String,String}(),
    )
end

TZSource(file::AbstractString) = load!(TZSource(), file)

function TZSource(files::AbstractVector{<:AbstractString})
    tz_source = TZSource()

    for file in files
        load!(tz_source, file)
    end

    return tz_source
end
```

src/types.jl
ararslan: As long as it's not exported, it seems okay to me.

src/tzdata/compile.jl
```diff
@@ -508,90 +539,101 @@ function resolve!(zone_name::AbstractString, zoneset::ZoneDict, ruleset::RuleDic
...
 
-function resolve(zone_name::AbstractString, zoneset::ZoneDict, ruleset::RuleDict; max_year::Integer=MAX_YEAR, debug=false)
-    ordered = OrderedRuleDict()
-    return resolve!(zone_name, zoneset, ruleset, ordered; max_year=max_year, debug=debug)
-end
+    # For the intial pass we'll collect the zone and rule lines.
+    for line in eachline(io)
+        # Lines that start with whitespace can be considered a "continuation line" which
+        # means the last found kind/name should persist.
+        persist = occursin(r"^\s", line)
```
ararslan: It looks like this is only used once, so perhaps the comment could be moved atop the conditional where this is used and the condition could instead be `!occursin(r"^\s", line)`.

src/tzdata/compile.jl
```diff
@@ -508,90 +539,101 @@ function resolve!(zone_name::AbstractString, zoneset::ZoneDict, ruleset::RuleDic
...
+function load!(tz_source::TZSource, file::AbstractString)
+    open(file, "r") do io
+        load!(tz_source, io)
     end
+end
 
-    return zones, rules
+function compile(name::AbstractString, tz_source::TZSource; kwargs...)
+    ordered = OrderedRuleDict()
+    if name in keys(tz_source.links)
```
ararslan:
```julia
if haskey(tz_source.links, name)
```

src/tzdata/compile.jl
omus: The `strip` line after this will unfortunately eliminate the space we need to check this.

ararslan: Oh, good call!

src/tzdata/compile.jl
omus: Good call!

src/tzdata/compile.jl
```diff
@@ -508,90 +539,101 @@ function resolve!(zone_name::AbstractString, zoneset::ZoneDict, ruleset::RuleDic
...
 
     isdir(dest_dir) || error("Destination directory doesn't exist")
     empty!(TIME_ZONES)
 
-    for (name, tz) in time_zones
-        parts = split(name, "/")
-        tz_dir, tz_file = joinpath(dest_dir, parts[1:end-1]...), parts[end]
+    for tz in time_zones
+        parts = split(TimeZones.name(tz), '/')
+        tz_path = joinpath(dest_dir, parts...)
```
ararslan: Are you splitting on slashes and splatting into `joinpath` so that Windows paths are normalized from `/` separation to `\`? If so, I believe you can just make this
```julia
tz_path = normpath(dest_dir, TimeZones.name(tz))
```

omus: I am converting `/` into the file separator. I do like the idea of switching this to `normpath` but I'm hesitant as the documentation for the function does not mention this behaviour.

ararslan: Oh you're right, it doesn't. That's odd. Do whatever feels right.

PR #173, opened 2019-01-08
 Separate types into multiple files

omus:
Reduces the main "TimeZones.jl" to just the basics.


omus:
Note when I separated the tests I also turned them into testsets.

Issue #154, opened 2018-09-20
Allow backwards compatible time zones for localzone


omus:
I've noticed that the default time zone used by macOS High Sierra uses `US/Pacific` which is a time zone which IANA keeps around for backwards compatibility. We may want to allow the use of these backwards compatible names when using `localzone()`.

omus:
Another example is `America/Montreal`

PR #174, opened 2019-01-10
Support legacy time zones

omus:
In order to address https://github.com/JuliaTime/TimeZones.jl/issues/154 we needed to compile the time zones which are included in the "backward"  tz source file. Previously these zones were not included as they are deemed as legacy time zones and should be phased out. As such I've introduced a `Class` type to restrict which time zones are allowed to be created by the `TimeZone` constructor. Doing this allows `localzone` to support both standard and legacy time zones while not allowing legacy time zones by default.

One other big change to note is that `VariableTimeZone`s are considered `==` if the transitions are the same. Previously the name was also required to be the same.

Fixes: https://github.com/JuliaTime/TimeZones.jl/issues/154

### DIFF
--- a/src/build.jl
+++ b/src/build.jl
@@ -6,7 +6,7 @@ Builds the TimeZones package with the specified tzdata `version` and `regions`. 
 The `force` flag is used to re-download tzdata archives.
 """
 function build(version::AbstractString="latest"; force::Bool=false)
-    version, tz_category = TimeZones.TZData.build(version)
+    TimeZones.TZData.build(version)
 
     if Sys.iswindows()
         TimeZones.WindowsTimeZoneIDs.build(force=force)
@@ -14,10 +14,6 @@ function build(version::AbstractString="latest"; force::Bool=false)
 
     # Reset cached information
     empty!(TIME_ZONE_CACHE)
-    empty!(TIME_ZONE_NAMES)
-    for (class, tz_names) in tz_category
-        TIME_ZONE_NAMES[class] = sort!(tz_names)
-    end
 
     @info "Successfully built TimeZones"
 end
--- a/src/build.jl
+++ b/src/build.jl
@@ -1,16 +1,12 @@
-using .TZData: REGIONS
-
 """
-    build(version="latest", regions=REGIONS; force=false) -> Nothing
+    build(version="latest"; force=false) -> Nothing
 
 Builds the TimeZones package with the specified tzdata `version` and `regions`. The
 `version` is typically a 4-digit year followed by a lowercase ASCII letter (e.g. "2016j").
-Users can also specify which `regions`, or tz source files, should be compiled. Available
-regions are listed under `TimeZones.REGIONS` and `TimeZones.LEGACY_REGIONS`. The `force`
-flag is used to re-download tzdata archives.
+The `force` flag is used to re-download tzdata archives.
 """
-function build(version::AbstractString="latest", regions=REGIONS; force::Bool=false)
-    TimeZones.TZData.build(version, regions)
+function build(version::AbstractString="latest"; force::Bool=false)
+    TimeZones.TZData.build(version)
 
     if Sys.iswindows()
         TimeZones.WindowsTimeZoneIDs.build(force=force)
--- a/src/build.jl
+++ b/src/build.jl
@@ -6,10 +6,17 @@ Builds the TimeZones package with the specified tzdata `version` and `regions`. 
 The `force` flag is used to re-download tzdata archives.
 """
 function build(version::AbstractString="latest"; force::Bool=false)
-    TimeZones.TZData.build(version)
+    version, tz_category = TimeZones.TZData.build(version)
 
     if Sys.iswindows()
         TimeZones.WindowsTimeZoneIDs.build(force=force)
+    end
+
+    # Reset cached information
+    empty!(TIME_ZONE_CACHE)
+    empty!(TIME_ZONE_NAMES)
+    for (class, tz_names) in tz_category
+        TIME_ZONE_NAMES[class] = sort!(tz_names)
     end
 
     @info "Successfully built TimeZones"
--- a/src/class.jl
+++ b/src/class.jl
@@ -36,7 +36,7 @@ function Base.getproperty(::Type{Class}, field::Symbol)
     end
 end
 
-function classify(str::AbstractString, regions::AbstractSet{<:AbstractString})
+function Class(str::AbstractString, regions::AbstractSet{<:AbstractString})
     class = Class.NONE
     occursin(FIXED_TIME_ZONE_REGEX, str) && (class |= Class.FIXED)
     !isempty(intersect(regions, TZData.STANDARD_REGIONS)) && (class |= Class.STANDARD)
@@ -44,7 +44,7 @@ function classify(str::AbstractString, regions::AbstractSet{<:AbstractString})
     return class
 end
 
-classify(str::AbstractString, regions::AbstractVector) = classify(str, Set{String}(regions))
+Class(str::AbstractString, regions::AbstractVector) = Class(str, Set{String}(regions))
 
 Base.:(|)(a::Class, b::Class) = Class(a.val | b.val)
 Base.:(&)(a::Class, b::Class) = Class(a.val & b.val)
--- a/src/class.jl
+++ b/src/class.jl
@@ -19,17 +19,17 @@ struct Class
 end
 
 function Base.getproperty(::Type{Class}, field::Symbol)
-    if field == :NONE
+    if field === :NONE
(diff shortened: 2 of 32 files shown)


omus:
Something that came up in an offline discussion:

One of the reasons we use a bit mask here is that a time zone could be associated with multiple classes. For example the "UTC" time zone is classified as `FIXED | STANDARD | LEGACY` (I'll note the `STANDARD` classification is due to the special "utc" tz source file we have).

omus:
Weird the CI doesn't look like it ran

src/types/timezone.jl
```diff
@@ -1,28 +1,38 @@
...
+const TIME_ZONE_CACHE = Dict{String,Tuple{TimeZone,Class}}()
 
 """
-    TimeZone(str::AbstractString) -> TimeZone
+    TimeZone(str::AbstractString, mask::Class=Class.DEFAULT) -> TimeZone
 
-Constructs a `TimeZone` subtype based upon the string. If the string is a recognized time
-zone name then data is loaded from the compiled IANA time zone database. Otherwise the
-string is assumed to be a static time zone.
+Constructs a `TimeZone` subtype based upon the string and the provided `mask`. If the
```
iamed2: I think this over-emphasizes `mask`. Most people won't ever need to touch it. Maybe list `TimeZone(str)` first at the top and note that it defaults to `DEFAULT`, then list `TimeZone(str, mask)`?

src/types/variabletimezone.jl
```diff
@@ -26,8 +26,14 @@ function rename(tz::VariableTimeZone, name::AbstractString)
     VariableTimeZone(name, tz.transitions, tz.cutoff)
 end
 
-function Base.:(==)(a::VariableTimeZone, b::VariableTimeZone)
-    a.name == b.name && a.transitions == b.transitions
+Base.:(==)(a::VariableTimeZone, b::VariableTimeZone) = a.transitions == b.transitions
+
+function Base.isequal(a::VariableTimeZone, b::VariableTimeZone)
+    return (
+        isequal(a.name, b.name) &&
```
iamed2: This seems to be reversed from the ZonedDateTime definitions?

test/interpret.jl
```diff
@@ -1,5 +1,5 @@
 # Contains both positive and negative UTC offsets and observes daylight saving time.
-apia = compile("Pacific/Apia", tzdata["australasia"])
+apia = first(compile("Pacific/Apia", tzdata["australasia"]))
```
iamed2: You had to add `first` a lot. Is there a way to avoid that?

src/types/timezone.jl
omus: I can make these separate docstrings (still one function definition) to hide most of the mask details

test/interpret.jl
omus: I could make:
```julia
function compile_tz(args...)
    tz, class = compile(args...)
    return tz
end
```
I didn't think it was worth it though

iamed2: Yeah probably not I guess

omus:
I failed to notice some additional changes required in other commits from f76e093. Should be addressed now.

src/build.jl
```diff
@@ -1,20 +1,19 @@
...
 `version` is typically a 4-digit year followed by a lowercase ASCII letter (e.g. "2016j").
-Users can also specify which `regions`, or tz source files, should be compiled. Available
-regions are listed under `TimeZones.REGIONS` and `TimeZones.LEGACY_REGIONS`. The `force`
-flag is used to re-download tzdata archives.
+The `force` flag is used to re-download tzdata archives.
 """
-function build(version::AbstractString="latest", regions=REGIONS; force::Bool=false)
-    TimeZones.TZData.build(version, regions)
+function build(version::AbstractString="latest"; force::Bool=false)
+    version = TimeZones.TZData.build(version)
```
ararslan: Am I blind or is `version` not being used?

src/class.jl
```diff
@@ -0,0 +1,63 @@
...
+- `Class.NONE`: Mask that will match nothing.
+- `Class.DEFAULT`: Default mask used by functions: `Class.FIXED | Class.STANDARD`
+- `Class.ALL`: Mask allowing all supported classes.
+"""
+struct Class
+    val::UInt8
+end
+
+function Base.getproperty(::Type{Class}, field::Symbol)
+    if field == :NONE
```
ararslan: Might as well use `===` when comparing against symbols. The compiler likes `===`. All hail the compiler.

src/class.jl
```diff
@@ -0,0 +1,63 @@
...
+- `Class.LEGACY`: The time zone name is included in the deprecated IANA tz source files.
+- `Class.NONE`: Mask that will match nothing.
+- `Class.DEFAULT`: Default mask used by functions: `Class.FIXED | Class.STANDARD`
+- `Class.ALL`: Mask allowing all supported classes.
+"""
+struct Class
+    val::UInt8
+end
+
+function Base.getproperty(::Type{Class}, field::Symbol)
```
ararslan: Seems like a use case for `Enum`s?

src/tzdata/build.jl
```diff
@@ -68,7 +79,11 @@ function build(version::AbstractString="latest", regions=REGIONS)
         rm(joinpath(COMPILED_DIR, file), recursive=true)
     end
 
-    version = build(version, regions, ARCHIVE_DIR, TZ_SOURCE_DIR, COMPILED_DIR, verbose=true)
+    version = build(
```
ararslan: Seemed more readable the other way, IMO.

ararslan:
The use of bitwise operations on unsigned flag values to make a mask feels very C-ish to me. I wonder if there wouldn't be a more satisfying API we could use for this...

src/class.jl
iamed2: Doesn't work with flags and `|`

ararslan: It does if you define a `|` method

omus: Even if you define `|` you can't get any arbitrary combination of values:
```julia
julia> @enum Class NONE=0x00 FIXED=0x01 STANDARD=0x02 LEGACY=0x04

julia> Base.:(|)(a::Class, b::Class) = Class(Int(a) | Int(b))

julia> STANDARD | LEGACY
ERROR: ArgumentError: invalid value for Enum Class: 6
```

src/tzdata/build.jl
omus: Was modified to fit within the 92 column line length

src/class.jl
omus: Good point

src/build.jl
omus: It was used in previous incarnations. I'll correct this.

omus:
I'll note that using the `Class` type won't be common place and is mostly used internally. I'm open to other APIs here but so far the best approach seems to be a bit mask encapsulated in a type.

ararslan:
Okay, if you and Eric are satisfied with that, that's fine.

src/class.jl
rofinn: FWIW, this `Class` approach seems fine to me. It's pretty close to what I did with file modes in FilePaths.jl:

https://github.com/rofinn/FilePathsBase.jl/blob/master/src/mode.jl

src/class.jl
```diff
@@ -0,0 +1,72 @@
...
```
rofinn: This part seems a little awkward to me. I'd probably just use constants that aren't exported.

```
const NONE = Class(0x00)
...
const DEFAULT = FIXED | STANDARD | LEGACY
...
```

rofinn: Alternatively, maybe you could do `Class[:STANDARD]`, `Class("STANDARD")` or `class"STANDARD"` just to make it clear that `Class` isn't a module when you read `Class.STANDARD`?

ararslan: Alternatively, we can just make it a module:
```julia
baremodule Class
const NONE = 0x00
const FIXED = 0x01
...
end
```
Then you get `Class.FIXED` syntax for free and don't need to define any special bitwise operations for `Class` objects since the actual values you're using already have those defined.

ararslan: True. You'd be able to recover some of that with
```julia
baremodule Class
@enum TZClass begin
    const NONE = 0x00
    ...
end
end
```
but that's getting into unnecessary complication territory.

omus: I'm fine with this use of `getproperty` but since there is some controversy I'm going to switch `Class.STANDARD` to `Class(:STANDARD)`. 

I'm not a fan of `Class[:STANDARD]` as that makes me feel like this is an instance I can pull values out of like an array or a dataframe. `Class("STANDARD")` is ok but I already have a constructor which uses the time zone name as a string.

PR #176, opened 2019-01-18
Add Project.toml

omus:
We need to keep the REQUIRE file for now.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,22 @@
+name = "TimeZones"
+uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
+authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
+
+[deps]
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+EzXML = "8f5d6c58-4d21-5cfd-889c-e3ad7ee6a615"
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -0,0 +1,23 @@
+name = "TimeZones"
+uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
+authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
+version = "0.8.4"
```
ararslan: Do we need a version here or can it just be left out?

Project.toml
```diff
@@ -0,0 +1,23 @@
...
+[deps]
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+EzXML = "8f5d6c58-4d21-5cfd-889c-e3ad7ee6a615"
+Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
+Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
+
+[compat]
+julia = "0.7, 1.0"
```
ararslan: Note that running locally with `julia --project=.` will rearrange this section and put `julia` at the bottom.

Issue #175, opened 2019-01-17
Timezones should broadcast as scalars


oxinabox:
In julia 1.0 broadcasting defaults to iterating.

So if you do say

```
ZonedDateTime.([now(), now(), now()], tz"UTC")
```
you get the error

`MethodError: no method matching length(::FixedTimeZone)`

The workaround is
`ZonedDateTime.([now(), now(), now()], Ref(tz"UTC"))`

Which works fine.

But since we take broadcasting seriously we should be able to set the broadcaststyle of an `AbstractTimeZone` to be scalar-like.
Then the original example will work

oxinabox:
Is there a backport of it anywhere?
Maybe in Compat?

PR #177, opened 2019-01-18
Define broadcastable for scalars

omus:
Note: `AbstractDateTime` already had this defined.

Fixes: https://github.com/JuliaTime/TimeZones.jl/issues/175

### DIFF
--- a/src/types/timezone.jl
+++ b/src/types/timezone.jl
@@ -105,3 +105,11 @@ function istimezone(str::AbstractString, mask::Class=Class(:DEFAULT))
 
     return tz !== nothing && mask & class != Class(:NONE)
 end
+
+# After `broadcastable` was defined but before this was addressed in the `Dates` stdlib
+# - Broadcastable introduction: https://github.com/JuliaLang/julia/pull/26601
+# - Fixed in Dates: https://github.com/JuliaLang/julia/pull/30159
+# Note: The change was backported to 1.1 as well.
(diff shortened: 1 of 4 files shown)


src/types/timezone.jl
```diff
@@ -105,3 +105,10 @@ function istimezone(str::AbstractString, mask::Class=Class(:DEFAULT))
 
     return tz !== nothing && mask & class != Class(:NONE)
 end
+
+# After `broadcastable` was defined but before this was addressed in the `Dates` stdlib
+# - Broadcastable introduction: https://github.com/JuliaLang/julia/pull/26601
+# - Fixed in Dates: https://github.com/JuliaLang/julia/pull/30159
+if v"0.7.0-DEV.4743" <= VERSION <= v"1.2.0-DEV.114"
+    Base.broadcastable(tz::TimeZone) = Ref(tz)
```
iamed2: This will overwrite the base definition on v1.1

omus: I'll update this rule to skip the back-ported revisions

omus: Addressed to avoid the back ported changes in revision 6778bef3ad

omus:
> The thing about these tests is that they'd still pass if broadcasting was enabled if you defined iteration for time zones.

Would an explicit `broadcastable` test in addition to the current test be better?

iamed2:
I think you should just broadcast with something else and test the result

PR #190, opened 2019-03-28
Default to using 2019a tzdata

omus:
Release notes: https://mm.icann.org/pipermail/tz-announce/2019-March/000055.html

At the present time there is no windowsZones.xml file update.


omus:
AppVeyor and Travis CI on macOS is happy. Linux is super backlogged so I'll get this merged without Travis on Linux.

omus:
Note: the CI failed as I deleted the branch before the CI finished.

PR #191, opened 2019-03-28
Officially support Julia 1.1

omus:


Project.toml
```diff
@@ -13,7 +13,7 @@ Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 [compat]
 EzXML = ">= 0.8"
 Mocking = ">= 0.5.6"
-julia = "0.7, 1.0"
+julia = "0.7, 1.0, 1.1"
```
oxinabox: `1.0` already covers this, under Carat (default) version spec.

oxinabox: It is a minor point but listing it here does propergate missunderstanding caret specifiers.

oxinabox: I think "Correct" would be either

```suggestion
julia = "~0.7, ~1.0, ~1.1"
```

or 
```suggestion
julia = "0.7, 1.0"
```
 == 
```suggestion
julia = "0.7, 1,"
```

omus: My bad. I think I like the explicitness of the first suggestion

iamed2: My preferred way of spelling this is `"0.7, 1"`

omus: Ok, @iamed2 had a chat and I am going to go with `julia = "0.7, 1"` which implies that I'm trusting semver is being properly followed. If I would go with the explicit version check I'm basically saying I don't trust semver and I only will trust packages I explicitly test against.

PR #192, opened 2019-03-28
Address corner cases in next_transition_instant

omus:
```julia
julia> next_transition_instant(tz"Pacific/Honolulu")
ERROR: BoundsError: attempt to access 8-element Array{TimeZones.Transition,1} at index [9]
Stacktrace:
 [1] getindex(::Array{TimeZones.Transition,1}, ::Int64) at ./array.jl:731
 [2] next_transition_instant(::ZonedDateTime) at /Users/omus/.julia/packages/TimeZones/WMDpl/src/discovery.jl:172
 [3] next_transition_instant(::VariableTimeZone) at /Users/omus/.julia/packages/TimeZones/WMDpl/src/discovery.jl:177
 [4] top-level scope at none:0

julia> next_transition_instant(astimezone(ZonedDateTime(get(wpg.cutoff), tz"UTC"), wpg))
ERROR: UnhandledTimeError: TimeZone America/Winnipeg does not handle dates on or after 2038-03-14T08:00:00 UTC
Stacktrace:
 [1] Type at /Users/omus/.julia/packages/TimeZones/WMDpl/src/types.jl:150 [inlined]
 [2] astimezone(::ZonedDateTime, ::VariableTimeZone) at /Users/omus/.julia/packages/TimeZones/WMDpl/src/conversions.jl:86
 [3] top-level scope at none:0
```

### DIFF
--- a/src/discovery.jl
+++ b/src/discovery.jl
@@ -126,11 +126,12 @@ end
 
 
 """
-    next_transition_instant(zdt::ZonedDateTime) -> ZonedDateTime
-    next_transition_instant(tz::TimeZone=localzone()) -> ZonedDateTime
+    next_transition_instant(zdt::ZonedDateTime) -> Union{ZonedDateTime, Nothing}
+    next_transition_instant(tz::TimeZone=localzone()) -> Union{ZonedDateTime, Nothing}
 
 Determine the next instant at which a time zone transition occurs (typically
-due to daylight-savings time).
+due to daylight-savings time). If no there exists no future transition then `nothing` will
+be returned.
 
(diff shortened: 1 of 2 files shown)


iamed2:
Seems like if anyone were to want to use this API they wouldn't want the warning

omus:
> Seems like if anyone were to want to use this API they wouldn't want the warning

The warning is only for `show_next_transition` which I expect will only be used interactively. `next_transition_instant` will not have a warning. Does that seem sensible?

iamed2:
Yep!

Issue #185, opened 2019-02-18
`ArgumentError` when `ZonedDateTime` called on a String


nickrobinson251:
Passing just a String to `ZonedDateTime` gives `ArgumentError: Unknown time zone ""` -- I would have expected `MethodError` if passing in a String were not allowed.

```julia
julia> using TimeZones

julia> ZonedDateTime("2018-11-01")
ERROR: ArgumentError: Unknown time zone ""
Stacktrace:
 [1] (::getfield(TimeZones, Symbol("##3#4")){String})() at /Users/nick/.julia/packages/TimeZones/59qDu/src/types/timezone.jl:54
 [2] get!(::getfield(TimeZones, Symbol("##3#4")){String}, ::Dict{String,Tuple{TimeZone,TimeZones.Class}}, ::String) at ./dict.jl:453
 [3] TimeZone(::String, ::TimeZones.Class) at /Users/nick/.julia/packages/TimeZones/59qDu/src/types/timezone.jl:46 (repeats 2 times)
 [4] ZonedDateTime(::Int64, ::Int64, ::Int64, ::Int64, ::Int64, ::Int64, ::Int64, ::String) at /Users/nick/.julia/packages/TimeZones/59qDu/src/types/zoneddatetime.jl:122
 [5] parse at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/Dates/src/parse.jl:285 [inlined]
 [6] ZonedDateTime(::String, ::Dates.DateFormat{Symbol("yyyy-mm-ddTHH:MM:SS.ssszzz"),Tuple{Dates.DatePart{'y'},Dates.Delim{Char,1},Dates.DatePart{'m'},Dates.Delim{Char,1},Dates.DatePart{'d'},Dates.Delim{Char,1},Dates.DatePart{'H'},Dates.Delim{Char,1},Dates.DatePart{'M'},Dates.Delim{Char,1},Dates.DatePart{'S'},Dates.Delim{Char,1},Dates.DatePart{'s'},Dates.DatePart{'z'}}}) at /Users/nick/.julia/packages/TimeZones/59qDu/src/parse.jl:84 (repeats 2 times)
 [7] top-level scope at none:0
```

This is not a helpful error message, since it suggests (to me at least) 
that I need `ZonedDateTime("2018-11-01", tz"UTC")`, but that doesn't work and what I really need is something like `ZonedDateTime(2018, 11, 01, tz"UTC")`

[...]

PR #193, opened 2019-03-28
Improve parse constructor exception message

omus:
Fixes: https://github.com/JuliaTime/TimeZones.jl/issues/185


omus:
I was considering deprecating these constructors but I think the inconsistency with Dates isn't worth the trouble.

PR #194, opened 2019-03-29
Overhaul printing of types

omus:
Should allow `repr` to display the program representation of the instances. Related to these Dates stdlib changes:

- https://github.com/JuliaLang/julia/pull/30200
- https://github.com/JuliaLang/julia/pull/30817

### DIFF
--- a/src/utcoffset.jl
+++ b/src/utcoffset.jl
@@ -64,6 +64,7 @@ function Base.show(io::IO, o::UTCOffset)
         # daylight saving time offsets and exotic DST offsets (e.g. midsummer time).
         print(io, "UTC", offset_string(o.std), "/", offset_string(o.dst))
     else
+        # Fallback to calling the default show instead of reimplementing it.
         invoke(show, Tuple{IO, Any}, io, o)
     end
 end
--- a/src/utcoffset.jl
+++ b/src/utcoffset.jl
@@ -57,8 +57,17 @@ function offset_string(offset::UTCOffset, iso8601::Bool=false)
 end
 
 Base.print(io::IO, o::UTCOffset) = print(io, offset_string(o, true))
+
 function Base.show(io::IO, o::UTCOffset)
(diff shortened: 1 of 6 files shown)


test/helpers.jl
```diff
@@ -28,3 +28,15 @@ function ignore_output(body::Function; stdout::Bool=true, stderr::Bool=true)
 
     return result
 end
+
+# Alternatively in Julia 0.7.0-DEV.4517 we could use
```
iamed2: Aren't all the versions you use higher than this?

omus: They are. I've decided to leave this in as:
```julia
sprint(show, args..., context=:compact => true)
```
is a bit long compared to:
```julia
sprint(show_compact, args...)
```

iamed2: Maybe remove the "in Julia 0.7.0-DEV.4517" then? The comment currently makes it sound like you're planning to replace it.

src/utcoffset.jl
```diff
@@ -57,8 +57,17 @@ function offset_string(offset::UTCOffset, iso8601::Bool=false)
...
 function Base.show(io::IO, o::UTCOffset)
-    # Show DST as a separate offset since we want to distinguish between normal hourly
-    # daylight saving time offsets and exotic DST offsets (e.g. midsummer time).
-    print(io, "UTC", offset_string(o.std), "/", offset_string(o.dst))
+    if get(io, :compact, false)
+        # Show DST as a separate offset since we want to distinguish between normal hourly
+        # daylight saving time offsets and exotic DST offsets (e.g. midsummer time).
+        print(io, "UTC", offset_string(o.std), "/", offset_string(o.dst))
+    else
+        invoke(show, Tuple{IO, Any}, io, o)
```
rofinn: Seems a little weird that your invoking the default show, might be good to include a comment.

rofinn: Same goes for elsewhere.

PR #199, opened 2019-05-08
Remove outdated PkgEval badges

iamed2:
They don't point to anything anymore and AFAIK there isn't a replacement


omus:
They had a good run.

PR #188, opened 2019-03-14
Fix for 32-bit "uU" dateformat issue

fchorney:
closes #187

PR #207, opened 2019-07-19
Fix UndefVarError in validargs

iamed2:
Error which was fixed:

```
  UndefVarError: str not defined
  Stacktrace:
   [1] validargs(::Type{ZonedDateTime}, ::Int64, ::Int64, ::Int64, ::Int64, ::Int64, ::Int64, ::Int64, ::String) at /Users/ericdavies/.julia/packages/TimeZones/Z0kpK/src/types/zoneddatetime.jl:177
   [2] tryparse(::Type{ZonedDateTime}, ::String, ::DateFormat{Symbol("yyyy-mm-dd HH:MM:SS.sss Z"),Tuple{Dates.DatePart{'y'},Dates.Delim{Char,1},Dates.DatePart{'m'},Dates.Delim{Char,1},Dates.DatePart{'d'},Dates.Delim{Char,1},Dates.DatePart{'H'},Dates.Delim{Char,1},Dates.DatePart{'M'},Dates.Delim{Char,1},Dates.DatePart{'S'},Dates.Delim{Char,1},Dates.DatePart{'s'},Dates.Delim{Char,1},Dates.DatePart{'Z'}}}) at /Users/ericdavies/repos/julia1p1/usr/share/julia/stdlib/v1.1/Dates/src/parse.jl:293
```

For some reason this only happened when _no_ time zone was present, and not when an invalid time zone was present. But AFAICT neither case was being tested.

### DIFF
--- a/src/types/zoneddatetime.jl
+++ b/src/types/zoneddatetime.jl
@@ -116,9 +116,13 @@ end
     ZonedDateTime(DateTime(y,m,d,h,mi,s,ms), tz)
 end
 
-# Parsing constructor. Note we typically don't support passing in time zone information as a
-# string since we cannot do not know if we need to support resolving ambiguity.
-# Month Union issue: https://github.com/JuliaTime/TimeZones.jl/issues/187#issuecomment-473012078
+# Parsing constructor needed as part of the Dates parsing interface. Note we typically don't
+# support passing in time zone information as a string since we cannot do not know if we
+# need to support resolving ambiguity.
+#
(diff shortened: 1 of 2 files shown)


omus:
Looks like your change caught something that was missed in https://github.com/JuliaTime/TimeZones.jl/pull/188. Basically when using the "u" or "U" specifier an `Int32` will be passed around on 32-bit. The PR I referenced corrected this for `parse` but not `tryparse`.

PR #212, opened 2019-09-09
Support Mocking 0.7

omus:
Needed to drop support for Julia 0.7 as this is not supported by Mocking 0.7.


omus:
I'll work on getting Julia 1.2 working in: https://github.com/JuliaTime/TimeZones.jl/pull/213

omus:
Avoiding bumping the lower bound of Mocking to avoid conflicts with downstream packages.

PR #214, opened 2019-09-10
Default to using 2019b tzdata

omus:
Release notes: https://mm.icann.org/pipermail/tz-announce/2019-July/000056.html

```
# https://github.com/unicode-org/cldr/blob/master/common/supplemental/windowsZones.xml
cd deps/local
curl -vR https://raw.githubusercontent.com/unicode-org/cldr/master/common/supplemental/windowsZones.xml -o windowsZonesLatest.xml
cp -p windowsZonesLatest.xml windowsZones2019b.xml  # Once verified to contain new data
mv windowsZonesLatest.xml windowsZones.xml
```

Currently, 2019b fails to build due to this rule:
```
Rule	Zion	2005	2012	-	Apr	Fri<=1	2:00	1:00	D
```
The internal logic tries to find a Friday in April which is before April 1st. This exists on 2005 but not 2006. According to the documentation:

```
# The proposed law agreed upon by the Knesset Interior Committee on
# 2005-02-14 is that, for 2005 and beyond, DST starts at 02:00 the
# last Friday before April 2nd (i.e. the last Friday in March or April
# 1st itself if it falls on a Friday) and ends at 02:00 on the Saturday
```
I'll need to update some of the tzdata compilation code to handle this.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -3,4 +3,4 @@ import TimeZones: build
 # ENV variable allows users to modify the default to be "latest". Do NOT use "latest"
 # as the default here as can make it difficult to debug to past versions of working code.
 # Note: Also allows us to only download a single tzdata version during CI jobs.
-build(get(ENV, "JULIA_TZ_VERSION", "2019a"))
+build(get(ENV, "JULIA_TZ_VERSION", "2019b"))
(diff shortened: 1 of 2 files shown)


omus:
Requires: https://github.com/JuliaTime/TimeZones.jl/pull/219

Issue #218, opened 2019-09-24
Julia nightly for Windows no longer includes 7z


omus:
See: https://ci.appveyor.com/project/omus/timezones-jl/builds/27616089/job/1si5157e27obt2r8

I believe this was changed in: https://github.com/JuliaLang/julia/pull/33125

omus:
Fixed by: https://github.com/JuliaTime/TimeZones.jl/pull/230

PR #219, opened 2019-09-26
Fix handling of ON rules using `<=`

omus:
The 2019b tzdata includes an ON rule "Fri<=1" which revealed that our handling of `<=` rules to be incorrect. Our handling of `>=` rules appears to always have been handled correctly.

Due to us previously assuming the ON rule would always result in a date within the IN month some major modifications needed to be made to properly support this.

For context here is the rule which revealed the issue from the 2019b tzdata:
```
# The proposed law agreed upon by the Knesset Interior Committee on
# 2005-02-14 is that, for 2005 and beyond, DST starts at 02:00 the
# last Friday before April 2nd (i.e. the last Friday in March or April
# 1st itself if it falls on a Friday) and ends at 02:00 on the Saturday
# night _before_ the fast of Yom Kippur.
#
# Those who can read Hebrew can view the announcement at:
#
#	ftp://ftp.cs.huji.ac.il/pub/tz/announcements/2005+beyond.ps

# Rule	NAME	FROM	TO	TYPE	IN	ON	AT	SAVE	LETTER/S
Rule	Zion	2005	2012	-	Apr	Fri<=1	2:00	1:00	D
```

### DIFF
--- a/docs/src/api-private.md
+++ b/docs/src/api-private.md
@@ -18,7 +18,7 @@ TimeZones.TZData.tzdata_version_dir
 TimeZones.TZData.tzdata_version_archive
 TimeZones.TZData.read_news
 TimeZones.TZData.compile!
-TimeZones.TZData.tryparse_dayofmonth
+TimeZones.TZData.tryparse_dayofmonth_function
 TimeZones.TZData.order_rules
 ```
 
--- a/src/tzdata/compile.jl
+++ b/src/tzdata/compile.jl
@@ -96,10 +96,10 @@ const DAYS = Dict(
 
 const LAST_DAY_OF_WEEK = Dict{String, Function}()
 
-# Create adjuster functions such as `last_sunday`.
+# Create adjuster functions such as `islastsunday`.
 for (abbr, dayofweek) in DAYS
-    str = "last" * abbr
-    f = Symbol("last_" * lowercase(dayname(dayofweek)))
+    str = "last" * abbr  # e.g. "lastSun"
+    f = Symbol("islast" * lowercase(dayname(dayofweek)))  # e.g. :islastsunday
     LAST_DAY_OF_WEEK[str] = @eval begin
         function $f(dt)
             return dayofweek(dt) == $dayofweek &&
--- a/src/tzdata/compile.jl
+++ b/src/tzdata/compile.jl
@@ -94,17 +94,32 @@ const DAYS = Dict(
     "Mon" => 1, "Tue" => 2, "Wed" => 3, "Thu" => 4, "Fri" => 5, "Sat" => 6, "Sun" => 7,
 )
 
-const LAST_DAY_OF_WEEK = Dict{String, Function}()
+const LAST_WEEKDAY_OF_MONTH = Dict{String, Function}()
 
-# Create adjuster functions such as `islastsunday`.
+# Create functions such as `is_last_sunday` and `last_sunday_of_month`.
 for (abbr, dayofweek) in DAYS
-    str = "last" * abbr  # e.g. "lastSun"
-    f = Symbol("islast" * lowercase(dayname(dayofweek)))  # e.g. :islastsunday
-    LAST_DAY_OF_WEEK[str] = @eval begin
-        function $f(dt)
+    on_str = "last" * abbr  # e.g. "lastSun"
+    weekday = dayname(dayofweek)  # e.g. "Sunday"
+    is_last_weekday = Symbol("is_last_", lowercase(weekday))
+    last_weekday_of_month = Symbol("last_", lowercase(weekday), "_of_month")
+
+    LAST_WEEKDAY_OF_MONTH[on_str] = @eval begin
+        function $is_last_weekday(dt)
             return dayofweek(dt) == $dayofweek &&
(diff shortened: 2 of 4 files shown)


omus:
Documentation is failing due to referencing the old `tryparse_dayofmonth` function name

omus:
Allowing failures on Windows nightly to work around: https://github.com/JuliaTime/TimeZones.jl/issues/218

docs/src/api-private.md
```diff
@@ -18,7 +18,7 @@ TimeZones.TZData.tzdata_version_dir
 TimeZones.TZData.tzdata_version_archive
 TimeZones.TZData.read_news
 TimeZones.TZData.compile!
-TimeZones.TZData.tryparse_dayofmonth
+TimeZones.TZData.tryparse_dayofmonth_function
```
nicoleepp: Seems like a weird change to just add `function` to the end

nicoleepp: To make this more clear I would prefer `gen_tryparse_dayofmonth_function` but it is already a really long function name so no need to change.

src/tzdata/compile.jl
```diff
@@ -118,41 +133,62 @@ end
...
 ```
 """
-function tryparse_dayofmonth(str::AbstractString)
+function tryparse_dayofmonth_function(str::AbstractString)
     if occursin(r"^last\w{3}$", str)
         # We pre-built these functions above
         # They follow the format: "lastSun", "lastMon", etc.
-        LAST_DAY_OF_WEEK[str]
-    elseif (m = match(r"^(?<dow>\w{3})(?<op>[<>]=)(?<dom>\d{1,2})$", str)) !== nothing
+        LAST_WEEKDAY_OF_MONTH[str]
```
nicoleepp: Were we just using the wrong function for what this regex actually is?

src/tzdata/compile.jl
```diff
@@ -118,41 +133,62 @@ end
...
+            date = Date(year, month, dom)
+            tonext(d -> dayofweek(d) == dow, date; step=step, same=true, limit=7)
         end
     elseif occursin(r"^\d{1,2}$", str)
-        # Matches just a plain old day of the month
+        # Matches just a simple day of the month
         dom = parse(Int, str)
-        dt -> day(dt) == dom
+
+        function (year, month)
```
nicoleepp: Should use explicit `returns`

src/tzdata/compile.jl
```diff
@@ -188,16 +225,27 @@ function parse_date(s::AbstractString)
...
-    dt = DateTime(periods...)
 
     # Adjust the DateTime to reflect the requirements of the day-of-month function.
+    # Note: `numeric_dom` will only be `false` when `dom_str` is not-empty which implies
+    # there are at least 3 elements within `periods` (year, month, day).
     if !numeric_dom
-        dom = tryparse_dayofmonth(dom_str)
-        dom !== nothing || throw(ArgumentError("Unable to parse day-of-month: \"$dom_str\""))
-        dt = tonext(dom, dt; step=Day(1), same=true)
+        dom_func = tryparse_dayofmonth_function(dom_str)
```
nicoleepp: What does `dom_str` stand for?

docs/src/api-private.md
omus: While revising this code I thought of `tryparse_dayofmonth` as a function that would return an integer. I added "_function" to be more explicit that a function is returned. Additionally, I think adding a "gen" prefix doesn't add much since `parse` here means: convert a string into a Julia representation. Also, the internals here could be changed to return a pre-created function instead of a closure.

src/tzdata/compile.jl
omus: No, what was changed here was:
- `LAST_DAY_OF_WEEK` was renamed to `LAST_WEEKDAY_OF_MONTH`
- Part of the regex `[<>]=` was changed to `<=|>=` to try and be clearer

The diff is doing a bad job here

src/tzdata/compile.jl
omus: Day-of-month as a string

src/tzdata/compile.jl
nicoleepp: Yeah I guess my question is more so why did we rename `LAST_DAY_OF_WEEK` to `LAST_WEEKDAY_OF_MONTH` as the functionality they imply is different

omus: I think I understand the confusion. The original constant `LAST_DAY_OF_WEEK` was incorrectly named and should have been called: `IS_LAST_WEEKDAY_OF_MONTH`. Additionally, the function was updated to  return a `Date` instead of a `Bool` which is why the `IS` prefix was dropped

src/tzdata/compile.jl
omus: Will do

omus:
I just added the explicit return and added type assertions to the simple day of month anonymous function for consistency.

omus:
Will wait for CI and then will merge

omus:
> May want to include a version bump with this

I'll hold off as I want to do that in: https://github.com/JuliaTime/TimeZones.jl/pull/214

PR #221, opened 2019-10-01
add style-guide badge

oxinabox:


omus:
I'll have to double check this repo is following the style guide

oxinabox:
It aspires to. And that is what matters

nickrobinson251:
the only departure from it that i can see is seom functions not using the `return` keyword

oxinabox:
Let's merge this

PR #227, opened 2019-11-01
trailing 'Z' support

gfZeng:
`ZonedDateTime("2019-11-01T15:09:13.978Z")` now throws

trailing `Z` should be equivalent to `+00:00`

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -1,6 +1,8 @@
 using Dates: DateFormat, DatePart, min_width, max_width
 
 function tryparsenext_fixedtz(str, i, len, min_width::Int=1, max_width::Int=0)
+    i == len && str[i] == 'Z' && return ("Z", i+1)
+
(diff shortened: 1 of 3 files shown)


src/parse.jl
```diff
@@ -1,6 +1,9 @@
 using Dates: DateFormat, DatePart, min_width, max_width
 
 function tryparsenext_fixedtz(str, i, len, min_width::Int=1, max_width::Int=0)
+
+    str[i] == 'Z' && return ("Z", i+len)
```
omus: I believe this may interfere when parsing `Zulu`? We should probably verify this is the last character in the string

gfZeng:
PR updated

IMO, `UTC_ZERO` better than `ZERO_UTC`

gfZeng:
> Could you add some tests? Thanks for this!

PR updated

PR #228, opened 2019-11-04
Improve messaging when TimeZones fails to build

omus:
Avoid confusing users with:
```
ERROR: InitError: ArgumentError: Unknown time zone "Europe/London"
```
which is usually caused from a failed build. Unfortunately Julia 1.0 hides the log messages which makes it very non-obvious what is going on.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,6 +1,3 @@
-import TimeZones: build
+using TimeZones: build
 
-# ENV variable allows users to modify the default to be "latest". Do NOT use "latest"
-# as the default here as can make it difficult to debug to past versions of working code.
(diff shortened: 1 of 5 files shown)


iamed2:
If you run TimeZones.build with a specific version, then, run `]build TimeZones`, which version will you have?

omus:
> If you run TimeZones.build with a specific version, then, run `]build TimeZones`, which version will you have?

As part of this change the behaviour here is now consistent. Both `TimeZones.build()` and `]build TimeZones` now perform identical operations. Previously `]build TimeZones` would have built a specified revision and `TimeZones.build()` would have built the "latest" revision.

PR #230, opened 2019-11-18
Fix 7zip path on Julia 1.3

oxinabox:


oxinabox:
Need to check CI is setup right for app-veyor

Issue #125, opened 2018-02-09
Names of `utc` and `localtime`


omus:
Currently TimeZones has the functions `utc` and `localtime` to which convert a `ZonedDateTime` into a UTC `DateTime` or a local `DateTime`.

I've never really liked the names of these functions. Particularly `localtime` as I would prefer `local` but unfortunately since that is a keyword it is out as an option. One alternative would be to use `wall` which would stand for "wall time" which is a term used in the tzdata documentation.

Additionally, these functions currently return only a `DateTime` and we may want to allow users to return a `Date` directly instead of having to do `Date(localtime(zdt))`. Possibly we could do something like:

```julia
utc(DateTime, zdt)
localtime(DateTime, zdt)
wall(DateTime, zdt)

utc(Date, zdt)
localtime(Date, zdt)
wall(Date, zdt)
```

Alternatively we could do:

```julia
Date(zdt, :wall)
# Or
Date(zdt, Wall)
```

Possibly https://github.com/JuliaTime/TimeZones.jl/issues/121 will come into play here.

Issue #209, opened 2019-07-29
Overload `ZoneDataTime(::Date, ::AbstractTimeZone)`


oxinabox:
I think that we should have

```
function ZonedDateTime(date::Date, timezone::AbstractTimeZone)
    return ZonedDateTime(DateTime(date), timezone)
end
```

omus:
We'd need to still deal with DST but it could be added. I'll note that there are plans for a time zone aware Date

omus:
No, this is complete

Issue #184, opened 2019-02-08
Add plotting recipie for ZonedDateTime


oxinabox:
I plot a lot of `ZonedDateTimes`
and I am getting tired of calling `DataeTime` on them first.

As i understand it would could add a RecipiesBase.jl recipie for them,
and they would just plot, and that would be great.

RecipiesBase is a pretty lightweight dependency.

omus:
I think this is a great idea

glennmoy:
To chime in. This is a MWE of what we would like to implement:

```
using Dates
using TimeZones

timezone = tz"EST"
start_date = DateTime(2017,1,1,0,0,0);
end_date = DateTime(2017,1,1,10,30,0);
date_range = start_date:Hour(1):end_date

dates = ZonedDateTime.(date_range, Ref(timezone))
date_ticks = DateTime.(TimeZones.astimezone.(dates, Ref(tz"UTC")))

>11-element Array{DateTime,1}:
 2017-01-01T05:00:00
 2017-01-01T06:00:00
 2017-01-01T07:00:00
 2017-01-01T08:00:00
 2017-01-01T09:00:00
 2017-01-01T10:00:00
 2017-01-01T11:00:00
 2017-01-01T12:00:00
 2017-01-01T13:00:00
 2017-01-01T14:00:00
 2017-01-01T15:00:00

```

daschw:
Here's the Date(Time)s recipe implementation of Plots.jl: https://github.com/JuliaPlots/Plots.jl/blob/master/src/recipes.jl#L1095-L1104.
A recipe for ZonedDateTimes could look something like
```julia
@recipe function f(::Type{T}, val::T) where{T<:ZonedDateTime} = (dt -> Dates.value(dt.utc_datetime), Plots.datetimeformatter)
```

oxinabox:
I am going to work on this today

daschw:
Let me know if I can support you.

PR #231, opened 2019-11-21
Date <-> ZonedDateTime

### LINKED ISSUE #222: Add Date(::ZonedDateTime) method
There is a `DateTime(::ZonedDateTime)` method, but no `Date(::ZonedDateTime)` method:

```julia
julia> using Dates, TimeZones

julia> t = ZonedDateTime(2014, 7, 15, 14, 35, 46, tz"America/Chicago")
2014-07-15T14:35:46-05:00

julia> DateTime(t)
2014-07-15T14:35:46

julia> Date(t)
ERROR: MethodError: Cannot `convert` an object of type ZonedDateTime to an object of type Date
Closest candidates are:
  convert(::Type{Date}, ::DateTime) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.1/Dates/src/conversions.jl:31
  convert(::Type{Date}, ::Day) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.1/Dates/src/conversions.jl:36
  convert(::Type{T}, ::T) where T at essentials.jl:154
  ...
Stacktrace:
 [1] Date(::ZonedDateTime) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.1/Dates/src/conversions.jl:12
 [2] top-level scope at none:0
```

Desired behavior:

```julia
julia> Date(zdt::ZonedDateTime) = Date(DateTime(zdt))
Date

julia> Date(t)
2014-07-15
```

In other words, I expect that `DateTime` truncates the time zone information from a zoned date time and `Date` truncates both the time zone information and the hour-minute-seconds information.

oxinabox:
I got really tired of writing `Date(DateTime(zdt))` and `ZonedDateTime(DateTime(date))`

Fixes #222 and #209 

```
julia> zdt = ZonedDateTime(Date(2011, 6, 1), tz"America/Winnipeg")
2011-06-01T00:00:00-05:00

julia> Date(zdt)
2011-06-01

julia> DateTime(Date(2011, 6, 1))
2011-06-01T00:00:00
```

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -4,10 +4,12 @@ using Dates
 using Printf
 using Serialization
 using Unicode
-import Dates: TimeZone
+
+import Dates: TimeZone, UTC
 
 export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDateTime,
-    DateTime, TimeError, AmbiguousTimeError, NonExistentTimeError, UnhandledTimeError,
+    DateTime, Date, Time, UTC, Local, TimeError, AmbiguousTimeError, NonExistentTimeError,
+    UnhandledTimeError,
     # discovery.jl
     timezone_names, all_timezones, timezones_from_abbr, timezone_abbrs,
     next_transition_instant, show_next_transition,
--- a/src/adjusters.jl
+++ b/src/adjusters.jl
@@ -6,10 +6,10 @@ using Dates: DatePeriod, TimePeriod, firstdayofweek, lastdayofweek,
 # Truncation
 # TODO: Just utilize floor code for truncation?
 function Base.trunc(zdt::ZonedDateTime, ::Type{P}) where P <: DatePeriod
-    ZonedDateTime(trunc(localtime(zdt), P), timezone(zdt))
+    ZonedDateTime(trunc(DateTime(zdt, Local), P), timezone(zdt))
 end
 function Base.trunc(zdt::ZonedDateTime, ::Type{P}) where P <: TimePeriod
-    local_dt = trunc(localtime(zdt), P)
+    local_dt = trunc(DateTime(zdt, Local), P)
     utc_dt = local_dt - zdt.zone.offset
     ZonedDateTime(utc_dt, timezone(zdt); from_utc=true)
 end
@@ -18,7 +18,7 @@ Base.trunc(zdt::ZonedDateTime, ::Type{Millisecond}) = zdt
 # Adjusters
 for prefix in ("firstdayof", "lastdayof"), suffix in ("week", "month", "year", "quarter")
     func = Symbol(prefix * suffix)
-    @eval begin
-        Dates.$func(dt::ZonedDateTime) = ZonedDateTime($func(localtime(dt)), dt.timezone)
+    @eval function Dates.$func(zdt::ZonedDateTime)
+        ZonedDateTime($func(DateTime(zdt, Local)), zdt.timezone)
     end
 end
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -5,9 +5,9 @@ using Mocking: Mocking, @mock
 const utc_tz = FixedTimeZone("UTC")
 
 """
-    DateTime(::ZonedDateTime, ::Type{Local}) -> DateTime
+    DateTime(zdt::ZonedDateTime, ::Union{Type{Local}, Type{UTC}}) -> DateTime
 
-Create an implicit local time `DateTime` from the given `ZonedDateTime`.
+Create an implicit local/UTC `DateTime` from the given `ZonedDateTime`.
 
 # Example
 
@@ -17,33 +17,21 @@ julia> zdt = ZonedDateTime(2014, 5, 30, 21, tz"UTC-4")
 
 julia> DateTime(zdt, Local)
 2014-05-30T21:00:00
-```
-"""
-Dates.DateTime(zdt::ZonedDateTime, ::Type{Local}) = zdt.utc_datetime + zdt.zone.offset
-
-
-"""
-    DateTime(::ZonedDateTime, ::Type{UTC}) -> DateTime
-
-Create an implicit utc `DateTime` from the given `ZonedDateTime`.
-
-# Example
-
-```jldoctest
-julia> zdt = ZonedDateTime(2014, 5, 30, 21, tz"UTC-4")
-2014-05-30T21:00:00-04:00
 
 julia> DateTime(zdt, UTC)
 2014-05-31T01:00:00
 ```
 """
+DateTime(::ZonedDateTime, ::Union{Type{Local}, Type{UTC}})
+
+Dates.DateTime(zdt::ZonedDateTime, ::Type{Local}) = zdt.utc_datetime + zdt.zone.offset
 Dates.DateTime(zdt::ZonedDateTime, ::Type{UTC}) = zdt.utc_datetime
 
 
 """
-    Date(::ZonedDateTime, ::Type{Local}) -> DateTime
+    Date(zdt::ZonedDateTime, ::Union{Type{Local}, Type{UTC}}) -> Date
 
-Create an implicit local time `Date` from the given `ZonedDateTime`.
+Create an implicit local/UTC `Date` from the given `ZonedDateTime`.
 
 # Example
 
@@ -53,33 +41,21 @@ julia> zdt = ZonedDateTime(2014, 5, 30, 21, tz"UTC-4")
 
 julia> Date(zdt, Local)
 2014-05-30
-```
-"""
-Dates.Date(zdt::ZonedDateTime, ::Type{Local}) = Date(DateTime(zdt, Local))
-
-
-"""
-    Date(::ZonedDateTime, ::Type{UTC}) -> DateTime
-
-Create an implicit utc `Date` from the given `ZonedDateTime`.
-
-# Example
-
-```jldoctest
-julia> zdt = ZonedDateTime(2014, 5, 30, 21, tz"UTC-4")
-2014-05-30T21:00:00-04:00
 
 julia> Date(zdt, UTC)
 2014-05-31
 ```
 """
+Date(::ZonedDateTime, ::Union{Type{Local}, Type{UTC}})
+
+Dates.Date(zdt::ZonedDateTime, ::Type{Local}) = Date(DateTime(zdt, Local))
 Dates.Date(zdt::ZonedDateTime, ::Type{UTC}) = Date(DateTime(zdt, UTC))
 
 
 """
-    Time(::ZonedDateTime, ::Type{Local}) -> Time
+    Time(zdt::ZonedDateTime, ::Union{Type{Local}, Type{UTC}}) -> Time
 
-Create an implicit local `Time` from the given `ZonedDateTime`.
+Create an implicit local/UTC `Time` from the given `ZonedDateTime`.
 
 # Example
 
@@ -89,26 +65,14 @@ julia> zdt = ZonedDateTime(2014, 5, 30, 21, tz"UTC-4")
 
 julia> Time(zdt, Local)
 21:00:00
-```
-"""
-Dates.Time(zdt::ZonedDateTime, ::Type{Local}) = Time(DateTime(zdt, Local))
-
-
-"""
-    Time(::ZonedDateTime, ::Type{UTC}) -> DateTime
-
-Create an implicit utc `Time` from the given `ZonedDateTime`.
-
-# Example
-
-```jldoctest
-julia> zdt = ZonedDateTime(2014, 5, 30, 21, tz"UTC-4")
-2014-05-30T21:00:00-04:00
 
 julia> Time(zdt, UTC)
 01:00:00
 ```
 """
+Time(::ZonedDateTime, ::Union{Type{Local}, Type{UTC}})
+
+Dates.Time(zdt::ZonedDateTime, ::Type{Local}) = Time(DateTime(zdt, Local))
 Dates.Time(zdt::ZonedDateTime, ::Type{UTC}) = Time(DateTime(zdt, UTC))
 
 
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -5,18 +5,111 @@ using Mocking: Mocking, @mock
 const utc_tz = FixedTimeZone("UTC")
 
 """
-    DateTime(::ZonedDateTime) -> DateTime
+    DateTime(::ZonedDateTime, ::Type{Local}) -> DateTime
 
-Returns an equivalent `DateTime` without any `TimeZone` information.
+Create an implicit local time `DateTime` from the given `ZonedDateTime`.
+
+# Example
+
+```jldoctest
+julia> zdt = ZonedDateTime(2014, 5, 30, 21, tz"UTC-4")
+2014-05-30T21:00:00-04:00
+
+julia> DateTime(zdt, Local)
+2014-05-30T21:00:00
+```
 """
-DateTime(zdt::ZonedDateTime) = localtime(zdt)
+Dates.DateTime(zdt::ZonedDateTime, ::Type{Local}) = zdt.utc_datetime + zdt.zone.offset
+
 
 """
-    Date(::ZonedDateTime) -> Date
+    DateTime(::ZonedDateTime, ::Type{UTC}) -> DateTime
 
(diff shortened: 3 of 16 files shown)


src/conversions.jl
```diff
@@ -11,6 +11,14 @@ Returns an equivalent `DateTime` without any `TimeZone` information.
 """
 DateTime(zdt::ZonedDateTime) = localtime(zdt)
 
+"""
+    Date(::ZonedDateTime) -> Date
+
+Returns an equivalent `Date` without any `TimeZone`, or time information.
+"""
+Dates.Date(zdt::ZonedDateTime) = Date(DateTime(zdt))
```
omus: I've been intending on dropping `DateTime(::ZonedDateTime)` in favor of `localtime(::Type{DateTime}, ::ZonedDateTime)` or something similar. This is definitely a lossy operation and it's important to communicate the implicit time zone.

test/conversions.jl
```diff
@@ -6,28 +6,34 @@ warsaw = first(compile("Europe/Warsaw", tzdata["europe"]))
...
-zdt = ZonedDateTime(dt, warsaw)
-@test DateTime(zdt) == dt
-
-# Converting from ZonedDateTime to DateTime isn't possible as it is always inexact.
-@test_throws MethodError convert(DateTime, zdt)
-
-# Vectorized accessors
-n = 10
-arr = fill(zdt, n)
-@test Dates.DateTime.(arr) == fill(dt, n)
```
fchorney: we don't want this test anymore?

test/conversions.jl
```diff
@@ -6,28 +6,34 @@ warsaw = first(compile("Europe/Warsaw", tzdata["europe"]))
...
-# now function
-dt = now(Dates.UTC)::DateTime
-zdt = now(warsaw)
-@test zdt.timezone == warsaw
-@test Dates.datetime2unix(TimeZones.utc(zdt)) ≈ Dates.datetime2unix(dt)
-
-# today function
-@test abs(today() - today(warsaw)) <= Dates.Day(1)
-@test today(apia) - today(midway) == Dates.Day(1)
+@testset "Between ZonedDateTime and DateTime" begin
```
omus:
```suggestion
@testset "Construct ZonedDateTime / DateTime" begin
```

test/conversions.jl
```diff
@@ -6,28 +6,34 @@ warsaw = first(compile("Europe/Warsaw", tzdata["europe"]))
...
+    # Constructing a ZonedDateTime from a DateTime and the reverse
+    dt = DateTime(2015, 1, 1, 0)
+    zdt = ZonedDateTime(dt, warsaw)
+    @test DateTime(zdt) == dt
+
+    # Converting from ZonedDateTime to DateTime isn't possible as it is always inexact.
+    @test_throws MethodError convert(DateTime, zdt)
+end
+
+@testset "Between Date and ZonedDateTime" begin
```
omus:
```suggestion
@testset "Construct Date / ZonedDateTime" begin
```

test/conversions.jl
oxinabox: Isn't this just testing that broadcasting works?
Testing that broadcasting works on arrays is not our problem.

This test is leftover from ancient times  (https://github.com/JuliaTime/TimeZones.jl/pull/7)
when `DateTime(::Vector{ZonedDateTime})` was defined.
And so needed tests.
But we got rid of that because julia gained broadcasting,
and I guess the test just got converted over to use the new broadcasting,
without realizing that the code it was suppose to test didn't exist anymore.

src/conversions.jl
omus: https://github.com/JuliaTime/TimeZones.jl/issues/125

oxinabox: I would really dislike that.
Often I just have a collection of `DateTime`,
one per day.
and I want to plot them (cf #184),
or I want to display them in a table,
and I just want to get rid of extra information.
I don't actually care about the time or even the exact date often.

I think it pretty clear to the user that  it is losing information.

omus: Looking at that issue again `Date(zdt, UTC)` / `Date(zdt, Local)` isn't bad

test/conversions.jl
omus: We can drop this code now

src/conversions.jl
oxinabox: I am happy to see this change when we resolve #125 for `DateTime`s.
Til then this match the API we have

omus: Let's just make the change now? Both the `UTC` and `Local` types already exist

omus: If you don't want to take on the change yourself I can make the remainder of the changes. I don't think I would merge this as is.

oxinabox: I do want to merge as is, because the status quo of matching how `DateTime` works is known.
All other things are not known and we might want to do a little design and thinking.

### OTOH:
 - `Date(zdt, Local)` and `Date(zdt, UTC)` // `DateTime(zdt, Local)` and `DateTime(zdt, UTC)` do sound right to me though.
 - A bunch of places that do `DateTime(ztd)` will need to be changed due to deprecation. and any places that do `Date(DateTime(ztd))` will be a subset of those. So doing both at once would mean  we can go staight to `Date(ztd, Local)`. 

I think I am convinced.

omus: > I think I am convinced

Convinced enough to do the work? ;). As I said I can do the leg work if you prefer.

oxinabox: I'm not sure how deep I would have to go,
so if you know exactly what you have to change be my guest.

Otherwise this might sit til after neurips

oxinabox: Waiting til after NeurIPS might be fine though.
We've lived this long

omus:
Alright, I've made the updates and bumped the version to 1.0.0 (it's time)

src/TimeZones.jl
```diff
@@ -4,10 +4,12 @@ using Dates
 using Printf
 using Serialization
 using Unicode
-import Dates: TimeZone
+
+import Dates: TimeZone, UTC
```
oxinabox: Lets not make more of this violation of BlueStyle?

```suggestion
import Dates: TimeZone
using: Dates: UTC
```

src/adjusters.jl
```diff
@@ -6,10 +6,10 @@ using Dates: DatePeriod, TimePeriod, firstdayofweek, lastdayofweek,
 # Truncation
 # TODO: Just utilize floor code for truncation?
 function Base.trunc(zdt::ZonedDateTime, ::Type{P}) where P <: DatePeriod
-    ZonedDateTime(trunc(localtime(zdt), P), timezone(zdt))
+    ZonedDateTime(trunc(DateTime(zdt, Local), P), timezone(zdt))
```
oxinabox:
```suggestion
    return ZonedDateTime(trunc(DateTime(zdt, Local), P), timezone(zdt))
```

src/adjusters.jl
```diff
@@ -18,7 +18,7 @@ Base.trunc(zdt::ZonedDateTime, ::Type{Millisecond}) = zdt
 # Adjusters
 for prefix in ("firstdayof", "lastdayof"), suffix in ("week", "month", "year", "quarter")
     func = Symbol(prefix * suffix)
-    @eval begin
-        Dates.$func(dt::ZonedDateTime) = ZonedDateTime($func(localtime(dt)), dt.timezone)
+    @eval function Dates.$func(zdt::ZonedDateTime)
+        ZonedDateTime($func(DateTime(zdt, Local)), zdt.timezone)
```
oxinabox:
```suggestion
        return ZonedDateTime($func(DateTime(zdt, Local)), zdt.timezone)
```

src/adjusters.jl
```diff
@@ -18,7 +18,7 @@ Base.trunc(zdt::ZonedDateTime, ::Type{Millisecond}) = zdt
 # Adjusters
 for prefix in ("firstdayof", "lastdayof"), suffix in ("week", "month", "year", "quarter")
     func = Symbol(prefix * suffix)
```
oxinabox: Not related but is it cleaner to do:

```suggestion
    func = Symbol(prefix, suffix)
```

src/deprecated.jl
```diff
@@ -1,7 +1,9 @@
-using Base: @deprecate, @deprecate_binding
+using Base: @deprecate
 
-# BEGIN TimeZones 0.9 deprecations
+# BEGIN TimeZones 1.0 deprecations
```
oxinabox: 1.0, huh

src/TimeZones.jl
omus: How does this violate it?

src/deprecated.jl
omus: It's time. I usually don't make API breaking changes but I use minor versions to indicate features.

src/TimeZones.jl
oxinabox: We have a guideline against using `import` ever.

omus: In this particular case I like the use of `import` as these are concepts that would make sense to be defined in TimeZones.jl but are currently in the Dates stdlib. By using `import` I'm explicitly calling out the special relationship here.

src/conversions.jl
```diff
@@ -90,31 +155,31 @@ function astimezone(zdt::ZonedDateTime, tz::FixedTimeZone)
...
 end
 
 function zdt2unix(::Type{T}, zdt::ZonedDateTime) where T<:Integer
-    floor(T, datetime2unix(utc(zdt)))
+    floor(T, datetime2unix(DateTime(zdt, UTC)))
 end
 
 function zdt2unix(::Type{T}, zdt::ZonedDateTime) where T<:Real
-    convert(T, datetime2unix(utc(zdt)))
+    convert(T, datetime2unix(DateTime(zdt, UTC)))
```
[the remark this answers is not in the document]
omus: Let's keep the addition of `return` statements to a separate PR otherwise this PR could become held up.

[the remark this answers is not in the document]
omus: https://github.com/invenia/BlueStyle/issues/43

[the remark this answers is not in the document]
oxinabox: I don't see why clicking "apply" would hold this up.
No need to go do the others I've missed but might as well get these ones.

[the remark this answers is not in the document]
omus: What I mean is I'm good with these changes without the applying the `return` statements

omus:
I squashed some commits in preparation for merging. A couple of things to note:
- Adding return statements will not be done in this PR as I think that is a separate conversation
- We're bumping to version 1.0 here as there haven't been any major API changes for a while and the ones I have planned can always go into 2.0

src/conversions.jl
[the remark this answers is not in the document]
oxinabox: what I mean is:
Rather than having this discussion, you can click "Commit suggestion" above.
And that will take less time than concluding the discussion.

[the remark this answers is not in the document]
oxinabox: Transfering from slack:

@omus  said:
> In the case of a disagreement follow the status quo of what is currently present in the repo 

I said:
> That is an important distinction that I failed to take into account.
> I am now convinced.

src/TimeZones.jl
oxinabox: Ok thats fair

src/conversions.jl
```diff
@@ -5,11 +5,76 @@ using Mocking: Mocking, @mock
...
+Date(::ZonedDateTime, ::Union{Type{Local}, Type{UTC}})
+
+Dates.Date(zdt::ZonedDateTime, ::Type{Local}) = Date(DateTime(zdt, Local))
+Dates.Date(zdt::ZonedDateTime, ::Type{UTC}) = Date(DateTime(zdt, UTC))
+
+
+"""
+    Time(zdt::ZonedDateTime, ::Union{Type{Local}, Type{UTC}}) -> Time
+
+Create an implicit local/UTC `Time` from the given `ZonedDateTime`.
```
oxinabox: Couild we add a bit more text?

```suggestion
Create an implicit local/UTC `Time` from the given `ZonedDateTime`.

 - `Local` means a time local to the timezone in the `ZonedDateTime`, i.e. just discarding the timezone information. (**Not** local to the system time of the user invoking the code).
  - `UTC` means to shift to the `UTC` timezone and then discard all timezone information.
```

Maybe also this particular text should be put in to a string constant and interpolatred in at a few places?
Like in `Date` and `DateTime` as well?

Probably could be written better

omus: I rewrote the docstrings to be more sensible

omus: I rewrote the docstrings to be more sensible

omus:
Quick note: I'm still planning on doing the 1.0 version change as part of this but since this adds a deprecation I'll be doing a 0.11 release first so I can drop the deprecation as part of 1.0

src/discovery.jl
```diff
@@ -158,13 +158,13 @@ function next_transition_instant(zdt::ZonedDateTime)
...
     index = searchsortedfirst(
-        tz.transitions, TimeZones.utc(zdt),
+        tz.transitions, DateTime(zdt, UTC),
         by=el -> isa(el, TimeZones.Transition) ? el.utc_datetime : el,
     )
 
     index <= length(tz.transitions) || return nothing
 
-    # Use the UTC datetime of the transition and the offset information prior to the
+    # Use the UTC daftetime of the transition and the offset information prior to the
```
mattBrzezinski:
```suggestion
    # Use the UTC datetime of the transition and the offset information prior to the
```

src/TimeZones.jl
```diff
@@ -4,10 +4,12 @@ using Dates
...
 
 export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDateTime,
-    DateTime, TimeError, AmbiguousTimeError, NonExistentTimeError, UnhandledTimeError,
+    DateTime, Date, Time, UTC, Local, TimeError, AmbiguousTimeError, NonExistentTimeError,
```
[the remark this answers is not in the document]
omus: I can do some organization. What do you think of grouping by the file in which these are defined (and having a comment indicating the file)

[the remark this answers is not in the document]
mattBrzezinski: This might be too granular, but:

```
export @tz_str
export AmbiguousTimeError, NonExistentTimeError, TimeError
export Date, DateTime, Local, Time, TimeZone, UTC, VariableTimeZone, ZonedDateTime
export isdatetime
```

I think it's less about what the groupings are but more that there is some order to them.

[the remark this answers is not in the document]
omus: I'll note that there is an ordering to this but it's not as obvious as I'd like. A couple of key things I would like to communicate as part of the exports:
- Quite a few of these are re-exports from `Dates` which are just here in case someone only does `using TimeZones`
- There are some types that are closely related like `UTC` and `Local` which I think should be sequential

I'll post an revised version in a sec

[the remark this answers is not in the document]
omus: What I'm currently thinking:
```julia
export
    TimeZone, FixedTimeZone, VariableTimeZone, @tz_str, istimezone, localzone
    UTC, Local,
    ZonedDateTime, timezone,

    # Exceptions
    TimeError, AmbiguousTimeError, NonExistentTimeError, UnhandledTimeError,

    # Discovery tools
    timezone_names, all_timezones, timezones_from_abbr, timezone_abbrs,
    next_transition_instant, show_next_transition,

    # Extended functionality from Dates. Note: Some of these exports have already been done
    # but are listed here for completeness.
    TimeZone, UTC,
    DateTime, Date, Time,
    now, today,
    year, month, week, day, dayofmonth, hour, minute, second, millisecond
    yearmonthday, yearmonth, monthday,
    firstdayofyear, firstdayofquarter, firstdayofmonth, firstdayofweek,
    lastdayofyear, lastdayofquarter, lastdayofmonth, lastdayofweek
```
I'm unsure on what is better at this point and possibly this kind of logical grouping should just be in the documentation? In any case I think this could be made into an issue and not hold up this PR.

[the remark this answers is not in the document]
mattBrzezinski: Agreed, this isn't a dealbreaker for the PR going through and can be discussed afterwards.

Side-note; does exporting functionality multiple times have any consequences to it?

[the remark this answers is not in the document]
omus: > does exporting functionality multiple times have any consequences to it?

No it shouldn't

PR #234, opened 2019-12-20
Fix a stack overflow issue

omus:
I forgot what types this function took and tried `show_next_transition(Date(2019,6,1))` which resulted in a `Segmentation fault: 11` on Julia 1.0.5.

### DIFF
--- a/src/discovery.jl
+++ b/src/discovery.jl
@@ -253,5 +253,4 @@ function show_next_transition(io::IO, tz::TimeZone=localzone())
     show_next_transition(io, @mock now(tz))
 end
 
-show_next_transition(io::IO, x...) = throw(MethodError(show_next_transition, (io, x...)))
-show_next_transition(x...) = show_next_transition(stdout, x...)
+show_next_transition(x) = show_next_transition(stdout, x)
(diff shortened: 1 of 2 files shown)


omus:
I'm not going to bother tagging this right away as it's very low in importance

src/discovery.jl
```diff
@@ -253,4 +253,5 @@ function show_next_transition(io::IO, tz::TimeZone=localzone())
     show_next_transition(io, @mock now(tz))
 end
 
+show_next_transition(io::IO, x...) = throw(MethodError(show_next_transition, (io, x...)))
```
iamed2: Why haven't you just added a type on `x` below?

omus: Also worked. I was trying to be generic if I added additional methods of the form `show_next_transition(io::IO, x)`

omus: I guess I could also do that: `show_next_transition(x) = show_next_transition(stdout, x)`

omus:
Nightly failures due to: https://github.com/JuliaTime/TimeZones.jl/issues/235

PR #236, opened 2020-01-10
Address revert of period printing on 1.4

omus:
Another PR for this series (https://github.com/JuliaTime/TimeZones.jl/pull/217). Fixes https://github.com/JuliaTime/TimeZones.jl/issues/235

PR #237, opened 2020-01-10
Drop AppVeyor

omus:
Makes CI time to completion faster. See https://github.com/invenia/Mocking.jl/pull/76#issuecomment-572140910 for an analysis.


omus:
AppVeyor CI integration will fail until it is removed. That won't be done until this is merged.

omus:
The coverage difference confuses me. I feel like `fast_fail` may be the issue so I'll rebase this PR when https://github.com/JuliaTime/TimeZones.jl/pull/236 is merged.

Issue #210, opened 2019-08-22
TZ environment variable format


chrisoei:
```
julia> using TimeZones

julia> mytz = "America/Los_Angeles"
"America/Los_Angeles"

julia> istimezone(mytz)
true

julia> ENV["TZ"] = mytz
"America/Los_Angeles"

julia> localzone()
ERROR: ArgumentError: Unhandled TZ environment variable format: "America/Los_Angeles"
```
Setting TZ to ":America/Los_Angeles" works, but most Linux systems I've seen don't use the colon in front.

omus:
Thanks for the bug report. This can be easily added but I'd like to find a reference where this is documented. The [`tzset` man page](http://man7.org/linux/man-pages/man3/tzset.3.html) doesn't list that as a valid format.

chrisoei:
If you run tzselect on Ubuntu 18.04 and choose the United States Pacific time zone, it shows:

```You can make this change permanent for yourself by appending the line
	TZ='America/Los_Angeles'; export TZ
to the file '.profile' in your home directory; then log out and log in again.
```
I think this is the TZ environment variable format Ubuntu uses during the installation procedure.

Similarly, Ubuntu's bash accepts this format:
$ echo $TZ
America/Los_Angeles
$ date
Fri Aug 23 13:24:17 PDT 2019
$ export TZ="America/New_York"
$ date
Fri Aug 23 16:24:28 EDT 2019

It's possible this format is outdated but still widely in use.

Also, some tutorials like
https://www.cyberciti.biz/faq/linux-unix-set-tz-environment-variable/
https://www.thegeekstuff.com/2010/09/change-timezone-in-linux/
https://www.vertica.com/docs/9.2.x/HTML/Content/Authoring/InstallationGuide/BeforeYouInstall/TZenvironmentVar.htm
mention this format for the TZ environment variable.

omus:
I was hoping to find a reference to this format in a TZ specification document. I agree though that this is used commonly enough that it's worth adding.

PR #241, opened 2020-01-21
Fix TZ environmental variable support when not using colon

omus:
Fixes: https://github.com/JuliaTime/TimeZones.jl/issues/210

### DIFF
--- a/src/local.jl
+++ b/src/local.jl
@@ -32,15 +32,34 @@ function localzone()
 
         # Try getting the time zone from the "TZ" environment variable
         # http://linux.die.net/man/3/tzset
+        #
+        # Note: The macOS "man tzset 3" states some additional information about the colon
+        # being optional:
+        #
+        # > If TZ appears in the environment and its value begins with a colon (`:'), the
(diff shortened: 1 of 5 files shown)


src/local.jl
```diff
@@ -32,25 +32,41 @@ function localzone()
...
-            # When the TZ format starts with a colon this indicates that the time zone information
-            # should be read from a file.
+            # If the TZ format starts with a colon we'll prefer using the time zone
+            # information from the specified file.
             if startswith(name, ':')
-                name = name[2:end]
+                name = name[2:end]  # Name is either an relative or absolute path
             else
-                return parse_tz_format(name)
+                tz = tryparse_tz_format(name)
```
iamed2: This logic is reversed from the manual-specified logic, but I'm fine with that if you are

omus: The Linux tzset man page lists the filespec format as the the "second" format (http://man7.org/linux/man-pages/man3/tzset.3.html). There doesn't seem to be a canonical specification so I'll keep this as they are.

omus:
I needed to update the documentation more to have the CI pass. Everything should be good now.

PR #242, opened 2020-01-23
Cleanup

omus:
Some minor things I noticed I wanted to take care of before tagging 0.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,15 +55,21 @@ jobs:
       julia: nightly
     - <<: *doc_deploy
 
-env:
-  - JULIA_TZ_VERSION="2016j"  # Ideally only set for the "test" stage
 notifications:
   email: false
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  # Avoid setting JULIA_TZ_VERSION if the deps/build.jl file has been modified
-  - git fetch origin +:refs/remotes/origin/HEAD; if ! git diff --quiet origin/HEAD HEAD -- deps/build.jl; then unset JULIA_TZ_VERSION; fi
-  - echo $JULIA_TZ_VERSION  # Display version here as Julia 0.7+ doesn't show the build logs by default
+  # Only download the tzdata version used in TimeZones.jl's tests to avoid unnecessary
+  # load on IANA's servers. However, if the DEFAULT_TZDATA_VERSION constant was
+  # changed we need to also build the version specified by the constant to ensure it can be
+  # compiled.
+  # Note: The `git diff` command has a `--quiet` flag that will suppress output but doesn't
+  # catch all changes.
+  - git fetch origin +:refs/remotes/origin/HEAD
+  - |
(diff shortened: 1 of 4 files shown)


omus:
I've made Travis upset

omus:
As it turns out I used a single-line comment in an implicit YAML multiline string which was the culprit. Using a literal block fixed this.

omus:
Validating that updating `DEFAULT_TZ_VERSION` works: https://travis-ci.org/JuliaTime/TimeZones.jl/builds/641093990

Project.toml
```diff
@@ -12,8 +12,8 @@ Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 
 [compat]
-EzXML = "0.8, 0.9, 1"
-Mocking = "0.5.6, 0.6, 0.7"
+EzXML = "0.9.1, 1"
```
mattBrzezinski: Is there a reason for not using `0.9`?

.travis.yml
```diff
@@ -55,15 +55,34 @@ jobs:
...
+  # Only download the tzdata version used in TimeZones.jl's tests to avoid unnecessary
+  # load on IANA's servers. However, if the DEFAULT_TZDATA_VERSION constant was
+  # changed we need to also build the version specified by the constant to ensure it can be
+  # compiled.
+  # Note: The `git diff` command has a `--quiet` flag that will suppress output but doesn't
+  # catch all changes.
+  - git fetch origin +:refs/remotes/origin/HEAD
+  - |
+    if git diff --exit-code -G"^const DEFAULT_TZDATA_VERSION" origin/HEAD..HEAD > /dev/null; then
+        export JULIA_TZ_VERSION="2016j"  # Matches version used by tests
```
mattBrzezinski: I don't have a good solution to this, but it would be nice if we could unify this variable value and [this](https://github.com/JuliaTime/TimeZones.jl/pull/242/files#diff-fce720c43af3c52c862fd7451c7374b8R11) one somehow.

Project.toml
omus: `0.9.1` was the first revision with a Project.toml. I mentioned this in the comment message: https://github.com/JuliaTime/TimeZones.jl/pull/242/commits/eb0263b1005b03a488105231bb7bac6dc6c12211

.travis.yml
omus: I agree. The only idea that comes to mind is having a file that contains the version used. I could then reference the same file in the tests and here.

Project.toml
mattBrzezinski: Proper commit messages 😍 , I'll have to get back into the habit of reading them!

.travis.yml
mattBrzezinski: That's what I was thinking, but I wonder if there's anything better.

Either way, not a blocking PR issue and it can be figured out in the future. Just going to need to be watchful incase we change the `tz version`.

omus: Fair point

PR #243, opened 2020-01-23
Set project version to 1.0

omus:
I [previously thought](https://github.com/JuliaTime/TimeZones.jl/pull/231#issuecomment-561340976) it was time to bump the package version to 1.0 as I needed to do another minor release. This PR follows through on that promise.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,9 +1,4 @@
 using Base: @deprecate
 
 # BEGIN TimeZones 1.0 deprecations
-
-@deprecate localtime(zdt::ZonedDateTime) DateTime(zdt, Local) false
-@deprecate utc(zdt::ZonedDateTime) DateTime(zdt, UTC) false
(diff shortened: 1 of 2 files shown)


src/deprecated.jl
```diff
@@ -1,9 +1,4 @@
 using Base: @deprecate
 
 # BEGIN TimeZones 1.0 deprecations
```
nicoleepp: We probably want to get rid of these comments too

omus: I usually leave these in for when I add new deprecations. I'll note that these should have been labeled 0.11 which was an error on my part in https://github.com/JuliaTime/TimeZones.jl/pull/231

nicoleepp: Ah that makes more sense

oxinabox:
I was convinced last time that it was time for 1.0 thus I remain convinced

PR #251, opened 2020-03-12
Add Plot Recipe for ZonedDateTime

oxinabox:
Closes https://github.com/JuliaTime/TimeZones.jl/issues/184

It converts everything to UTC `DateTimes`, because that is the generally sensibly option.
and is the easiest.

More fancy we could convert everything in the vector to the majority timezone,
and then add the timezone to the axis title?
But am not that fancy.
Maybe a follow up PR

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -3,7 +3,7 @@ module TimeZones
 using Dates
 using Printf
 using Serialization
-using RecipesBase: @receipe
+using RecipesBase: RecipesBase, @recipe
 using Unicode
 
 import Dates: TimeZone, UTC
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -3,7 +3,7 @@ module TimeZones
 using Dates
 using Printf
 using Serialization
-using RecipesBase
+using RecipesBase: @receipe
 using Unicode
 
 import Dates: TimeZone, UTC
@@ -68,9 +68,7 @@ include("ranges.jl")
 include("discovery.jl")
 include("rounding.jl")
 include("parse.jl")
-
 include("plotting.jl")
-
 include("deprecated.jl")
 
 end # module
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -3,6 +3,7 @@ module TimeZones
 using Dates
 using Printf
 using Serialization
+using RecipesBase
 using Unicode
 
 import Dates: TimeZone, UTC
@@ -67,6 +68,9 @@ include("ranges.jl")
 include("discovery.jl")
 include("rounding.jl")
 include("parse.jl")
+
+include("plotting.jl")
+
 include("deprecated.jl")
 
 end # module
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -1,0 +1,5 @@
+#Note: recipes seem to need to use short-form functions
+@recipe f(::Type{T}, val::T) where {T<:ZonedDateTime} = (
+    zdt ->  Dates.value(DateTime(zdt, UTC)),
+    dt -> string(DateTime(Dates.UTM(dt))),
+)
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -1,4 +1,3 @@
-#==
 @recipe function f(xs::AbstractVector{<:ZonedDateTime}, ys)
     # if no zoned date times nothing to do but return empty list
     isempty(xs) && return xs
@@ -9,10 +8,3 @@
     xguide := strip(get(plotattributes, :xguide, "") * " (timezone: $tz)")
     new_xs, ys
 end
-
-==#
-
-@recipe f(::Type{T}, val::T) where {T<:ZonedDateTime} = (
-    zdt ->  Dates.value(DateTime(zdt, UTC)),
-    dt -> string(DateTime(Dates.UTM(dt))) * "+00:00",  # make clear we are in UTC
-)
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -1,4 +1,4 @@
-#Note: recipes seem to need to use short-form functions
+# Note: recipes seem to need to use short-form functions
 @recipe f(::Type{T}, val::T) where {T<:ZonedDateTime} = (
     zdt ->  Dates.value(DateTime(zdt, UTC)),
     dt -> string(DateTime(Dates.UTM(dt))),
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -1,5 +1,5 @@
 # Note: recipes seem to need to use short-form functions
 @recipe f(::Type{T}, val::T) where {T<:ZonedDateTime} = (
-    zdt ->  Dates.value(DateTime(zdt, UTC)),
+    zdt -> Dates.value(DateTime(zdt, UTC)),
     dt -> string(DateTime(Dates.UTM(dt))),
 )
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -1,5 +1,7 @@
-# Note: recipes seem to need to use short-form functions
-@recipe f(::Type{T}, val::T) where {T<:ZonedDateTime} = (
-    zdt -> Dates.value(DateTime(zdt, UTC)),
-    dt -> string(DateTime(Dates.UTM(dt))),
-)
+# Note: type-recipes seem to need to use short-form functions
+@recipe function f(::Type{<:ZonedDateTime}, val::ZonedDateTime)
+    return (
+        zdt -> Dates.value(DateTime(zdt, UTC)),
+        dt -> string(DateTime(Dates.UTM(dt))),
+    )
+end
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -1,3 +1,12 @@
+#==
+Plot ZonedDateTime, on x-axis.
+We convert it to DateTimes, in the local timezone,
+and we list that timezone in the x-axis label.
+
+This is just one of many options for how to display this info.
+See this PR: https://github.com/JuliaTime/TimeZones.jl/pull/251#issuecomment-603806198
+for details on the options and their tradeoffs.
+==#
 @recipe function f(xs::AbstractVector{<:ZonedDateTime}, ys)
     # if no zoned date times nothing to do but return empty list
     isempty(xs) && return xs
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -1,3 +1,4 @@
+#==
 @recipe function f(xs::AbstractVector{<:ZonedDateTime}, ys)
     # if no zoned date times nothing to do but return empty list
     isempty(xs) && return xs
@@ -8,3 +9,10 @@
     xguide := strip(get(plotattributes, :xguide, "") * " (timezone: $tz)")
     new_xs, ys
 end
+
+==#
+
+@recipe f(::Type{T}, val::T) where {T<:ZonedDateTime} = (
+    zdt ->  Dates.value(DateTime(zdt, UTC)),
+    dt -> string(DateTime(Dates.UTM(dt))) * "+00:00",  # make clear we are in UTC
+)
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -5,6 +5,10 @@
     new_xs = DateTime.(astimezone.(xs, tz), Local)
 
     # xguide is the proper name for xlabel
-    xguide := strip(get(plotattributes, :xguide, "") * " (timezone: $tz)")
+    label = get(plotattributes, :xguide, "")
+    if !isempty(label)
+        label *= " "  # leave space between original label and the timezone info
+    end
+    xguide := label *= "(timezone: $tz)"
     new_xs, ys
 end
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -15,9 +15,7 @@ for details on the options and their tradeoffs.
 
     # xguide is the proper name for xlabel
     label = get(plotattributes, :xguide, "")
-    if !isempty(label)
-        label *= " "  # leave space between original label and the timezone info
-    end
-    xguide := label *= "(timezone: $tz)"
+    label *= isempty(label) ? "Time zone: $tz" : " ($tz)" 
+    xguide := label
     new_xs, ys
 end
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -1,7 +1,10 @@
-# Note: type-recipes seem to need to use short-form functions
-@recipe function f(::Type{<:ZonedDateTime}, val::ZonedDateTime)
-    return (
-        zdt -> Dates.value(DateTime(zdt, UTC)),
-        dt -> string(DateTime(Dates.UTM(dt))),
-    )
+@recipe function f(xs::AbstractVector{<:ZonedDateTime}, ys)
+    # if no zoned date times nothing to do but return empty list
+    isempty(xs) && return xs
+    tz = timezone(first(xs))  # convert all to same timezone as first one
+    new_xs = DateTime.(astimezone.(xs, tz), Local)
+
+    # xguide is the proper name for xlabel
+    xguide := strip(get(plotattributes, :xguide, "") * " (timezone: $tz)")
+    new_xs, ys
 end
--- a/test/plotting.jl
+++ b/test/plotting.jl
@@ -10,30 +10,30 @@
     # what the point should be after recipe is applied
     expected_dates = DateTime.(zoned_dates, Local)
 
-    @testset "No label" begin
+    @testset "No label (should now say the Timezone)" begin
         # The below use of `apply_recipe` is equivelent to:
         # plot(zoned_dates, 1:11)
         result, = RecipesBase.apply_recipe(Dict{Symbol, Any}(), zoned_dates, 1:11)
         xs, ys = result.args
         @test xs == expected_dates
         @test ys == 1:11
-        @test result.plotattributes[:xguide] == "(timezone: EST)"
+        @test result.plotattributes[:xguide] == "Time zone: EST"
     end
 
     @testset "label (should append to it)" begin
         # The below use of `apply_recipe` is equivelent to:
         # plot(zoned_dates, 1:11; xguide="X-Axis")
-        result, = RecipesBase.apply_recipe(Dict{Symbol, Any}(:xguide=>"Hi"), zoned_dates, 1:11)
+        result, = RecipesBase.apply_recipe(Dict{Symbol, Any}(:xguide=>"X-Axis"), zoned_dates, 1:11)
         xs, ys = result.args
         @test xs == expected_dates
         @test ys == 1:11
-        @test result.plotattributes[:xguide] == "Hi (timezone: EST)"
+        @test result.plotattributes[:xguide] == "X-Axis (EST)"
     end
 
     @testset "No items" begin
         empty_xs = ZonedDateTime[]
         empty_ys = 0:-1
-        result = RecipesBase.apply_recipe(Dict{Symbol, Any}(:xguide=>"Hi"), empty_xs, empty_ys)
+        result = RecipesBase.apply_recipe(Dict{Symbol, Any}(:xguide=>"X-Axis"), empty_xs, empty_ys)
         @test true  # we are just making sure it it didn't error
     end
 end
--- a/test/plotting.jl
+++ b/test/plotting.jl
@@ -1,33 +1,39 @@
 @testset "ZonedDateTime plot recipe" begin
+    # Note: in these tests we use `RecipesBase.apply_recipe` rather than `plot` as it
+    # lets use avoid a Plots.jl dependency and issues with running that during tests.
+    # `RecipesBase.apply_recipe` is not a documented API, but is fairly usable.
+    # Comments above each use show the matching `plot` function command
     start_zdt = ZonedDateTime(2017,1,1,0,0,0, tz"EST")
     end_zdt = ZonedDateTime(2017,1,1,10,30,0, tz"EST")
     zoned_dates = start_zdt:Hour(1):end_zdt
 
-    function check_extrema(x_axis)
-        date_val(zdt) = Dates.value(DateTime(zdt, Local))
-
-        @test x_axis[:extrema].emin ≈ date_val(start_zdt) rtol=0.000_000_1
-        @test x_axis[:extrema].emax ≈ date_val(end_zdt) rtol=0.000_000_1
-    end
+    # what the point should be after recipe is applied
+    expected_dates = DateTime.(zoned_dates, Local)
 
     @testset "No label" begin
-        result = scatter(zoned_dates, 1:11)
-        x_axis = result.subplots[1][:xaxis]
-        check_extrema(x_axis)
-        @test x_axis[:guide] == "(timezone: EST)"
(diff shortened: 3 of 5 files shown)


src/plotting.jl
```diff
@@ -0,0 +1,5 @@
+#Note: recipes seem to need to use short-form functions
```
nicoleepp:
```suggestion
# Note: recipes seem to need to use short-form functions
```

src/plotting.jl
```diff
@@ -0,0 +1,5 @@
...
+@recipe f(::Type{T}, val::T) where {T<:ZonedDateTime} = (
+    zdt ->  Dates.value(DateTime(zdt, UTC)),
```
nicoleepp:
```suggestion
    zdt -> Dates.value(DateTime(zdt, UTC)),
```

oxinabox:
Here is what the plot in the test looks like:
image

src/plotting.jl
```diff
@@ -0,0 +1,5 @@
+# Note: recipes seem to need to use short-form functions
```
asinghvi17: They don't need short form functions, they work with any function definition syntax.

test/runtests.jl
```diff
@@ -1,12 +1,15 @@
...
 
+using Plots
 using Test
 using TimeZones
 using TimeZones: PKG_DIR
 using TimeZones.TZData: ARCHIVE_DIR, TZSource, compile, build
 using Unicode
 
 Mocking.activate()
+unicodeplots()  # so don't need X11 to test plotting
```
asinghvi17: you should be able to set `ENV["GKS_WORKSTATION"] = 100` or something to avoid needing X11, not sure what the exact name was (you can probably find out from the GR docs).  Consider using something like VisualRegressionTests.jl for visual regression if you need it.

src/plotting.jl
oxinabox: Doesn't seem to  AFAICT.

If I change it to:
```
@recipe function f(::Type{T}, val::T) where {T<:ZonedDateTime}
    return (
        zdt -> Dates.value(DateTime(zdt, UTC)),
        dt -> string(DateTime(Dates.UTM(dt))),
    )
end
```

my tests fail with:

```
ZonedDateTime plot recipe: Error During Test at /Users/oxinabox/JuliaEnvs/TimeZones.jl/test/plotting.jl:1
  Got exception outside of a @test
  type #31#33 has no field args
  Stacktrace:
   [1] getproperty(::Function, ::Symbol) at ./Base.jl:20
   [2] _apply_type_recipe(::Dict{Symbol,Any}, ::StepRange{ZonedDateTime,Hour}) at /Users/oxinabox/.julia/packages/Plots/rNwM4/src/series.jl:183
   [3] macro expansion at /Users/oxinabox/.julia/packages/Plots/rNwM4/src/series.jl:228 [inlined]
   [4] apply_recipe(::Dict{Symbol,Any}, ::StepRange{ZonedDateTime,Hour}, ::UnitRange{Int64}) at /Users/oxinabox/.julia/packages/RecipesBase/G4s6f/src/RecipesBase.jl:279
   [5] _process_userrecipes(::Plots.Plot{Plots.UnicodePlotsBackend}, ::Dict{Symbol,Any}, ::Tuple{StepRange{ZonedDateTime,Hour},UnitRange{Int64}}) at /Users/oxinabox/.julia/packages/Plots/rNwM4/src/pipeline.jl:85
   [6] _plot!(::Plots.Plot{Plots.UnicodePlotsBackend}, ::Dict{Symbol,Any}, ::Tuple{StepRange{ZonedDateTime,Hour},UnitRange{Int64}}) at /Users/oxinabox/.julia/packages/Plots/rNwM4/src/plot.jl:178
   [7] #plot#138(::Base.Iterators.Pairs{Symbol,Symbol,Tuple{Symbol},NamedTuple{(:seriestype,),Tuple{Symbol}}}, ::typeof(plot), ::StepRange{ZonedDateTime,Hour}, ::Vararg{Any,N} where N) at /Users/oxinabox/.julia/packages/Plots/rNwM4/src/plot.jl:57
   [8] (::RecipesBase.var"#kw##plot")(::NamedTuple{(:seriestype,),Tuple{Symbol}}, ::typeof(plot), ::StepRange{ZonedDateTime,Hour}, ::UnitRange{Int64}) at ./none:0
   [9] #scatter#402(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(scatter), ::StepRange{ZonedDateTime,Hour}, ::Vararg{Any,N} where N) at /Users/oxinabox/.julia/packages/RecipesBase/G4s6f/src/RecipesBase.jl:393
   [10] scatter(::StepRange{ZonedDateTime,Hour}, ::Vararg{Any,N} where N) at /Users/oxinabox/.julia/packages/RecipesBase/G4s6f/src/RecipesBase.jl:393
   [11] top-level scope at /Users/oxinabox/JuliaEnvs/TimeZones.jl/test/plotting.jl:6
   [12] top-level scope at /usr/local/src/julia/julia-1.3/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
   [13] top-level scope at /Users/oxinabox/JuliaEnvs/TimeZones.jl/test/plotting.jl:2
   [14] include at ./boot.jl:328 [inlined]
   [15] include_relative(::Module, ::String) at ./loading.jl:1105
   [16] include(::Module, ::String) at ./Base.jl:31
   [17] include(::String) at ./client.jl:424
   [18] top-level scope at /Users/oxinabox/JuliaEnvs/TimeZones.jl/test/runtests.jl:73
```

asinghvi17: what if you do:
```julia
@recipe function f(::Type{<: ZonedDateTime}, val::ZonedDateTime)
```
instead?

oxinabox: same error.

src/plotting.jl
```diff
@@ -0,0 +1,7 @@
+# Note: type-recipes seem to need to use short-form functions
```
omus:
```suggestion
```

src/TimeZones.jl
```diff
@@ -3,6 +3,7 @@ module TimeZones
 using Dates
 using Printf
 using Serialization
+using RecipesBase
```
omus: I want to move towards only importing what is being used:
```suggestion
using RecipesBase: @receipe
```

src/TimeZones.jl
```diff
@@ -67,6 +68,9 @@ include("ranges.jl")
 include("discovery.jl")
 include("rounding.jl")
 include("parse.jl")
+
+include("plotting.jl")
+
```
omus:
```suggestion
```

src/TimeZones.jl
```diff
@@ -67,6 +68,9 @@ include("ranges.jl")
 include("discovery.jl")
 include("rounding.jl")
 include("parse.jl")
+
```
omus:
```suggestion
```

src/plotting.jl
```diff
@@ -0,0 +1,7 @@
...
+@recipe function f(::Type{<:ZonedDateTime}, val::ZonedDateTime)
+    return (
+        zdt -> Dates.value(DateTime(zdt, UTC)),
+        dt -> string(DateTime(Dates.UTM(dt))),
+    )
+end
```
omus: Can you go into how this works? I would like to have the time zone appear in some form. Probably just the numeric offset would be a good place to start.

test/plotting.jl
```diff
@@ -0,0 +1,12 @@
+@testset "ZonedDateTime plot recipe" begin
+    start_zdt = ZonedDateTime(2017,1,1,0,0,0, tz"EST")
+    end_zdt = ZonedDateTime(2017,1,1,10,30,0, tz"EST")
+    zoned_dates = start_zdt:Hour(1):end_zdt
+
+    result = scatter(zoned_dates, 1:11)
```
omus: Currently failing all CI tests:
```
ZonedDateTime plot recipe: Error During Test at /home/travis/build/JuliaTime/TimeZones.jl/test/plotting.jl:1
  Got exception outside of a @test
  type #51#53 has no field args
  Stacktrace:
   [1] getproperty(::Function, ::Symbol) at ./Base.jl:20
   [2] _apply_type_recipe(::Dict{Symbol,Any}, ::StepRange{ZonedDateTime,Hour}) at /home/travis/.julia/packages/Plots/rNwM4/src/series.jl:183
   [3] macro expansion at /home/travis/.julia/packages/Plots/rNwM4/src/series.jl:228 [inlined]
   [4] apply_recipe(::Dict{Symbol,Any}, ::StepRange{ZonedDateTime,Hour}, ::UnitRange{Int64}) at /home/travis/.julia/packages/RecipesBase/G4s6f/src/RecipesBase.jl:279
   [5] _process_userrecipes(::Plots.Plot{Plots.UnicodePlotsBackend}, ::Dict{Symbol,Any}, ::Tuple{StepRange{ZonedDateTime,Hour},UnitRange{Int64}}) at /home/travis/.julia/packages/Plots/rNwM4/src/pipeline.jl:85
   [6] _plot!(::Plots.Plot{Plots.UnicodePlotsBackend}, ::Dict{Symbol,Any}, ::Tuple{StepRange{ZonedDateTime,Hour},UnitRange{Int64}}) at /home/travis/.julia/packages/Plots/rNwM4/src/plot.jl:178
   [7] #plot#138(::Base.Iterators.Pairs{Symbol,Symbol,Tuple{Symbol},NamedTuple{(:seriestype,),Tuple{Symbol}}}, ::typeof(plot), ::StepRange{ZonedDateTime,Hour}, ::Vararg{Any,N} where N) at /home/travis/.julia/packages/Plots/rNwM4/src/plot.jl:57
   [8] (::RecipesBase.var"#kw##plot")(::NamedTuple{(:seriestype,),Tuple{Symbol}}, ::typeof(plot), ::StepRange{ZonedDateTime,Hour}, ::UnitRange{Int64}) at ./none:0
   [9] #scatter#402(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(scatter), ::StepRange{ZonedDateTime,Hour}, ::Vararg{Any,N} where N) at /home/travis/.julia/packages/RecipesBase/G4s6f/src/RecipesBase.jl:393
   [10] scatter(::StepRange{ZonedDateTime,Hour}, ::Vararg{Any,N} where N) at /home/travis/.julia/packages/RecipesBase/G4s6f/src/RecipesBase.jl:393
   [11] top-level scope at /home/travis/build/JuliaTime/TimeZones.jl/test/plotting.jl:6
   [12] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
   [13] top-level scope at /home/travis/build/JuliaTime/TimeZones.jl/test/plotting.jl:2
   ...
```

test/runtests.jl
```diff
@@ -67,6 +70,8 @@ include("helpers.jl")
     include("rounding.jl")
     include("parse.jl")
```
omus:
```suggestion
```

test/plotting.jl
omus: See https://github.com/JuliaTime/TimeZones.jl/pull/251#discussion_r391649594

src/plotting.jl
oxinabox: It remains true.

src/plotting.jl
oxinabox: I have no idea how this one works internally.
but the effect is to convert it ot UTC.
And in that case I think showing the timezone is not required. UTC the sensible default.
(One could even argue that all `DateTimes` are in UTC, since with have `datetime2unixepoch` and `unixepoch` is defined as being in UTC, right?)


If you want to show the timezone,
then I think we should not convert to UTC, but should convert to the timezone of the first element, then use `Local` time, and display the timezone in the axis label.
But that would be a totally differnent PR.

omus: I think end users will just assume either UTC or Local depending on their personal preference. As I said I'd like to show the time zone offset to avoid any confusion that could occur.

> (One could even argue that all DateTimes are in UTC, since with have datetime2unixepoch and unixepoch is defined as being in UTC, right?)

`DateTime`s have an implicit time zone but without context you don't know what that is. `2020-03-16T12:00` to me means noon in *any* time zone until additional context is provided.

oxinabox: Shall I close this PR then and make another one?

omus: Maybe make the other PR first. I'm unclear on why this PR shouldn't be just updated

oxinabox: It will use a entirely different mechanism. 
A user-recipe rather than a type-recipe, I think is the terminolgy.
Not even sure the tests will be the same.
But I can do it in this one.

oxinabox:
Here is new demo.
We always use the timezone of first element.
And we show it in the xlabel (after what ever the label itself is)

image

oxinabox:
Travis CI windows is failing to build Plots because FFMPEG

giordano:
> Travis CI windows is failing to build Plots because FFMPEG

More precisely, it's Travis fault that has an incomplete Windows installation, see https://github.com/JuliaIO/FFMPEG.jl/issues/14 and references therein

omus:
@oxinabox would you mind doing a demo over a fall DST transition? I suspect the duplicate hour could be confusing as they would both appear the same on the axis.

oxinabox:
```
start_zdt = ZonedDateTime(2017,10, 28, 12, 1, 1, tz"Europe/London")
end_zdt = ZonedDateTime(2017,10, 29, 12, 1, 1, tz"Europe/London")
zoned_dates = start_zdt:Hour(1):end_zdt

scatter(zoned_dates, 1:length(zoned_dates); marker=:xcross, legend=false)
scatter!(DateTime.(zoned_dates, UTC), 1:length(zoned_dates); marker=:cross)
```
image

omus:
> If we didn't got UTC what is better to show in the xlabel?

I was thinking that including the offset as part of the ticks itself may be a better option than changing the xlabel.

oxinabox:
>   I was thinking that including the offset as part of the ticks itself may be a better option than changing the xlabel.

I don't think i can do that, not easily.

Only way I know to change  the ticks themselves is a type-recipe.
Like this was using before.
But then that means I get to provide 2 functions:

`::ZonedDateTime -> ::Number` that is used to workout where to put it.
and `::Number -> String` to workout what to put on the ticks.

I guess I could just always convert to UTC and always append UTC to the end of the string.

Not as cool as showing:
```
2017-10-29T00:01:01+01:00
2017-10-29T01:01:01+01:00
2017-10-29T01:01:01+00:00
2017-10-29T02:01:01+00:00
```

But usable

daschw:
I already mentioned this on slack but I'll write it here again. https://github.com/JuliaPlots/Plots.jl/pull/2483 will allow you to define a vector type recipe like
```julia
@recipe function f(::Type{T}, zdt::T) where T <: AbstractArray{<:DateTime}
```
and have guides and other axis attributes automatically mapped to the correct axis.
The advantage compared to the element-wise type recipe is that you would not have to touch the formatter. With the default `DateTime` formatter you would get nicer tick calculations, i.e. *round* ticks starting at the beginning of days/months/years. (In Plots a special `optimize_datetime_ticks` is called if the axis formatter is `datetime_formatter`, which is defined in Plots. It would be nice if we had a more general approach here)
The disadvantage is that you should not touch the formatter (to not loose the round tick calculation) and, hence you'd have to put the timezone information in the guides.

daschw:
I think you could still implement both, vector and element-wise , type recipes. The vector type recipe takes precedence in the new Plots releases and users with older versions of Plots would fallback to the element-wise type recipes, because before JuliaPlots/Plots.jl#2483 vector type recipes with the signature above would never get called.

giordano:
Is `Plots` really necessary for testing?  In `Measurements.jl` [I use `RecipesBase.apply_recipe`](https://github.com/JuliaPhysics/Measurements.jl/blob/bcaf37516d33c4b2f7b5d8dbdbc362e69f692c12/test/plots.jl) for that, without adding extra dependencies.

oxinabox:
Ok this is the latest commit: we are back to a type-recipe:
image

oxinabox:
So we have various options for 

## how to indicate timezone:

### Option t1: Don't
**Pro:** avoids visual noise
**Con:** people will not know the timezone

### Option t2: Include in Axis Label
**Pro:** Doesn't add much visual noise
**Con1:** It can make axis label long, kinda ugly

### Option t3: Include in tick labels
**Pro:** Kinda where the info belongs.
**Con1:** Makes ticklabels very long, may overlap, dense on information hard to read.
**Con2:** Breaks Plots.jl's setup that does tick alignment on Hours/Days etc


----------------

## what timezone to use:
Ignoring the case that a vector has mix of timezones.

### Option Z1: UTC
**Pro:** the universal default

### Option Z2: Local
**Pro:** the natural zone to use, so it obvious various timecentric events
**Con:** weird during DST transitions.

omus:
> ### Option t1: Don't

I would definitely not go with this option. If people want this they can manually convert to `DateTime`

> ### Option t2: Include in Axis Label

If we're using the axis label to show time zone information does this mean that the label is unavailable to the user or if they set it the time zone information is lost?

> ### Option t3: Include in tick labels

My preference. I feel like the amount of labelled ticks should just be reduced to reduce the visual noise. I'm not sure if recipes are supposed to control that or it would be automatic

> What timezone to use

It seems clear to me to use the time zone which the `ZonedDateTime` is using but only display the offset. If the user wants to see UTC they can simply convert the `ZonedDateTime` to UTC first. We could even introduce a special case where we show `Z` instead of `+00:00`

oxinabox:
>>  Option t2: Include in Axis Label
> If we're using the axis label to show time zone information does this mean that the label is unavailable to the user or if they set it the time zone information is lost?

No, it is appended to the user provided label.
See this test case from that commit
https://github.com/JuliaTime/TimeZones.jl/pull/251/commits/d706959fce1f738b0389fad32eb600ee7bbafe61#diff-bda42efb2a97172199a8600136b40696R20-R24

oxinabox:
For now @omus and I have decided to go forward with option: t2 + z2.
Which corresponds to what was showin in https://github.com/JuliaTime/TimeZones.jl/pull/251#issuecomment-600632254

daschw:
> No, it is appended to the user provided label.

In that case the user could not choose to have no label, though. The alternative would be to provide the timezone as default label and leave the option to overwrite it to the users.

oxinabox:
> In that case the user could not choose to have no label, though.

Correct, we are disallowing the user from throwing away information.
If they want to do that then they can convert to `DateTime` in the timezone of their choice.

In a follow up PR, we may add a kwarg,
so the user can chose:
 - `timezone_info = :none` for Option t1
 - `timezone_info = :inaxislabel` for Option t2
 - `timezone_info = :inticks` for Option t3

oxinabox:
@omus  can you look at the last PR?
It changes the tests to be based on @giordano 's method (as shown [here](https://github.com/JuliaPhysics/Measurements.jl/blob/bcaf37516d33c4b2f7b5d8dbdbc362e69f692c12/test/plots.jl)) of using `apply_recipe` directly, rather than actually calling Plots.

Big advantage is that this means it works with TravisCI on windows.
And we don't have a test time dependency on Plots at all so no need to worry about that breaking etc and having to update tests

omus:
> It changes the tests to be based on @giordano 's method (as shown here) of using `apply_recipe` directly, rather than actually calling Plots.

I'm ok with it if you add a comment in the test indicating that we're using an undocumented part of the API. Let's also leave that as a separate commit so it's clear what needs to be changed if we find it to be an ineffective way to test the plotting support.

src/plotting.jl
```diff
@@ -0,0 +1,23 @@
...
+    isempty(xs) && return xs
+    tz = timezone(first(xs))  # convert all to same timezone as first one
+    new_xs = DateTime.(astimezone.(xs, tz), Local)
+
+    # xguide is the proper name for xlabel
+    label = get(plotattributes, :xguide, "")
+    if !isempty(label)
+        label *= " "  # leave space between original label and the timezone info
+    end
+    xguide := label *= "(timezone: $tz)"
```
omus: From our talk I was thinking we should do something like:
```julia
    if isempty(label)
        label = "Time zone: $tz"
    else
        label *= " ($tz)"
    end
    xguide := label

File: test/plotting.jl [hunk 13]
```diff
@@ -0,0 +1,30 @@
...
+        @test xs == expected_dates
+        @test ys == 1:11
+        @test result.plotattributes[:xguide] == "Hi (timezone: EST)"
+    end
+
+    @testset "No items" begin
+        empty_xs = ZonedDateTime[]
+        empty_ys = 0:-1
+        result = RecipesBase.apply_recipe(Dict{Symbol, Any}(:xguide=>"Hi"), empty_xs, empty_ys)
+        @test true  # we are just making sure it it didn't error
```
omus: We can test that the `:xguide` is unmodified

test/plotting.jl
```diff
@@ -0,0 +1,30 @@
...
+    @testset "No label" begin
+        result, = RecipesBase.apply_recipe(Dict{Symbol, Any}(), zoned_dates, 1:11)
+        xs, ys = result.args
+        @test xs == expected_dates
+        @test ys == 1:11
+        @test result.plotattributes[:xguide] == "(timezone: EST)"
+    end
+
+    @testset "label (should append to it)" begin
+        result, = RecipesBase.apply_recipe(Dict{Symbol, Any}(:xguide=>"Hi"), zoned_dates, 1:11)
```
omus: How about being a little more informative than `"Hi"`?
```suggestion
        result, = RecipesBase.apply_recipe(Dict{Symbol, Any}(:xguide=>"X-Axis"), zoned_dates, 1:11)
```

omus:
Otherwise this is looking good. Looks like we may not need to do: https://github.com/JuliaTime/TimeZones.jl/pull/252

src/plotting.jl
oxinabox: Ok, sounds good to me

test/plotting.jl
oxinabox: We can't actually, because this does not actually return the usual thing.
Because the recipe bails out early because this is an invalid senario.

oxinabox:
I just checked this against the Intervals PR: all is working correctly:

```

using Plots
using Intervals
using TimeZones
using Dates

start_zdt = ZonedDateTime(2017,10, 28, 12, 1, 1, tz"Europe/London")
end_zdt = ZonedDateTime(2017,10, 29, 12, 1, 1, tz"Europe/London")
zoned_dates = start_zdt:Hour(1):end_zdt

date_intervals = [dt .. (dt + Hour(1)) for dt in zoned_dates]
plot(HE.(zoned_dates), 1:26)
```

image

You can see the very short interval where the duration is shown like it is 0 because of DST.
Which looks funny, but in practice out plots that use those types together tend to look more like:

```
using Plots
using Intervals
using TimeZones
using Dates

start_zdt = ZonedDateTime(2017, 6, 28, 12, 1, 1, tz"Europe/London")
end_zdt = ZonedDateTime(2017,10, 29, 12, 1, 1, tz"Europe/London")
zoned_dates = start_zdt:Hour(1):end_zdt

date_intervals = [dt .. (dt + Hour(1)) for dt in zoned_dates]
plot(HE.(zoned_dates), (0.5*(1:100) .+ sin.(1:100) .+ 0.2rand(100)) )
```

image

PR #254, opened 2020-04-16
make work with RecipeBase 1.0

oxinabox:
I tested it still works with 0.8 as well

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.1.0"
+version = "1.2.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
@@ -15,7 +15,7 @@ Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 [compat]
 EzXML = "0.9.1, 1"
 Mocking = "0.7"
-RecipesBase = "0.8"
+RecipesBase = "0.8, 1.0"
 julia = "1"
 
 [extras]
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.1.0"
+version = "1.2.0"
```
omus, line 4: Isn't this a bug fix? If not it should probably have some tests

nickrobinson251: The decision we made on  how to bump package version when changing dependency compatability was:
-    Patch: if it is just adding compat for both new and old versions of the dependency and no code changes required.
-    Minor: if it was dropping old version of dependency or *if required code changes* (normally both go together)
-    Major: almost never. Only if leaking abstractions e.g. subtyping somethat that has its behavour changed

Since this PR changes code, i believe minor is right.

oxinabox: According to  statement on 31 March  Dev meeting several weeks ago:

- Patch if it is just adding compat for both new and old and no code changes required.
- Minor: if it was dropping old version or if required code changes (normally both go together)
- Major: almost never. Only if leaking abstractions e.g. subtyping somethat that has its behavour changed

---

Code changes were required to make compatible with 1.0
(but these changes in this case work with both 0.8 and 1.0)
Thus minor.

> If not it should probably have some tests

It turns a test that failed, in RecipieBase 1.0, into one that succeeds.
The empty plot test.

omus: Minor sounds fine. I'll note the return when `xs` is empty did change so we should update the test so this is actually tested for. Currently the test only ensures an error doesn't occur

oxinabox: Done

omus:
Looks like we forgot to publish 1.1.0. I'll include this change in that release: https://github.com/JuliaTime/TimeZones.jl/pull/255

PR #256, opened 2020-04-20
Remove use of `systemsetup` for macOS localzone

Nosferican:
Rather than using `sudo systemsetup -gettimezone` one can get the localzone on OSX based on `/etc/localtime` (i.e., no sudo required). Should fix issue with 
```
ERROR: IOError: could not spawn `systemsetup -gettimezone`: no such file or directory (ENOENT)
```
on various applications such as VS Code even with full disk privileges.

### DIFF
--- a/src/local.jl
+++ b/src/local.jl
@@ -18,15 +18,7 @@ function localzone()
     mask = Class(:STANDARD) | Class(:LEGACY)
 
     @static if Sys.isapple()
-        name = @mock read(`systemsetup -gettimezone`, String)  # Appears to only work as root
-        if startswith(name, "Time Zone: ")
-            name = strip(replace(name, "Time Zone: " => ""))
-        else
-            # link will be something like /usr/share/zoneinfo/Europe/Warsaw
-            name = @mock readlink("/etc/localtime")
-            name = match(r"(?<=zoneinfo/).*$", name).match
-        end
-        return TimeZone(name, mask)
+        return TimeZone(readlink("/etc/localtime")[27:end], mask)
     elseif Sys.isunix()
         name = ""
 
--- a/src/local.jl
+++ b/src/local.jl
@@ -18,7 +18,10 @@ function localzone()
     mask = Class(:STANDARD) | Class(:LEGACY)
 
     @static if Sys.isapple()
-        return TimeZone(readlink("/etc/localtime")[27:end], mask)
+        # Link will point to something like "/usr/share/zoneinfo/Europe/Warsaw"
+        link = @mock readlink("/etc/localtime")
+        name = match(r"(?<=zoneinfo/).*$", link).match
+        return TimeZone(name, mask)
     elseif Sys.isunix()
         name = ""
 
--- a/test/local_mocking.jl
+++ b/test/local_mocking.jl
@@ -11,13 +11,9 @@ win_name = name == "Europe/Warsaw" ? "Central European Standard Time" : "Samoa S
 tz = TimeZone(name)
 
 if Sys.isapple()
-
     # Determine time zone from /etc/localtime.
-    patches = [
-        @patch read(cmd::AbstractCmd, ::Type{String}) = ""
-        @patch readlink(filename::AbstractString) = "/usr/share/zoneinfo/$name"
-    ]
-    apply(patches) do
+    patch = @patch readlink(filename::AbstractString) = "/usr/share/zoneinfo/$name"
+    apply(patch) do
         @test localzone() == tz
     end
 
--- a/test/local_mocking.jl
+++ b/test/local_mocking.jl
@@ -12,7 +12,7 @@ tz = TimeZone(name)
 
 if Sys.isapple()
     # Determine time zone from /etc/localtime.
-    patch = @patch readlink(filename::AbstractString) = "/usr/share/zoneinfo/$name"
+    patch = @patch readlink(::AbstractString) = "/usr/share/zoneinfo/$name"
     apply(patch) do
         @test localzone() == tz
     end
@@ -89,7 +89,7 @@ end
 function with_localzone(func::Function, name::AbstractString)
     @static if Sys.isapple()
         patches = [
-            @patch read(cmd::AbstractCmd, ::Type{String}) = "Time Zone:  $name\n"
+            @patch readlink(::AbstractString) = "/usr/share/zoneinfo/$name"
         ]
     elseif Sys.iswindows()
         patches = [
--- a/test/local_mocking.jl
+++ b/test/local_mocking.jl
@@ -11,11 +11,6 @@ win_name = name == "Europe/Warsaw" ? "Central European Standard Time" : "Samoa S
 tz = TimeZone(name)
 
 if Sys.isapple()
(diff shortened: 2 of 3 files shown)


Nosferican:
@omus, Could you adapt the test code to allow for the transition?

src/local.jl
```diff
@@ -18,15 +18,7 @@ function localzone()
...
-        name = @mock read(`systemsetup -gettimezone`, String)  # Appears to only work as root
-        if startswith(name, "Time Zone: ")
-            name = strip(replace(name, "Time Zone: " => ""))
-        else
-            # link will be something like /usr/share/zoneinfo/Europe/Warsaw
-            name = @mock readlink("/etc/localtime")
-            name = match(r"(?<=zoneinfo/).*$", name).match
-        end
-        return TimeZone(name, mask)
+        return TimeZone(readlink("/etc/localtime")[27:end], mask)
```
omus, line 21: If you take the contents of the `else` the test should work again. I'll note I am not a fan of hard coding substrings like this as it's rather frail.

Nosferican: I am experiencing the issue on Mac OS Catalina 10.15.4 with Visual Studio Code (works fine from the Terminal and VS Code has full disk access)
```
julia> using TimeZones
julia> localzone()
ERROR: IOError: could not spawn `systemsetup -gettimezone`: no such file or directory (ENOENT)
Stacktrace:
 [1] _spawn_primitive(::String, ::Cmd, ::Array{Any,1}) at ./process.jl:99
 [2] #550 at ./process.jl:112 [inlined]
 [3] setup_stdios(::Base.var"#550#551"{Cmd}, ::Array{Any,1}) at ./process.jl:196
 [4] _spawn at ./process.jl:111 [inlined]
 [5] open(::Cmd, ::Base.DevNull; write::Bool, read::Bool) at ./process.jl:370
 [6] open at ./process.jl:361 [inlined]
 [7] open(::Cmd, ::String, ::Base.DevNull) at ./process.jl:341
 [8] read(::Cmd) at ./process.jl:410
 [9] read(::Cmd, ::Type{String}) at ./process.jl:421
 [10] macro expansion at /Users/jbsc/.julia/packages/Mocking/U41JO/src/mock.jl:29 [inlined]
 [11] localzone() at /Users/jbsc/.julia/packages/TimeZones/OSWcm/src/local.jl:21
 ...
```
For example, for LibPQ.jl, I have to first precompile the package from the terminal and then I can call it within VS Code but could not precompile it from VS Code due to the error above (running VS Code as root didn't change that).

I also looked at
```
julia> read(`date +"%Z", String)
```

The symlink has the following syntax,
```
/var/db/timezone/zoneinfo/*
```
I could change it to 
```
match(r"(?<=/var/db/timezone/zoneinfo/).*", readlink("/etc/localtime")).match
```
but that would be equally frail.

omus: Interesting. The error `IOError: could not spawn `systemsetup -gettimezone`: no such file or directory (ENOENT)` implies that the process itself cannot be found. I'd be interested in the output from these commands under these conditions:
```julia
run(`which systemsetup`)
ENV["PATH"]
```
The regular expression in the code you removed can handle the example symlink you provided:
```julia
julia> name = "/var/db/timezone/zoneinfo/*"
"/var/db/timezone/zoneinfo/*"

julia> match(r"(?<=zoneinfo/).*$", name).match
"*"
```

Nosferican: New environment. 
```
 [f269a46b] TimeZones v1.1.0
```
From VS Code (`julialang.language-julia` `v0.15.24`)

```
julia> run(`which systemsetup`)
ERROR: failed process: Process(`which systemsetup`, ProcessExited(1)) [1]

Stacktrace:
 [1] pipeline_error at ./process.jl:525 [inlined]
 [2] run(::Cmd; wait::Bool) at ./process.jl:440
 [3] run(::Cmd) at ./process.jl:438
 [4] top-level scope at REPL[9]:1
 [5] eval(::Module, ::Any) at ./boot.jl:331
 [6] eval_user_input(::Any, ::REPL.REPLBackend) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [7] run_backend(::REPL.REPLBackend) at /Users/jbsc/.julia/packages/Revise/C272c/src/Revise.jl:1075
 [8] top-level scope at none:0
```

```
julia> ENV["PATH"]
ERROR: KeyError: key "PATH" not found
Stacktrace:
 [1] (::Base.var"#459#460")(::String) at ./env.jl:79
 [2] access_env at ./env.jl:43 [inlined]
 [3] getindex(::Base.EnvDict, ::String) at ./env.jl:79
 [4] top-level scope at REPL[10]:1
 [5] eval(::Module, ::Any) at ./boot.jl:331
 [6] eval_user_input(::Any, ::REPL.REPLBackend) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [7] run_backend(::REPL.REPLBackend) at /Users/jbsc/.julia/packages/Revise/C272c/src/Revise.jl:1075
 [8] top-level scope at none:0
```

omus: Thank you. Ok the VS Code environment is very restrictive. I'm completely fine with relying on using `/etc/localtime` for time zone detection as `systemsetup` only actually works if you are running as `root` which will be very rare:
```
$ systemsetup -gettimezone; echo $?
You need administrator access to run this tool... exiting!
0
```

omus: This change should work and allow the tests to work:
```suggestion
        # Link will point to something like "/usr/share/zoneinfo/Europe/Warsaw"
        link = @mock readlink("/etc/localtime")
        name = match(r"(?<=zoneinfo/).*$", link).match
        return TimeZone(name, mask)
```

test/local_mocking.jl
```diff
@@ -11,11 +11,6 @@ win_name = name == "Europe/Warsaw" ? "Central European Standard Time" : "Samoa S
 tz = TimeZone(name)
 
 if Sys.isapple()
-    # Determine time zone via systemsetup.
-    patch = @patch read(cmd::AbstractCmd, ::Type{String}) = "Time Zone:  " * name * "\n"
-    apply(patch) do
-        @test localzone() == tz
-    end
```
omus, line 14:
```suggestion
```

omus:
For details on the changes see this discussion: https://github.com/JuliaTime/TimeZones.jl/pull/256#discussion_r411787766

TLDR; Under certain environments `systemsetup` cannot be found causing `localzone()` to unnecessarily fail. As `systemsetup` typically fails as it requires admin access we'll just remove it.

Nosferican:
It seems there are a couple of `FixedTimeZone has no field transitions` errors.

omus:
I'll dig into those tomorrow and get this merged

Nosferican:
🙏! Glad I could be of help.

Nosferican:
Tests clear. Bump a patch release and tag?

omus:
Thanks for your contribution. The fix is now available in version 1.1.1

PR #258, opened 2020-04-28
Support TZ environmental variable on macOS

omus:
The `TZ` environment variable is supported by Linux and macOS so we should support it on both as well. I believe this was an oversight on my part when originally writing `localzone`.

Additionally, I noticed an issue when "/etc/localtime" links to a path containing no "/" characters then we'd fall back to reading the tzfile instead of using the preferred TimeZones.jl pre-compiled time zones. This is a rather minor issue but it resulted in the creation of the `_path_tz_name` and its tests.

### DIFF
--- a/src/local.jl
+++ b/src/local.jl
@@ -115,17 +115,13 @@ function localzone()
         end
 
         # systemd distributions use symlinks that include the zone name,
-        # see manpage of localtime(5) and timedatectl(1)
+        # see man page of localtime(5) and timedatectl(1)
         link = "/etc/localtime"
         if @mock islink(link)
-            filepath = @mock readlink(link)
-            pos = findnext(isequal('/'), filepath, 1)
-
-            while pos !== nothing
-                name = SubString(filepath, pos + 1)
-                istimezone(name, mask) && return TimeZone(name, mask)
-                pos = findnext(isequal('/'), filepath, pos + 1)
-            end
+            # Link target example: "/usr/share/zoneinfo/Europe/Warsaw"
+            target = @mock readlink(link)
+            name = _path_tz_name(target, mask)
+            name !== nothing && return TimeZone(name, mask)
         end
 
         # No explicit setting existed. Use localtime
@@ -148,6 +144,18 @@ function localzone()
 
     error("Failed to find local time zone")
 end
+
+function _path_tz_name(path::AbstractString, mask::Class=Class(:ALL))
+    i = 0
+    while i !== nothing
+        name = SubString(path, i + 1)
+        istimezone(name, mask) && return name
+        i = findnext(isequal('/'), path, i + 1)
+    end
+
+    return nothing
+end
+
 
 # Using conditional expressions `(?(condition)yes-regexp)` as control flow to indicate that
 # that a captured group is dependent on a previous group begin matched. This could also be
--- a/src/local.jl
+++ b/src/local.jl
@@ -17,19 +17,14 @@ function localzone()
     # legacy names.
     mask = Class(:STANDARD) | Class(:LEGACY)
 
-    @static if Sys.isapple()
-        # Link will point to something like "/usr/share/zoneinfo/Europe/Warsaw"
-        link = @mock readlink("/etc/localtime")
-        name = match(r"(?<=zoneinfo/).*$", link).match
-        return TimeZone(name, mask)
-    elseif Sys.isunix()
+    @static if Sys.isunix()
         name = ""
 
         # Try getting the time zone from the "TZ" environment variable
         # http://linux.die.net/man/3/tzset
         #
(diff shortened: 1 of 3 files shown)


src/local.jl
```diff
@@ -59,7 +54,7 @@ function localzone()
 
             if startswith(name, '/')
                 return @mock open(name) do f
-                    read_tzfile(f, "local")
+                    read_tzfile(f, something(_path_tz_name(name, mask), "local"))
```
nicoleepp, line 57: This is meant to error if it can't construct the path right?

src/local.jl
```diff
@@ -78,54 +73,52 @@ function localzone()
...
-                        name = matched["name"]
-                        name = replace(name, ' ' => '_')
-                        break
+            # CentOS has a ZONE setting in /etc/sysconfig/clock,
+            # OpenSUSE has a TIMEZONE setting in /etc/sysconfig/clock and
+            # Gentoo has a TIMEZONE setting in /etc/conf.d/clock
+
+            zone_re = r"(TIME)?ZONE\s*=\s*\"(?<name>.*?)\""
+            for filepath in ("/etc/sysconfig/clock", "/etc/conf.d/clock")
+                (@mock isfile(filepath)) || continue
```
nicoleepp, line 99: Is this just the display making the indentation look off?

src/local.jl
omus: If the path specified by the TZ environment variable doesn't exist or it isn't a tzfile an error will be raised which is expected. We could possibly raise a custom error message or add a comment indicating that.

src/local.jl
omus: It's just the parenthesis. Unfortunately they're necessary otherwise the `@mock` will be interpreted as:
```julia
@mock (isfile(filepath)) || continue)
```

omus: FYI here's a view of the PR without the whitespace changes: https://github.com/JuliaTime/TimeZones.jl/pull/258/files?w=1

src/local.jl
nicoleepp: I think the default julia error if `nothing` is returned is probably enough information

src/local.jl
```diff
@@ -149,6 +142,20 @@ function localzone()
     error("Failed to find local time zone")
 end
 
+# Extract a time zone name from a path.
+# e.g. "/usr/share/zoneinfo/Europe/Warsaw" becomes "Europe/Warsaw"
+function _path_tz_name(path::AbstractString, mask::Class=Class(:ALL))
+    i = 0
```
nicoleepp, line 148: This changes the behaviour as before we were starting off with i = 1 in the default case?

test/local.jl
```diff
@@ -34,6 +35,26 @@ using TimeZones: TimeZone, localzone, parse_tz_format
...
+        for name in ("UTC", "Europe/Warsaw", "America/Indiana/Indianapolis")
+            # Use eval to improve readability of tests when failures occur
+            @eval begin
+                @test _path_tz_name($name) == $name
+                @test _path_tz_name("/usr/share/zoneinfo/$($name)") == $name
+                @test _path_tz_name("/var/db/timezone/zoneinfo/$($name)") == $name
+            end
+        end
+
+        # The algorithm that extracts the tz name could accidentally return the valid
```
nicoleepp, line 51: Is this expected behaviour or are we testing that this doesn't happen anymore? Comment could use clarification either way

src/local.jl
omus: Oh, I think I misunderstood what you meant. The `something` here is just there to fall back to "local" if `_path_tz_name` returns `nothing`:
```julia
julia> something(nothing, "local")
"local"
```
Essentially we want to use a valid time zone name if we can find one but if this is some crazy custom time zone we'll just name it "local"

src/local.jl
omus: Before this PR `i = findnext(isequal('/'), path, 1)`. The reason this was problematic was that for time zone names without a `/` we would fail to return a time zone. e.g. "UTC". (I'll note this also fails in other specific relative paths but that is harder to explain)

test/local.jl
omus: I'll expand on this comment

src/local.jl
omus: I think looking that the commit where this change was made should clear things up: https://github.com/JuliaTime/TimeZones.jl/pull/258/commits/0e5b8e346736131fd706193be31ea025ef4d8c8f

test/local.jl
omus: Let me know if the new comment is clearer

src/local.jl
nicoleepp: Oh that makes sense, reading is hard for me sometimes :)

PR #259, opened 2020-04-28
Full tzset support

### LINKED ISSUE #225: Unable to detect timezone using local
Hi, it appears that TimeZones throws out errors for some UNIX/LINUX systems where the timezone handling might be slightly different.

Is there any way to improve upon TimeZones handling of different formats?

julia> using Memento
ERROR: InitError: ArgumentError: Unhandled TZ environment variable format: "EST5EDT,M3.2.0,M11.1.0"
Stacktrace:
 [1] parse_tz_format(::String) at /n/home07/nwong/.julia/packages/TimeZones/A5RcB/src/local.jl:170
 [2] localzone() at /n/home07/nwong/.julia/packages/TimeZones/A5RcB/src/local.jl:43
 [3] Type at /n/home07/nwong/.julia/packages/Memento/UgJr2/src/formatters.jl:35 [inlined]
 [4] #config!#71(::String, ::Dict{AbstractString,Int64}, ::Bool, ::Bool, ::Bool, ::Bool, ::typeof(Memento.config!), ::Memento.Logger, ::String) at /n/home07/nwong/.julia/packages/Memento/UgJr2/src/config.jl:39
 [5] config! at /n/home07/nwong/.julia/packages/Memento/UgJr2/src/config.jl:36 [inlined]
 [6] #config!#70 at /n/home07/nwong/.julia/packages/Memento/UgJr2/src/config.jl:28 [inlined]
 [7] config! at /n/home07/nwong/.julia/packages/Memento/UgJr2/src/config.jl:28 [inlined]
 [8] #config!#69 at /n/home07/nwong/.julia/packages/Memento/UgJr2/src/config.jl:25 [inlined]
 [9] config! at /n/home07/nwong/.julia/packages/Memento/UgJr2/src/config.jl:25 [inlined]
 [10] __init__() at /n/home07/nwong/.julia/packages/Memento/UgJr2/src/Memento.jl:62
 [11] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:685
 [12] _require_search_from_serialized(::Base.PkgId, ::String) at ./loading.jl:765
 [13] _require(::Base.PkgId) at ./loading.jl:990
 [14] require(::Base.PkgId) at ./loading.jl:911
 [15] require(::Module, ::Symbol) at ./loading.jl:906
during initialization of module Memento

Please see the InitError argument on the first few lines.  Although this is an error caused by Memento, the root cause is due to the TimeZone variable.

omus:
Adds support for the direct time zone representation as described by the [tzset(3)](http://man7.org/linux/man-pages/man3/tzset.3.html) man page.

This work provided me with a good example to experiment with new efficient parsing interfaces which led to: `parsesub_*`. I'm hoping to make use of a similar pattern for `ZonedDateTime` and `Dates` parsing in the future.

Fixes #225


omus:
I'd prefer not to hold up support for tzset(3) with this new parsing interface. I've decided to mark it as internal for now so I merge things as they are now and rework the interface later.

src/local.jl
```diff
@@ -156,36 +159,20 @@ function _path_tz_name(path::AbstractString, mask::Class=Class(:ALL))
...
-daylight saving time.
 """
 function parse_tz_format(str::AbstractString)
-    tz = tryparse_tz_format(str)
-
-    if tz !== nothing
+    x = _parsesub_tz(str)
+    if x isa Tuple
+        tz, i = x
         return tz
```
mattBrzezinski, line 173:
```suggestion
    if x isa Tuple
        return x[1]
```

We're also duplicating this code below on [here](https://github.com/JuliaTime/TimeZones.jl/pull/259/files#diff-8044bd4fb98ddd8eb84ea707c8d7d00bR186), maybe make this into its own function?

src/parse.jl
```diff
@@ -101,3 +102,523 @@ function ZonedDateTime(str::AbstractString, format::AbstractString; locale::Abst
...
+    val = tryparsenext_base10(str, i, len)
+    if val === nothing
+        return ParseNextError("Expected $name minute digits after colon delimiter", str, i)
+    end
+    minute, ii = val
+    if minute < 0 || minute > 59
+        return ParseNextError("Minutes outside of expected range [0, 59]", str, i, prevind(str, ii))
+    end
+    i = ii
+    i > len && @goto done
```
mattBrzezinski, line 242: Not sure how others feel about `goto`s, but an alternative could be do make `done` a nested function, and just call that here while passing along `i`?

src/local.jl
omus: I prefer the way it was implemented as having the explicit `tz` variable name makes it clearer what is being returned. As for the duplicate code I'm still iterating on this interface. Any suggestions for what this function should look like?

src/parse.jl
omus: The `goto`s are mainly being used for performance reasons. Even if I make a nested function I'd still need to do add a `return` which I'm not sure would help with clairity

test/tz_env.jl
```diff
@@ -0,0 +1,203 @@
...
+# really deals with corner cases which are not required by any actual time zone definition
+# supporting for these features is purely academic.
+#
+# Here are some examples of the platform specific differences. Under macOS time zone offset
+# only supports hours under ±7 days (167 hours) and hours beyond this range result in
+# falling back to UTC:
+#
+# ```bash
+# $ TZ="FOO-0" date -r 0
+# Thu  1 Jan 1970 00:00:00 UTC
```
omus, line 16: I'm not sure what happened but this is incorrect for macOS:
```suggestion
# Thu  1 Jan 1970 00:00:00 FOO
```

test/tz_env.jl
```diff
@@ -0,0 +1,203 @@
...
+# $ TZ="FOO-1BAR-2,1/0,1/168" date -r 1609455600
+# Thu 31 Dec 2020 23:00:00 UTC
+# ```
+#
+# Under Linux the transition time appears to have no limit:
+#
+# ```bash
+# $ TZ="FOO-1BAR-2,1/0,1/8761" date --date @1609455600
+# Fri Jan  1 00:00:00 FOO 2021
+# $ TZ="FOO-1BAR-2,1/0,1/8762" date --date @1609455600
```
omus, line 55: Just for consistency with examples above
```suggestion
#
# $ TZ="FOO-1BAR-2,1/0,1/8762" date --date @1609455600
```

src/parse.jl
mattBrzezinski: Hmm I don't have permissions to resolve my own comments, but I'm okay with keeping them in if this is the case.

src/local.jl
mattBrzezinski: Fair enough. Alternatively we could do something similar to below (note; `_try` is quite a poor name);

```julia
function parse_tz_format(str::AbstractString; _try=false)
    x = _parsesub_tz(str)
    if x isa Tuple
        tz, i = x
        return tz
    else
        return _try ? nothing : throw(ParseNextError("Unhandled TZ environment variable. $(x.msg)", x.str, x.s, x.e))
    end
end
```

omus: I should point out that `parse` and `tryparse` are conventions set by Julia. I'll note the `parsesub` idea is my attempt to improve on the design.

I could combine these functions but I think it would need to be something like:
```julia
for f in (:parse, :tryparse)
    @eval function $(f)_tz_format(...)
        ...
    end
end
```

mattBrzezinski: Ah didn't know this was a standard practice in Julia. In that case this can stay as is, we're not really gaining anything from combining them in this case and it may just be complicated to read.

test/parse.jl
```diff
@@ -76,3 +77,265 @@ end
...
+
+        @testset "Last Saturday in January" begin
+            # Note: 2017 - 2020 only have 4 Saturdays in January
+            f1, i = _parsesub_tzdate("M1.5.6")
+            @test i == 7
+
+            f2, i = _parsesub_tzdate("M1.4.6")
+            @test i == 7
+
+            @test f1(2020) == f1(2020) == Date(2020, 1, 25)
```
mattBrzezinski, line 229:
```suggestion
            @test f1(2020) == f2(2020) == Date(2020, 1, 25)
```

omus: Good eye. Thanks!

omus:
Just rebased to include latest changes.

test/parse.jl
```diff
@@ -76,3 +77,265 @@ end
...
+    dst_start_time_msg = "Expected daylight saving start time and instead found end of string"
+    dst_end_missing_msg = "Expected to find daylight saving end and instead found end of string"
+    dst_end_date_msg = "Unable to parse daylight saving end date. Expected date and instead found end of string"
+
+    @testset "empty" begin
+        @test _parsesub_tz("") == (FixedTimeZone("UTC"), 1)
+    end
+
+    @testset "standard time only" begin
+        @test _parsesub_tz("FOO") == ParseNextError("Expected standard offset and instead found end of string", "FOO", 4, 3)
```
mattBrzezinski, line 248: Does not need to be included in here, but what are your thoughts on storing error messages in `src/exceptions.jl`, that way we can hard code these strings in one area and just reference them throughout a package?

src/parse.jl
```diff
@@ -101,3 +102,523 @@ function ZonedDateTime(str::AbstractString, format::AbstractString; locale::Abst
...
+        return ParseNextError("Time zone abbreviation must be at least three $char_type", str, name_start, name_end)
+    else
+        name = SubString(str, name_start, name_end)
+        return name, i
+    end
+end
+
+
+"""
+    _parsesub_offset(str, [i, len]) -> Union{Tuple{Integer, Integer}, Exception}
```
nicoleepp, line 178: We might want to start using ResultTypes in this package for type stability? 

https://github.com/iamed2/ResultTypes.jl

src/parse.jl
```diff
@@ -101,3 +102,523 @@ function ZonedDateTime(str::AbstractString, format::AbstractString; locale::Abst
...
+    i::Integer=firstindex(str),
+    len::Integer=lastindex(str),
+)
+    # An empty or unset TZ environmental variable defaults to UTC
+    if len == 0
+        return FixedTimeZone("UTC"), i
+    end
+
+    x = _parsesub_tzabbr(str, i, len)
+    if x isa Tuple
```
nicoleepp, line 470:
```suggestion
    x isa Tuple ? (stdname, i = x) : return x
```

Might be easier to read here and below. Feel free to resolve if you don't agree

src/exceptions.jl
```diff
@@ -46,3 +46,50 @@ function Base.showerror(io::IO, e::UnhandledTimeError)
...
+    e::Int
+end
+
+function ParseNextError(msg, str, s)
+    len = lastindex(str)
+    ParseNextError(msg, str, s, s >= len ? len : s)
+end
+
+function Base.showerror(io::IO, e::ParseNextError)
+    print(io, "$ParseNextError: ")
```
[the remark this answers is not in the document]
omus: This isn't recursive as `"$ParseNextError"` calls `string(::Type{ParseNextError})`

src/parse.jl
omus: I was avoiding using other packages here mainly as I want to use this pattern in the Dates stdlib. However I should do some benchmarking using ResultTypes

test/parse.jl
omus: The exceptions in the package itself aren't ever replicated so it would only be for the benefit of the tests. In some cases with the tests I reference the error messages via a variable which has a similar effect.

I think having constants in the source would probably make the code harder to read so I'm inclined to keep things as is.

src/parse.jl
omus: I'm not sure that's better. I want to do some refactoring of this code later and I expect this pattern will be hidden by a macro

omus:
Rebased and applied fixups

src/parse.jl
omus: I'll note that with newer versions of Julia there isn't much a performance impact for using `Union`s. However there may be performance improvements still by using ResultTypes. I have the beginnings of some benchmarks using PkgBenchmarks which once established it would be good to try out ResultTypes to see if I can get additional speedups. With all of that said I think I'll just continue to sit on this PR if I continue to wait for the rest of my experimental work.

I'll leave this unresolved for now and I'll be sure to reference the ResultTypes benchmarks here.

PR #261, opened 2020-05-04
Default to using 2020a tzdata

omus:
Release notes: http://mm.icann.org/pipermail/tz-announce/2020-April/000058.html

At the present time there is no [windowsZones.xml](https://github.com/unicode-org/cldr/blob/master/common/supplemental/windowsZones.xml) file update.


mattBrzezinski:
Looks like we need to enable failures on `nightly` builds for the CI in this project.

omus:
TimeZones.jl strives to be compatible with all versions of Julia including `nightly`. The nightly failure here is a Julia failure and not from due to TimeZones

omus:
I've filed an issue: https://github.com/JuliaLang/julia/issues/35732

PR #262, opened 2020-05-04
Switch to travis-ci.com

omus:


omus:
Expected nightly failure

PR #270, opened 2020-07-06
Migration to Artifacts.

### LINKED ISSUE #63: Move tzdata download into an artifact
See [discussion](https://github.com/JuliaTime/TimeZones.jl/pull/58#discussion_r109893372) for details.

### LINKED ISSUE #266: Usage of Package Server for binary dependencies/option to specify server.
Hi, I see some packages are being downloaded from https://www.iana.org/ at https://github.com/JuliaTime/TimeZones.jl/blob/master/src/tzdata/download.jl#L89
would it be possible to use Artifacts https://julialang.github.io/Pkg.jl/v1/artifacts/ for this, so during installation, client would reach only to package server instead of iana.org directly?

Or, as a simpler solution, would it be possible to have some environment variable to specify the domain/base URI to use for binaries?
For Julia behind corporate firewall which can reach only whitelisted domains, it would be useful to have an option to specify this.

### LINKED ISSUE #269: Windows timezone XML on unicode website seems dead
Website http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml on https://github.com/JuliaTime/TimeZones.jl/blob/master/src/winzone/WindowsTimeZoneIDs.jl#L9 does not return XML.

racinmat:
It's something, definitely not perfect, but should be enough to stabilize this package and allow utilization of Package server and it's caching layer. 
It still does quite a lot of I/O to the dep, but it downloads the tarballs and xml only as Artifacts, no direct downloads should be present now. 
Fixes #269.
Fixes #266.
Fixes #63.

Not sure if it should be minor or major version bump, because it's dropping Julia < 1.3 because of Artifacts support.


racinmat:
Hm, here https://travis-ci.com/github/JuliaTime/TimeZones.jl/jobs/358103893 it's not working for Julia 1.3 for no apparent reason, and it's showing me error which does not make sense, I assume it's something weird with macros, it works for Julia 1.4 and newer.

racinmat:
I had to set 1.4 as minimal version because of https://github.com/JuliaLang/Pkg.jl/issues/1912.
I also allowed nightly to fail, and added 1.5 to tests. 
Is this PR ok, @omus ?

Artifacts.toml
```diff
@@ -0,0 +1,9759 @@
+[[tzdata_1996l]]
+arch = "x86_64"
+git-tree-sha1 = "835e88b3b0a560e79dfc0a2a6c350adeaa4ef907"
+lazy = true
+os = "windows"
```
omus, line 5: The tzdata artifacts aren't OS specific so this can be removed safely and this file can be 4x smaller!

.travis.yml
```diff
@@ -7,7 +7,7 @@ stages:
 .doc_test: &doc_test
   stage: "Documentation (Test)"
   os: linux
-  julia: 1.0
+  julia: 1.4
```
omus, line 10:
```suggestion
  julia: 1
```

.travis.yml
```diff
@@ -16,7 +16,7 @@ stages:
 .doc_deploy: &doc_deploy
   stage: "Documentation (Deploy)"
   os: linux
-  julia: 1.0
+  julia: 1.4
```
omus, line 19:
```suggestion
  julia: 1
```

.travis.yml
```diff
@@ -29,19 +29,18 @@ os:
   - osx
   - windows
 julia:
-  - 1.0  # LTS
-  - 1.4  # Latest release
+  - 1.4  # Latest stable release
```
omus, line 32: I'd like to keep supporting the Long Term Stable release. I think a way that can be done here is to use Artifacts on 1.3+ and use the direct download for earlier versions

.travis.yml
```diff
@@ -29,19 +29,18 @@ os:
...
+  - 1.5  # Latest release
   - nightly
 arch:
   - x64
   - x86
+
 jobs:
+  allow_failures:
+    - julia: nightly
```
omus, line 41: I disallow nightly failures on this package as I want to know immediately if there is an issue with this package and the latest version of Julia

Project.toml
```diff
@@ -16,7 +17,7 @@ Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 EzXML = "0.9.1, 1"
 Mocking = "0.7"
 RecipesBase = "0.8, 1.0"
-julia = "1"
+julia = "1.4"
```
omus, line 20: I think Artifacts work on 1.3 right?

README.md
```diff
@@ -18,7 +18,7 @@ TimeZones.jl
 * ZonedDateTime-Period arithmetic [similar to that of DateTime](https://docs.julialang.org/en/stable/manual/dates/#TimeType-Period-Arithmetic-1)
 * Local system time zone information as a TimeZone
 * Current system time in any TimeZone
-* Support for reading the [tzfile](http://man7.org/linux/man-pages/man5/tzfile.5.html) format
+* Support for reading the [tzfile](https://man7.org/linux/man-pages/man5/tzfile.5.html) format
```
omus, line 21: It would be best if you make these kind of changes in a separate PR then they can get reviewed and merged independently

dev/download.jl
```diff
@@ -0,0 +1,89 @@
...
+using TimeZones: DEPS_DIR
+using TimeZones.TZData: LATEST_FILE, LATEST_DELAY, latest_version, tzdata_version_dir, set_latest
+
+if Sys.iswindows()
+    if isdefined(Base, :LIBEXECDIR)
+        const exe7z = joinpath(Sys.BINDIR, Base.LIBEXECDIR, "7z.exe")
+    else
+        const exe7z = joinpath(Sys.BINDIR, "7z.exe")
+    end
+end
```
omus, line 11: You can just use the `exe7z` as defined in `TimeZones.TZData` instead of re-implementing it here

dev/generate_artifacts.jl
```diff
@@ -0,0 +1,82 @@
...
+        # Now bind that hash within our `Artifacts.toml`.  `force = true` means that if it already exists,
+        # just overwrite with the new content-hash.  Unless the source files change, we do not expect
+        # the content hash to change, so this should not cause unnecessary version control churn.
+        download_data = [(tzdata_url(version), content_sha)]
+        bind_artifact!(artifacts_toml, "tzdata_$version", tzfile_hash, lazy=true, download_info=download_data, platform=platform_key_abi())
+    end
+end
+
+# this can not work in current state because Artifacts require tarball, and there is only the direct XML
+# # and now, add windows xml thingy
```
omus, line 60: If Artifacts can handle zip files then you can use `https://github.com/unicode-org/cldr/archive/<tag>.zip` to get a copy of the entire CLDR repo

Project.toml
omus: See: https://github.com/JuliaTime/TimeZones.jl/pull/270#issuecomment-655798160

omus:
Overall this is looking very promising. Thanks for this. I think the most major change I'd like to see would be to keep Julia 1.0 support. Essentially this just involves leaving the original download code as it was and then just using a `VERSION >= v"1.4"` check to use Artifacts on those later versions of Julia.

.travis.yml
racinmat: 1 would be running it on 1.5, which is not stable, but only RC, or now?

Artifacts.toml
racinmat: Yes. The `git-tree-sha1` is platform specific, which confused me, see: https://github.com/JuliaLang/Pkg.jl/issues/1914
but there I learned it's not controlled on Windows, so I'll make it platform independent and use the linux ones.

.travis.yml
racinmat: So I should comment it out? 
Ok, I didn't want to make PR with CI crashing, but if you're ok with CI showing "failed" just because it fails on nightly, I don't have a problem with that.

.travis.yml
racinmat: I can't use Julia 1.3 here because of https://github.com/JuliaLang/Pkg.jl/issues/1912.
And the <1.4 compatibility should be done by having `if` on Julia version with direct download inside of `build`?

dev/download.jl
racinmat: I thought I'd remove it from `TimeZones.TZData` because it's not needed when it's using Artifacts, but if you want to have also the nonartifact support for longer version, then I'll put it back and use it from there.

dev/generate_artifacts.jl
racinmat: Oh, that's true, I didn't want to download the whole repo because of single xml, but I guess it will be the best current solution.

dev/generate_artifacts.jl
```diff
@@ -0,0 +1,82 @@
...
+"2009k", "2009l", "2009m", "2009n", "2009o", "2009p", "2009q", "2009r", "2009s", "2009t", "2009u", "2010a", "2010b",
+"2010c", "2010d", "2010e", "2010f", "2010g", "2010h", "2010i", "2010j", "2010k", "2010l", "2010m", "2010n", "2010o",
+"2011a", "2011b", "2011c", "2011d", "2011e", "2011f", "2011g", "2011h", "2011i", "2011j", "2011k", "2011l", "2011m",
+"2011n", "2012a", "2012b", "2012c", "2012d", "2012e", "2012f", "2012g", "2012h", "2012i", "2012j", "2013a", "2013b",
+"2013c", "2013d", "2013e", "2013f", "2013g", "2013h", "2013i", "2014a", "2014b", "2014c", "2014d", "2014e", "2014f",
+"2014g", "2014h", "2014i", "2014j", "2015a", "2015b", "2015c", "2015d", "2015e", "2015f", "2015g", "2016a", "2016b",
+"2016c", "2016d", "2016e", "2016f", "2016g", "2016h", "2016i", "2016j", "2017a", "2017b", "2017c", "2018a", "2018b",
+"2018c", "2018d", "2018e", "2018f", "2018g", "2018h", "2018i", "2019a", "2019b", "2019c", "2020a", "93g", "94a", "94b",
+"94d", "94e", "94f", "94h", "95b", "95c", "95d", "95e", "95f", "95g", "95h", "95i", "95k", "95l", "95m", "96a", "96b",
+"96c", "96d", "96e", "96h", "96i", "96k"]
```
omus, line 29: An automated way of getting this listing:
```julia
const VERSIONS = let
    file = download("https://data.iana.org/time-zones/releases/")
    html = read(file, String)
    rm(file)
    [m[:version] for m in eachmatch(r"href=\"tzdata(?<version>(?:\d{2}){1,2}[a-z]?).tar.gz\"", html)]
end
```
I'll note this excludes the `latest` version but it should probably be removed as it would only be the latest version as of the time the Artifacts.toml file was built and not the actual latest available by IANA

.travis.yml
omus: This is equivalent to 1.4 currently and won't be updated to 1.5 until sometime after it has officially released. See: https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/julia.rb#L185

dev/generate_artifacts.jl
racinmat: Ok, so I'll use this and remove the `latest` from artifacts, so when someone requires `latest`, the newest date from artifacts should be used?

.travis.yml
omus: That is what I'm thinking should be done

.travis.yml
omus: I'd just leave this how it was. I'll fix the nightly issue in another PR

.travis.yml
racinmat: Oh, I see, my bad, I somehow remember it wrongly.

Artifacts.toml
omus: Interesting. Thanks for providing that reference

README.md
racinmat: Ok. I wanted to make the whole CI thing passing before merging the PR, but I'll leave it for another PR.

dev/generate_artifacts.jl
omus: I was thinking maybe using `latest` on 1.4+ would fall back to the direct download (as in how things used to work). That could be problematic though as I think packages will require artifacts for this kind of use case eventually.

Maybe the solution is to use the latest from the artifacts and check that version against the latest release from IANA. If it's different a warning can be provided that the artifact file needs to be regenerated.

I'll note primarily `latest` is probably only used by me to alert me that the package needs to be updated.

dev/generate_artifacts.jl
omus: Did you open an issue against Pkg for allowing non-archive downloads?

.travis.yml
omus: No problem at all. It's not very intuitive

dev/generate_artifacts.jl
racinmat: No, should I?

dev/generate_artifacts.jl
racinmat: Ok, that's a good point.

dev/generate_artifacts.jl
omus: I think this is use case that is reasonable and an I think an issue is warranted

.travis.yml
aviks: It would be useful if the direct download is also done via BinaryProvider in `<v1.4`. That'll make things a little more consistent IMO. One example is at: https://github.com/invenia/LibPQ.jl/blob/master/deps/build.jl (you'll have to ignore the platform specific stuff since that's not relevant here)

README.md
omus: I'll extracted this as a separate PR: https://github.com/JuliaTime/TimeZones.jl/pull/273

.travis.yml
omus: I'm not sure switching BinaryProvider has any benefits in this case besides being more consistent with other packages. One scenario to consider here is that although TimeZones specifies a version of tzdata to use with each release the package also supports using other versions of tzdata with a TimeZones release.

I think for this specific PR we should keep things straight forward an just add artifact support for Julia 1.4+ and leave the direct download as it was for now.

dev/generate_artifacts.jl
racinmat: So, I made an issue, we'll see. https://github.com/JuliaLang/Pkg.jl/issues/1921

racinmat:
I have problem with defining this function:
```julia
if VERSION >= v"1.4"
    function active_dir()
        version = active_version()
        archive = @artifact_str "tzdata_$version"
        !isdir(archive) && error("Missing $version tzdata archive")
        return archive
    end
end
```
although it's wrapped in this if, and does not execute, the parser is still going through it and thus I get error 
`ERROR: LoadError: LoadError: LoadError: LoadError: UndefVarError: @artifact_str not defined``. 
Any ideas?

.travis.yml
racinmat: Added running julia on 1.0, 1.3, 1.4, 1.5 and nightly, passing everywhere so it should be ok now, I hope.

.travis.yml
racinmat: Ok, I commented it out, it passes also on nightly now, nice.

README.md
racinmat: Ok, reverted to http in this PR

dev/download.jl
racinmat: Everything returned back on it's place, using `extract` and `tzdata_download` from the package.

dev/generate_artifacts.jl
racinmat: Done as you suggested, stopped using `latest` in `Artifacts.toml`.

racinmat:
Some things are a bit cumbersome to make it not fail during precompilation with all these ifs. 
It should be ok now, 
Not sure if it makes sense to try to do something with coverage.
Ready to merge @omus or is there sth. more to do?

src/tzdata/TZData.jl
```diff
@@ -2,6 +2,14 @@ module TZData
 
 using Printf
 using ...TimeZones: DEPS_DIR
+if VERSION >= v"1.4"
+    using Pkg.Artifacts
```
staticfloat, line 6: Pkg.Artifacts is actually available in v1.3

racinmat: Yes, but because of https://github.com/JuliaLang/Pkg.jl/issues/1912 it's not working with variable artifacts, so I don't see the point in using the module if I can't actually install the artifacts in this version because of this bug.

staticfloat: Ah, I see.

omus:
> Ready to merge @omus or is there sth. more to do?

I'll take time to review this tomorrow. I appreciate your work on this.

src/tzdata/TZData.jl
```diff
@@ -2,6 +2,14 @@ module TZData
...
+else
+    # to avoid fail when Julia tried to precompile even non-compatible code
+    macro artifact_str(name)
+        :(throw("this should never bee called"))
+    end
```
omus, line 11: Should be able to avoid having to define this if `@static` is used with `if VERSION` checks inside of functions

src/tzdata/build.jl
```diff
@@ -38,25 +38,53 @@ function build(
...
     end
 
-    archive = joinpath(archive_dir, "tzdata$version.tar.gz")
-
-    # Avoid downloading a tzdata archive if we already have a local copy
-    if version == "latest" || !isfile(archive)
-        @info "Downloading $version tzdata"
-        archive = tzdata_download(version, archive_dir)
-
+    if VERSION >= v"1.4"
```
omus, line 41:
```suggestion
    @static if VERSION >= v"1.4"
```

src/tzdata/build.jl
```diff
@@ -38,25 +38,53 @@ function build(
...
+        # todo: I don't have latest here and it should not be used as latest
         if version == "latest"
-            m = match(TZDATA_VERSION_REGEX, basename(archive))
+            m = match(TZDATA_VERSION_REGEX, "tzdata$version.tar.gz")
             if m !== nothing
                 version = m.match
                 @info "Latest tzdata is $version"
             end
+
+            artifact_dir = @artifact_str "tzdata_$version"
```
omus, line 51: If `version == "latest" then `match` will return `nothing` and this will resolve to "tzdata_latest" which doesn't exist as an artifact.

src/tzdata/build.jl
```diff
@@ -38,25 +38,53 @@ function build(
...
+        artifact_dir = @artifact_str "tzdata_$version"
+        set_latest(version, now_utc)
+        #if !isfile(LATEST_FILE)
+        #    set_latest(version, now_utc)
+        #end
+
+        if !isempty(tz_source_dir)
+            @info "Copying region data from version $version"
+            for region in setdiff(regions, CUSTOM_REGIONS)
+                cp(joinpath(artifact_dir, region), joinpath(tz_source_dir, region), force=true)
```
omus, line 64:
```suggestion
                cp(joinpath(artifact_dir, region), tz_source_dir, force=true)
```

src/tzdata/version.jl
```diff
@@ -112,3 +112,16 @@ function active_archive()
...
     return archive
 end
+
+if VERSION >= v"1.4"
+    function active_dir()
+        version = active_version()
+        archive = @artifact_str "tzdata_$version"
+        !isdir(archive) && error("Missing $version tzdata archive")
+        return archive
+    end
```
omus, line 122: Is this only used for testing?

test/runtests.jl
```diff
@@ -44,7 +46,7 @@ include("helpers.jl")
     include("utils.jl")
     include("class.jl")
     include(joinpath("tzdata", "timeoffset.jl"))
-    include(joinpath("tzdata", "archive.jl"))
+    VERSION < v"1.4" && include(joinpath("tzdata", "archive.jl"))
```
omus, line 49: I realize this isn't needed on 1.4 but is there a reason no to include it?

test/tzdata/download.jl
```diff
@@ -1,23 +1,43 @@
...
+if VERSION >= v"1.4"
+    using Pkg.Artifacts
+end
+
 
 @test tzdata_url("2016j") == "https://data.iana.org/time-zones/releases/tzdata2016j.tar.gz"
 @test tzdata_url("latest") == "https://data.iana.org/time-zones/tzdata-latest.tar.gz"
 
 # Note: Try to keep the number of `tzdata_download` calls low to avoid unnecessary network traffic
 mktempdir() do temp_dir
```
omus, line 13: I think these tests should just be for the non-artifact version

test/tzdata/version.jl
```diff
@@ -1,6 +1,12 @@
 using TimeZones.TZData: ARCHIVE_DIR, TZDATA_VERSION_REGEX, TZDATA_NEWS_REGEX
 using TimeZones.TZData: read_news, extract, tzdata_version_dir, tzdata_version_archive
-using TimeZones.TZData: active_version, active_archive
+using TimeZones.TZData: active_version, active_archive, active_dir
+if VERSION >= v"1.4"
+    using Pkg.Artifacts
+    using Pkg.Artifacts: artifacts_dirs
+else
+    import TimeZones.TZData: @artifact_str
```
omus, line 8: Unused

test/tzdata/version.jl
```diff
@@ -29,11 +35,20 @@ end
 @test match(TZDATA_NEWS_REGEX, "Release 19999") === nothing
 
 
-archive = joinpath(ARCHIVE_DIR, "tzdata$TZDATA_VERSION.tar.gz")
+if VERSION >= v"1.4"
+    archive = @artifact_str "tzdata_$TZDATA_VERSION"
+else
+    archive = joinpath(ARCHIVE_DIR, "tzdata$TZDATA_VERSION.tar.gz")
+end
```
omus, line 42:
```suggestion
use_artifacts = VERSION >= v"1.4"

if use_artifacts
    artifact_dir = @artifact_str "tzdata_$TZDATA_VERSION"
else
    archive = joinpath(ARCHIVE_DIR, "tzdata$TZDATA_VERSION.tar.gz")
end
```

test/tzdata/version.jl
```diff
@@ -29,11 +35,20 @@ end
...
 
 mktempdir() do temp_dir
     # Read the first tzdata version
-    extract(archive, temp_dir, "NEWS")
+    if VERSION >= v"1.4"
+        temp_dir = archive  # so I don't need to have too many ifs
+    else
+        extract(archive, temp_dir, "NEWS")
+    end
```
omus, line 50:
```suggestion
    if use_artifacts
        cp(joinpath(artifact_dir, "NEWS"), temp_dir)
    else
        extract(archive, temp_dir, "NEWS")
    end
```

test/tzdata/version.jl
```diff
@@ -51,16 +66,27 @@ mktempdir() do temp_dir
...
 
-@test tzdata_version_archive(archive) == TZDATA_VERSION
-@test_throws ProcessFailedException tzdata_version_archive(@__FILE__) == TZDATA_VERSION
+if VERSION >= v"1.4"
+    @test tzdata_version_dir(archive) == TZDATA_VERSION
+    @test_throws Base.IOError tzdata_version_dir(@__FILE__) == TZDATA_VERSION
+else
+    @test tzdata_version_archive(archive) == TZDATA_VERSION
+    @test_throws ProcessFailedException tzdata_version_archive(@__FILE__) == TZDATA_VERSION
+end
```
omus, line 75:
```suggestion
if !use_artifacts
    @test tzdata_version_archive(archive) == TZDATA_VERSION
    @test_throws ProcessFailedException tzdata_version_archive(@__FILE__) == TZDATA_VERSION
end
```

test/tzdata/version.jl
```diff
@@ -51,16 +66,27 @@ mktempdir() do temp_dir
...
+if VERSION >= v"1.4"
+    archive = active_dir()
+    @test isdir(archive)
+    @test dirname(archive) ∈ artifacts_dirs()
+else
+    archive = active_archive()
+    @test isfile(archive)
+    @test dirname(archive) == ARCHIVE_DIR
+    @test basename(archive) == "tzdata$version.tar.gz"
+end
```
omus, line 92:
```suggestion
if !use_artifacts
    archive = active_archive()
    @test isfile(archive)
    @test dirname(archive) == ARCHIVE_DIR
    @test basename(archive) == "tzdata$version.tar.gz"
end
```

src/tzdata/TZData.jl
racinmat: Oh, that's what I was looking for, thanks! I thought something like that would be here but I could not find it.

src/tzdata/version.jl
racinmat: Yes. Should I delete it then?
I had it as Pkg friendly version of `active_archive`, which is also used only for testing.

test/tzdata/download.jl
racinmat: Makes sense.

test/tzdata/version.jl
racinmat: I just realied, if we'll use the `use_artifacts = VERSION >= v"1.4"` will this go well along with `@static` or not? Because if not, then the `@artifact_str` will be missing again.

src/tzdata/TZData.jl
racinmat: But we need to do this also in tests, thus not using `use_artifacts = VERSION >= v"1.4"`, right?

src/tzdata/build.jl
racinmat: Could you elaborate more on why to do this change? With this change, the tests were crashing, and the package does not build.
I'm returning it back because it fails tests: https://travis-ci.com/github/JuliaTime/TimeZones.jl/jobs/361052429

test/runtests.jl
racinmat: The `archive.jl` is testing the contents of `ARCHIVE_DIR` and whether it stored archives. When I run it on 1.4, which is not storing archives here, because they are stored in already unpacked form (Pkg.Artifacts unpacks the archivs automatically) in artifacts directory, this tests is failing, see: https://travis-ci.com/github/JuliaTime/TimeZones.jl/jobs/361057661

test/tzdata/version.jl
racinmat: I needed to copy into NEWS directory explicitly, otherwise tests were failing, e.g. https://travis-ci.com/github/JuliaTime/TimeZones.jl/jobs/361061783

src/tzdata/build.jl
racinmat: You're right, I missed that. I guess in that case it should download the newest one, right?

racinmat:
I incorporated some changes, and made some changes to the proposed solutions where it was needed to make it work, it's a bit cleaner now, though, thanks for the tips.

src/tzdata/build.jl
racinmat: So I'm directly downloading the newest one if it's still latest.

.travis.yml
```diff
@@ -30,17 +30,17 @@ os:
   - windows
 julia:
   - 1.0  # LTS
-  - 1.4  # Latest release
+  - 1.3  # latest release with old way of downloading artifacts
+  - 1.4  # Latest stable release
+  - 1.5  # Latest release
```
omus, line 35: Let's just make this dynamic. I don't think we need to test on both 1.4 and 1.5 as there isn't any special logic here that differs between 1.4 and 1.5
```suggestion
  - 1    # Latest release
```

.travis.yml
```diff
@@ -53,7 +53,7 @@ jobs:
   include:
     - <<: *doc_test
       os: linux
-      julia: 1.0
+      julia: 1
```
omus, line 56: I'd like to ensure that the documentation tests still succeed against the earliest version of Julia supported by this package.
```suggestion
      julia: 1.0
```

.travis.yml
```diff
@@ -30,17 +30,17 @@ os:
...
   - nightly
 arch:
   - x64
   - x86
 jobs:
   fast_finish: true
-
   # allow_failures:
   #   - julia: nightly
-  #     os: osx
```
omus, line 44: Can you revert this change?

dev/generate_artifacts.jl
```diff
@@ -0,0 +1,76 @@
...
+        content_sha = open(tzdata_download(version)) do f
+           bytes2hex(sha256(f))
+        end
+
+        # Now bind that hash within our `Artifacts.toml`.  `force = true` means that if it already exists,
+        # just overwrite with the new content-hash.  Unless the source files change, we do not expect
+        # the content hash to change, so this should not cause unnecessary version control churn.
+        download_data = [(tzdata_url(version), content_sha)]
+        #bind_artifact!(artifacts_toml, "tzdata_$version", tzfile_hash, lazy=true, download_info=download_data, platform=platform_key_abi())
+        bind_artifact!(artifacts_toml, "tzdata_$version", tzfile_hash, lazy=true, download_info=download_data)
```
omus, line 43: Let's use `tzdata$version` as the artifact name as this is more inline with archive name
```suggestion
        bind_artifact!(artifacts_toml, "tzdata$version", tzfile_hash, lazy=true, download_info=download_data)
```

dev/generate_artifacts.jl
```diff
@@ -0,0 +1,76 @@
...
+    # just overwrite with the new content-hash.  Unless the source files change, we do not expect
+    # the content hash to change, so this should not cause unnecessary version control churn.
+    download_data = [(win_xml_archive_url, content_sha)]
+    bind_artifact!(artifacts_toml, "tzdata_windowsZones", tzfile_hash, lazy=true, download_info=download_data)
+end
+
+# using Pkg.BinaryPlatforms
+# platform_key_abi()
+# typeof(platform_key_abi())
+# typeof(platform_key_abi())
```
omus, line 76:
```suggestion
```

dev/generate_artifacts.jl
```diff
@@ -0,0 +1,76 @@
+using Pkg.Artifacts
+using TimeZones.TZData: tzdata_url
+#using TimeZones.WindowsTimeZoneIDs: WINDOWS_ZONE_URL, WINDOWS_XML_FILE
```
omus, line 3:
```suggestion
```

dev/generate_artifacts.jl
```diff
@@ -0,0 +1,76 @@
...
+using SHA
+using Pkg.BinaryPlatforms
```
omus, line 5:
```suggestion
```

src/tzdata/build.jl
```diff
@@ -38,25 +38,57 @@ function build(
...
+                version = tzdata_version_dir(artifact_dir)
+                set_latest(version, now_utc)
+            else
+                archive = tzdata_download(version, archive_dir)
+                if !isempty(tz_source_dir)
+                    @info "Extracting $version tzdata archive"
+                    extract(archive, tz_source_dir, setdiff(regions, CUSTOM_REGIONS), verbose=verbose)
+                end
             end
         end
```
omus, line 59: I'll just point out that if `version == "latest"` then the regex will never match ("latest" will never be matched by the regex) which results in the `m !== nothing` check never actually getting called. Same problem as pointed out in https://github.com/JuliaTime/TimeZones.jl/pull/270/files#r454385743.

I wouldn't worry about this as I can fix it in a follow up PR.

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -2,11 +2,19 @@ module WindowsTimeZoneIDs
...
 using ...TimeZones: DEPS_DIR
 using EzXML
+if VERSION >= v"1.4"
+    using Pkg.Artifacts
+else
+    # to avoid fail when Julia tried to precompile even non-compatible code
+    macro artifact_str(name)
+        :(throw("this should never bee called"))
+    end
+end
```
omus, line 12:
```suggestion
if VERSION >= v"1.4"
    using Pkg.Artifacts
end
```

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -2,11 +2,19 @@ module WindowsTimeZoneIDs
...
 
 # A mapping of Windows timezone names to Olson timezone names.
 # Details on the contents of this file can be found at:
 # http://cldr.unicode.org/development/development-process/design-proposals/extended-windows-olson-zid-mapping
-const WINDOWS_ZONE_URL = "http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml"
+const WINDOWS_ZONE_URL = "https://raw.githubusercontent.com/unicode-org/cldr/release-37/common/supplemental/windowsZones.xml"
```
omus, line 17: Setting this to a fixed release is better but it doesn't align with the reported download message in the code. I'd say leave as "latest" for now
```suggestion
const WINDOWS_ZONE_URL = "https://raw.githubusercontent.com/unicode-org/cldr/latest/common/supplemental/windowsZones.xml"
```

test/runtests.jl
omus: Ok, I get why you're disabling these tests now.

test/tzdata/version.jl
```diff
@@ -51,16 +63,22 @@ mktempdir() do temp_dir
     @test_throws ErrorException tzdata_version_dir(dirname(@__FILE__))
 end
 
-@test tzdata_version_archive(archive) == TZDATA_VERSION
-@test_throws ProcessFailedException tzdata_version_archive(@__FILE__) == TZDATA_VERSION
+use_artifacts = VERSION >= v"1.4"
```
omus, line 66: Would make more sense to put this earlier in the file

test/tzdata/version.jl
omus: `@static` at the top-level I believe isn't necessary so it can be safely left out. Also note that `@static` can be applied to any `if` with a condition that can be resolved at compile time

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -2,11 +2,19 @@ module WindowsTimeZoneIDs
...
+    macro artifact_str(name)
+        :(throw("this should never bee called"))
+    end
+end
 
 # A mapping of Windows timezone names to Olson timezone names.
 # Details on the contents of this file can be found at:
 # http://cldr.unicode.org/development/development-process/design-proposals/extended-windows-olson-zid-mapping
-const WINDOWS_ZONE_URL = "http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml"
+const WINDOWS_ZONE_URL = "https://raw.githubusercontent.com/unicode-org/cldr/latest/common/supplemental/windowsZones.xml"
```
omus, line 17: My bad, this should be master and not latest
```suggestion
const WINDOWS_ZONE_URL = "https://raw.githubusercontent.com/unicode-org/cldr/master/common/supplemental/windowsZones.xml"
```

omus: This change should also resolve the conflict

.travis.yml
racinmat: yes, np

.travis.yml
racinmat: Well, Julia 1.5 uses Package server by default, and Julia 1.4 not, so this allows us to see potential problems with Package Server related to the Artifact, but also in general.
In Julia 1.4 Package server is not used by default.
Testing on both of them already helped me discover some weird behavior of Package Server.
But we can leave it only on `1`.

test/tzdata/version.jl
racinmat: Sure.

src/tzdata/build.jl
racinmat: Ok, I see, thanks for pointing it out.

racinmat:
Ok, I made the changes you pointed out, hope it's ok now.

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -46,8 +49,15 @@ function build(xml_file::AbstractString=WINDOWS_XML_FILE; force::Bool=false)
         if isfile(fallback_xml_file) && !force
             cp(fallback_xml_file, xml_file)
         else
-            @info "Downloading latest Windows to POSIX timezone ID XML"
-            download(WINDOWS_ZONE_URL, xml_file)
+            if VERSION >= v"1.4"
```
omus, line 52: Causing CI tests to fail on earlier versions of Julia
```suggestion
            @static if VERSION >= v"1.4"
```

Project.toml
```diff
@@ -1,12 +1,13 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.2.0"
+version = "1.3.0"
```
omus, line 4: You can leave this out of this PR. I'll bump the version number right before I make the release
```suggestion
version = "1.2.0"
```

racinmat: Ok, in some repos they like if PR also bumps the version.

racinmat:
Ok, commited these changes and now it passes.

omus:
GitHub has been experiencing some issues and it appears the last commit didn't actually run CI. I'll attempt to kick start it by opening/closing the PR.

test/tzdata/download.jl
```diff
@@ -1,4 +1,7 @@
-using TimeZones.TZData: tzdata_url, tzdata_download, isarchive, LATEST_FILE, read_latest
+using TimeZones.TZData: tzdata_url, tzdata_download, isarchive, LATEST_FILE, read_latest, REGIONS, latest_version,
+    tzdata_version
+using Dates
+
```
omus, line 4:
```suggestion
using TimeZones.TZData: tzdata_url, tzdata_download, isarchive, LATEST_FILE, read_latest
```

test/tzdata/version.jl
```diff
@@ -29,11 +34,20 @@ end
 @test match(TZDATA_NEWS_REGEX, "Release 19999") === nothing
 
 
-archive = joinpath(ARCHIVE_DIR, "tzdata$TZDATA_VERSION.tar.gz")
+@static if VERSION >= v"1.4"
```
omus, line 37:
```suggestion
@static if use_artifacts
```

test/tzdata/version.jl
```diff
@@ -29,11 +34,20 @@ end
...
+    artifact_dir = @artifact_str "tzdata$TZDATA_VERSION"
+else
+    archive = joinpath(ARCHIVE_DIR, "tzdata$TZDATA_VERSION.tar.gz")
+end
 
 mktempdir() do temp_dir
     # Read the first tzdata version
-    extract(archive, temp_dir, "NEWS")
+    @static if VERSION >= v"1.4"
```
omus, line 45:
```suggestion
    if use_artifacts
```

PR #276, opened 2020-07-15
Artifacts follow up

omus:
Follow up to #270. Changes include:

- Renamed `tzdatawindowsZones` to `unicode-cldr-release-37` to be more descriptive and include the version number
- Added a Project.toml and Manifest.toml for `dev/generate_artifacts.jl` script
- Revised `dev/generate_artifacts.jl` to no longer populate `~/.julia/artifacts` when generating the `Artifacts.toml`. The script also now reports the latest revisions to assist with maintenance.
- Introduced a patched `@artifact_str` for Julia 1.3 to allow artifacts to be used on Julia 1.3
- When using `JULIA_TZ_VERSION="latest"` an error is now raised if IANA has a version newer than what is recorded in the `Artifacts.toml`. Ideally, this will cause users who noticed this to update the `Artifacts.toml` and submit a PR.


dev/generate_artifacts.jl
```diff
@@ -1,69 +1,83 @@
...
-    # Query the `Artifacts.toml` file for the hash bound to the specific version
-    # (returns `nothing` if no such binding exists)
-    tzarchive_latest_hash = artifact_hash("tzdata$version", artifacts_toml)
+        archive = download(url)
+        try
+            # Compute the `sha256` of the archive.
+            archive_sha = bytes2hex(open(sha256, archive))
 
-    # If the name was not bound, or the hash it was bound to does not exist, create it!
-    if isnothing(tzarchive_latest_hash) || !artifact_exists(tzarchive_latest_hash)
```
nicoleepp, line 23: Did we stop dealing with the case where the artifact hash is not nothing but it doesn't exist?

omus: `artifact_exists` checks that an artifact is present in `~/.julia/artifacts`. Part of the changes done here were to avoid populating this `~/.julia/artifacts` unnecessarily when generating the `Artifacts.toml` I needed to remove the `artifact_exists` check and only check if the artifact was listed in the TOML file which is done via `artifact_hash(...) !== nothing`

dev/Manifest.toml
```diff
@@ -0,0 +1,150 @@
+# This file is machine-generated - editing it directly is not advised
```
mattBrzezinski, line 1: I'm not familiar with Artifacts, is this `Manifest` intended to be committed?

dev/Project.toml
```diff
@@ -0,0 +1,9 @@
+[deps]
+HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
+JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
+TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
+
+[compat]
+JSON = "0.20.1, 0.21"
+HTTP = "0.8.1"
+julia = "1.3"
```
mattBrzezinski, line 9:
```suggestion
julia = "1.3"

```

omus, line 9: I'll correct this

omus: I updated the commit where this was original made

dev/Manifest.toml
omus: This `Manifest.toml` is committed as when executing this script I want it to use the currently checked out version of TimeZones.jl. You can specify a relative path like that in a `Manifest.toml` file but not in the `Project.toml`. This is same reason we have a manifest in the docs/ dir.

omus: This is specific for the `dev/generate_artifacts.jl` code but not specific to artifacts themselves

mattBrzezinski: Ah gotcha, yeah I figured that it was similar to `docs`, but wasn't sure! Thanks for clearing that up.

Issue #282, opened 2020-09-04
MicroOptimizations for parsing


oxinabox:
I was looking though https://github.com/JuliaTime/TimeZones.jl/blob/cf5336158dc93f2427f9970d447763d9b09f0abc/src/parse.jl
to see if there were any optimizations left on the table.
There seem to be a few, but nothing super huge.
the `==` -> `===` while the smallest is probably the most common so might actually add up.

## What is the fastest way to do short-string equality?

```julia
julia> const ktrue = (SubString("GMT", 1:3));

julia> const kfalse3 = (SubString("ABC", 1:3));

julia> const kfalse4 = (SubString("ABCD", 1:4));


#########################

julia> is_UTC_or_GMT(name) = @inbounds length(name) === 3 && (
           (name[1]==='U' && name[2]==='T' && name[3]==='C') ||
           (name[1]==='G' && name[2]==='M' && name[3]==='T')
       );

julia> @btime (is_UTC_or_GMT)($ktrue);
  22.290 ns (0 allocations: 0 bytes)

julia> @btime (is_UTC_or_GMT)($kfalse3);
  15.152 ns (0 allocations: 0 bytes)

julia> @btime (is_UTC_or_GMT)($kfalse4);
  8.694 ns (0 allocations: 0 bytes)


#########################

julia> is_UTC_or_GMT_or(name) = (name == "UTC" || name == "GMT");

julia> @btime (is_UTC_or_GMT_or)($ktrue);
  46.549 ns (4 allocations: 224 bytes)

julia> @btime (is_UTC_or_GMT_or)($kfalse3);
  39.877 ns (4 allocations: 224 bytes)

julia> @btime (is_UTC_or_GMT_or)($kfalse4);
  39.992 ns (4 allocations: 224 bytes)

#########################

julia> is_UTC_or_GMT_in(name) = (name in  ("UTC", "GMT"));

julia> @btime (is_UTC_or_GMT_in)($ktrue);
  47.364 ns (4 allocations: 224 bytes)

julia> @btime (is_UTC_or_GMT_in)($kfalse3);
  41.084 ns (4 allocations: 224 bytes)

julia> @btime (is_UTC_or_GMT_in)($kfalse4);
  40.892 ns (4 allocations: 224 bytes)

```

Answer: unroll it into per-character equality checks.

### Does `===` vs `==` make a difference for character literals?

```julia
[...]

PR #284, opened 2020-09-04
use  in(char, string) rather than occursin(string,string)

oxinabox:
one of the things from
https://github.com/JuliaTime/TimeZones.jl/issues/282

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -57,7 +57,7 @@ function tryparsenext_tz(str, i, len, min_width::Int=1, max_width::Int=0)
         # If the time zone is recognized make sure that it is well-defined. For our
         # purposes we'll treat all abbreviations except for UTC and GMT as ambiguous.
         # e.g. "MST": "Mountain Standard Time" (UTC-7) or "Moscow Summer Time" (UTC+3:31).
-        if occursin("/", name) || name in ("UTC", "GMT")
+        if in('/', name) || name in ("UTC", "GMT")
             return name, i
         else
             return nothing
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -57,7 +57,7 @@ function tryparsenext_tz(str, i, len, min_width::Int=1, max_width::Int=0)
         # If the time zone is recognized make sure that it is well-defined. For our
         # purposes we'll treat all abbreviations except for UTC and GMT as ambiguous.
         # e.g. "MST": "Mountain Standard Time" (UTC-7) or "Moscow Summer Time" (UTC+3:31).
-        if in('/', name) || name in ("UTC", "GMT")
+        if '/' in name || name in ("UTC", "GMT")
             return name, i
         else
             return nothing


src/parse.jl
```diff
@@ -57,7 +57,7 @@ function tryparsenext_tz(str, i, len, min_width::Int=1, max_width::Int=0)
         # If the time zone is recognized make sure that it is well-defined. For our
         # purposes we'll treat all abbreviations except for UTC and GMT as ambiguous.
         # e.g. "MST": "Mountain Standard Time" (UTC-7) or "Moscow Summer Time" (UTC+3:31).
-        if occursin("/", name) || name in ("UTC", "GMT")
+        if in('/', name) || name in ("UTC", "GMT")
```
omus, line 60: May as well be consistent with the next condition
```suggestion
        if '/' in name || name in ("UTC", "GMT")
```

omus (commented):
Would be good to get before/after benchmarks posted

oxinabox:
compare https://github.com/JuliaTime/TimeZones.jl/issues/282#issuecomment-687169225
to

this PR:
```julia
julia> @btime parse(ZonedDateTime, "2017-11-14 11:03:53 +0100", dateformat"yyyy-mm-dd HH:MM:SS zzzzz");
  744.915 ns (13 allocations: 592 bytes)

julia> @btime parse(ZonedDateTime, "2016-04-11 08:00 UTC", dateformat"yyyy-mm-dd HH:MM ZZZ");
  660.746 ns (15 allocations: 704 bytes)

julia> @btime parse(ZonedDateTime, "2000+00", dateformat"yyyyz");
  464.526 ns (13 allocations: 496 bytes)

julia> @btime parse(ZonedDateTime, Test.GenericString("2018-01-01 00:00 UTC"), dateformat"yyyy-mm-dd HH:MM ZZZ");
  9.108 μs (136 allocations: 4.44 KiB)

julia> @btime tryparse(ZonedDateTime, "2013-03-20 11:00:00+04:00", dateformat"y-m-d H:M:SSz");
  888.667 ns (14 allocations: 640 bytes)

julia> @btime tryparse(ZonedDateTime, "2016-04-11 08:00 EST", dateformat"yyyy-mm-dd HH:MM zzz");
  90.784 ns (0 allocations: 0 bytes)

julia> @btime ZonedDateTime("2000-01-02T03:04:05.006+0700");
  830.208 ns (14 allocations: 688 bytes)

julia> @btime ZonedDateTime("2000-01-02T03:04:05.006Z");
  629.645 ns (9 allocations: 416 bytes)

julia> @btime ZonedDateTime("2018-11-01-0600", dateformat"yyyy-mm-ddzzzz");
  612.932 ns (13 allocations: 544 bytes)
```

The important on is:
```
julia> @btime parse(ZonedDateTime, Test.GenericString("2018-01-01 00:00 UTC"), dateformat"yyyy-mm-dd HH:MM ZZZ");
  9.108 μs (136 allocations: 4.44 KiB)
```
vs master:
```julia
julia> @btime parse(ZonedDateTime, Test.GenericString("2018-01-01 00:00 UTC"), dateformat"yyyy-mm-dd HH:MM ZZZ");
  9.600 μs (138 allocations: 4.48 KiB)
```

omus:
I'll just note the first test here is slower than what occurred on `master` by 39 ns. Probably just noise but I wanted to point out that it was noticed.

oxinabox:
😂  🦅 👁️ 
I also noticed but figured it was noise

omus:
Looks like mainly we're seeing the performance bump from using a `Char` over a `String`:
```julia
julia> using Random, BenchmarkTools

julia> randstring(501) * '/' * randstring(498);

julia> str = randstring(501) * '/' * randstring(498);

julia> length(str)
1000

julia> @btime occursin("/", $str)
  30.878 ns (0 allocations: 0 bytes)
true

julia> @btime in('/', $str)
  20.098 ns (0 allocations: 0 bytes)
true

julia> @btime occursin('/', $str)
  21.028 ns (0 allocations: 0 bytes)
true
```
I'm still good with this.

PR #285, opened 2020-09-04
using || rather than in to check short lists of characters

oxinabox:
one of the things from
https://github.com/JuliaTime/TimeZones.jl/issues/282

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -136,7 +136,7 @@ function _parsesub_tzabbr(
 
         if state == :simple && isletter(c)
             name_end = i
-        elseif state == :expanded && (isletter(c) || isdigit(c) || c in ('+', '-'))
+        elseif state == :expanded && (isletter(c) || isdigit(c) || c === '+' || c === '-'
             name_end = i
         elseif state == :started && c === '<'
             name_start = ii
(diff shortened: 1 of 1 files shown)


src/parse.jl
```diff
@@ -136,7 +136,7 @@ function _parsesub_tzabbr(
 
         if state == :simple && isletter(c)
             name_end = i
-        elseif state == :expanded && (isletter(c) || isdigit(c) || c in ('+', '-'))
+        elseif state == :expanded && (isletter(c) || isdigit(c) || c === '+' || c === '-'
```
oxinabox, line 139:
```suggestion
        elseif state == :expanded && (isletter(c) || isdigit(c) || c === '+' || c === '-')
```

omus (commented):
Looks good again. Please post before/after benchmarks

omus:
It definitely does help but it's already so small it probably won't be noticed. We can still make this change.
```julia
julia> c = '0'
'0': ASCII/Unicode U+0030 (category Nd: Number, decimal digit)

julia> @btime $c in ('+', '-')
  1.983 ns (0 allocations: 0 bytes)
false

julia> @btime $c === '+' || $c === '-'
  0.033 ns (0 allocations: 0 bytes)
false
```

PR #290, opened 2020-09-11
Apply `@artifact_str` fix to Julia 1.3.2-pre.0

omus:
Allows people who build from source on the `release-1.3` branch to be able to use TimeZones.jl. I also adjusted the lower bound which now includes the first version with the bad `@artifact_str` version (Julia commit 5e0653407f).

I'm still assuming a release of 1.3.2 would have a fixed Pkg (most likely that will never occur).

### DIFF
--- a/src/compat.jl
+++ b/src/compat.jl
@@ -10,7 +10,7 @@ if v"1.3.0" <= VERSION < v"1.6.0-DEV.84"
     end
 end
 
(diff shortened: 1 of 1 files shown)


omus:
Tested locally with the commit before 5e0653407f and `v"1.3.2-pre.0"` and both build successfully

PR #294, opened 2020-09-23
Test occurrence when set to non-positive

omus:
Adds test for current behaviour. I'm not sure I like this behaviour but at least there are tests now.


omus:
Expected macOS Julia nightly failure.

PR #298, opened 2020-09-28
Replace `DateTime(zdt, Local)` with `DateTime(zdt)`

omus:
It was brought up that there is some confusion around `Local` (meaning local to the time zone provided) and `localzone()` (meaning the user's system time zone). When we originally [switched from using `localtime`/`utc`](https://github.com/JuliaTime/TimeZones.jl/pull/231) it appears the intent was to be explicit. As it seems that most people want to drop the time zone info we'll change things such that `DateTime(zdt)` means drop the time zone and `DateTime(zdt, UTC)` provides the `DateTime` implicitly in `UTC`.

### DIFF
--- a/docs/src/api-public.md
+++ b/docs/src/api-public.md
@@ -34,9 +34,12 @@ ZonedDateTime(::DateTime, ::VariableTimeZone, ::Integer)
 ZonedDateTime(::DateTime, ::VariableTimeZone, ::Bool)
 astimezone
 TimeZones.timezone(::ZonedDateTime)
-DateTime(::ZonedDateTime, ::Union{Type{Local}, Type{UTC}})
-Date(::ZonedDateTime, ::Union{Type{Local}, Type{UTC}})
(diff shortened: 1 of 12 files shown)


nicoleepp:
Is the plan to make this a breaking release? If so, needs a version bump?

omus:
> Is the plan to make this a breaking release? If so, needs a version bump?

The change is not breaking and will just be a feature bump. TimeZones.jl does version bumps independent from the PRs so no version change is required here.

nicoleepp (approved):

omus:
I will say the one thing I don't like about this change is that it seems impossible to write a single docstring for both methods as we no longer have the symmetry of `Date(x, Local)`/`Date(x, UTC)`. These are very minor things and I think someone complains about `Date(x, Local)` being dropped we can remove the deprecation and just have `Date(x, Local)` and `Date(x)` co-exist.

Issue #309, opened 2020-11-23
serialization and julia versions


gdkrmr:
I have installed `TimeZones.jl` with the latest Julia master and then installed it again with Julia 1.5.3 and got the following error in Julia 1.5.3:

```julia
ERROR: LoadError: Cannot read stream serialized with a newer version of Julia.
Got data version 13 > current version 10
```

I had to rebuild the package in order to make it work again. I don't think this is ideal, maybe you should not use the Julia internal serialization but a different format.

nickrobinson251:
This is a duplicate: it's tracked at https://github.com/JuliaTime/TimeZones.jl/issues/288

(and yeah, this issue is annoying...)

omus:
Should be fixed in TimeZones 1.5.3

PR #295, opened 2020-09-24
WIP: Compile time zone information during package

omus:
Fixes #288. Originally TimeZones.jl serialized the compiled time zone information to reduce the time it took to load the package. Julia has had many improvements since then and now it seems feasible to compile the time zone information during package initialization

PR #315, opened 2020-12-04
When Compiling serialize seperately per julia version

### LINKED ISSUE #288: If TimeZone is installed and built on a newer version of Julia that TimeZone version is unusable on older Julia versions
```
julia> using TimeZones # 1.5.1

julia> tz"UTC"
ERROR: LoadError: Cannot read stream serialized with a newer version of Julia.
Got data version 12 > current version 10
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] readheader(::Serialization.Serializer{IOStream}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:715
 [3] handle_deserialize(::Serialization.Serializer{IOStream}, ::Int32) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:878
 [4] deserialize(::Serialization.Serializer{IOStream}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:773
 [5] deserialize at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:760 [inlined]
 [6] open(::typeof(Serialization.deserialize), ::String, ::Vararg{String,N} where N; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at ./io.jl:325
 [7] open at ./io.jl:323 [inlined]
 [8] (::TimeZones.var"#3#4"{String})() at /Users/kristoffercarlsson/.julia/packages/TimeZones/v0mfN/src/types/timezone.jl:50
 [9] get!(::TimeZones.var"#3#4"{String}, ::Dict{String,Tuple{TimeZone,TimeZones.Class}}, ::String) at ./dict.jl:450
 [10] TimeZone(::String, ::TimeZones.Class) at /Users/kristoffercarlsson/.julia/packages/TimeZones/v0mfN/src/types/timezone.jl:46 (repeats 2 times)
 [11] @tz_str(::LineNumberNode, ::Module, ::Any) at /Users/kristoffercarlsson/.julia/packages/TimeZones/v0mfN/src/types/timezone.jl:86
in expression starting at REPL[2]:1
```

Because I happened to install TimeZones on julia master first.

oxinabox:
Fixes #288 (and its duplicate #309).

I have tested locally that this works to fix that problem.
It's hard to test it in CI since it requires starting multiple julia instances.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.5.3"
+version = "1.5.2"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.5.2"
+version = "1.5.3"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/tzdata/TZData.jl
+++ b/src/tzdata/TZData.jl
@@ -16,7 +16,7 @@ end
 
 const ARCHIVE_DIR = joinpath(DEPS_DIR, "tzarchive")
 const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
-const COMPILED_DIR = joinpath(DEPS_DIR, string("compiled_julia_v", VERSION))
+const COMPILED_DIR = joinpath(DEPS_DIR, "compiled", string(VERSION))
 
 const ARTIFACT_TOML = joinpath(@__DIR__, "..", "..", "Artifacts.toml")
 
--- a/src/tzdata/TZData.jl
+++ b/src/tzdata/TZData.jl
@@ -16,7 +16,7 @@ end
 
 const ARCHIVE_DIR = joinpath(DEPS_DIR, "tzarchive")
 const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
-const COMPILED_DIR = joinpath(DEPS_DIR, "compiled")
+const COMPILED_DIR = joinpath(DEPS_DIR, string("compiled_julia_v", VERSION))
 
 const ARTIFACT_TOML = joinpath(@__DIR__, "..", "..", "Artifacts.toml")
 
--- a/src/tzdata/build.jl
+++ b/src/tzdata/build.jl
@@ -102,9 +102,9 @@ function build(
 end
 
 function build(version::AbstractString=tzdata_version())
-    isdir(ARCHIVE_DIR) || mkdir(ARCHIVE_DIR)
-    isdir(TZ_SOURCE_DIR) || mkdir(TZ_SOURCE_DIR)
-    isdir(COMPILED_DIR) || mkdir(COMPILED_DIR)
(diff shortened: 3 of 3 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.3.2"
+version = "1.3.3"
```
oxinabox, line 4: I say this is a bug fix.
This always should have worked but it.

oxinabox:
oops looks like i fixed it on an old version, rebasing and retest now

oxinabox:
confirm still fixed

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.5.2"
+version = "1.5.3"
```
oxinabox, line 4: I say this is a bug-fix. It always should have worked

fchorney (approved):
LGTM

Project.toml
fchorney: I think that makes sense

iamed2:
Can we make this serialize based on major.minor? Or can serialization change in patch versions?

oxinabox:
It probably can change based on patch versions if there is bug in how serialisation is done that needs changed to the format to fix, which is then backported.
More importantly if you are using Julia nightly/source build it can definitely change based on the pre-release build suffix.

I feel like keeping things simple and just using the whole VERSION is better than trying to be clever.
It makes it marginally slower to add and uses a tiny bit more harddrive.

src/tzdata/TZData.jl
```diff
@@ -16,7 +16,7 @@ end
 
 const ARCHIVE_DIR = joinpath(DEPS_DIR, "tzarchive")
 const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
-const COMPILED_DIR = joinpath(DEPS_DIR, "compiled")
+const COMPILED_DIR = joinpath(DEPS_DIR, string("compiled_julia_v", VERSION))
```
omus, line 19: I'd rather use separate sub-directories:
```suggestion
const COMPILED_DIR = joinpath(DEPS_DIR, "compiled", string(VERSION))
```
This requires that we change [this line](https://github.com/JuliaTime/TimeZones.jl/blob/01d025ae6420251e3e075f2dd1eae3f7a9b2521e/src/tzdata/build.jl#L107) to use `mkpath`

omus (changes requested):
Eventually we'll probably change the approach here to use scratch spaces: https://github.com/JuliaTime/TimeZones.jl/pull/295#issuecomment-705302332

src/tzdata/TZData.jl
oxinabox: I considered that but then i would also have to change that other line

omus (approved):
Looks good. Just going to get CI working to ensure everything is working as expected here

omus:
I'm quite sure the benchmark failure is due to the compiled files being in different directories. As a build isn't triggered when running against the old commit then the compiled files are missing according to the base version. As I'm not expecting performance issues from these changes I'll go ahead and merge this.

Project.toml
omus: Definitely is a patch but I like to bump versions separate from PRs for this repo.
```suggestion
version = "1.5.3"
```

PR #287, opened 2020-09-04
POC: Make `ZonedDateTime` an isbits type

omus:
Fixes https://github.com/JuliaTime/TimeZones.jl/issues/271. 

Current master:
```julia
julia> isbitstype(ZonedDateTime)
false

julia> @btime ZonedDateTime(2000, 1, 2, tz"America/Winnipeg")
  72.826 ns (3 allocations: 160 bytes)
2000-01-02T00:00:00-06:00

julia> @btime fill(ZonedDateTime(2000, 1, 2, tz"America/Winnipeg"), 100);
  212.670 ns (4 allocations: 1.03 KiB)
```

I original used a single `Dict` for lookup and the performance was:
```julia
julia> isbitstype(ZonedDateTime)
true

julia> @btime ZonedDateTime(2000, 1, 2, tz"America/Winnipeg")
  721.936 ns (2 allocations: 160 bytes)
2000-01-02T00:00:00-06:00

julia> @btime fill(ZonedDateTime(2000, 1, 2, tz"America/Winnipeg"), 100);
  748.686 ns (3 allocations: 1.92 KiB)
```
The latest version uses a `Vector` for lookup:
```julia
julia> isbitstype(ZonedDateTime)
true

julia> @btime ZonedDateTime(2000, 1, 2, tz"America/Winnipeg")
  177.060 ns (4 allocations: 208 bytes)
2000-01-02T00:00:00-06:00

julia> @btime fill(ZonedDateTime(2000, 1, 2, tz"America/Winnipeg"), 100);
  285.683 ns (5 allocations: 1.97 KiB)
```
I need to run some additional tests yet.

PR #323, opened 2021-04-09
Proof of concept for isbits (PR 2)

oxinabox:
This is a rebased version of  #287 
(and thus also would close #287)

I am experimenting with it again.

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -55,6 +55,7 @@ include("indexable_generator.jl")
 
 include("class.jl")
 include("utcoffset.jl")
+include("external.jl")
 include(joinpath("types", "timezone.jl"))
 include(joinpath("types", "fixedtimezone.jl"))
 include(joinpath("types", "variabletimezone.jl"))
--- a/src/external.jl
+++ b/src/external.jl
@@ -1,0 +1,15 @@
+struct ExternalField{T}
+    table::Dict{T,Int}
+    data::Vector{T}
+end
+
+ExternalField{T}() where T = ExternalField{T}(Dict{T,Int}(), Vector{T}())
+
+function add!(x::ExternalField{T}, val::T) where T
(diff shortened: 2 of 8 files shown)


oxinabox:
When I used this, I get:
```
BoundsError(TimeZone[tz"Z", tz"UTC", TimeZones.FixedTimeZone("LMT", -21036), TimeZones.FixedTimeZone("CST", -21600), TimeZones.FixedTimeZone("CDT", -21600, 3600), tz"EST", TimeZones.FixedTimeZone("CWT", -21600, 3600), TimeZones.FixedTimeZone("CPT", -21600, 3600), tz"America/Chicago"],
(16,)), Any[(getindex at array.jl:809 [inlined], 1),
(getindex at external.jl:15 [inlined], 1),
```

Which suggests that it is not serializing properly.

I recall @JeffBezanson saying something about overloading `serialize` actually doesn't work, because serializing `Vector{T}`  doesn't call `serialize(T)`.

Which seems to be true:

Setup:
```julia
julia> using Serialization, TimeZones

julia> zdt = ZonedDateTime(Date(2011, 6, 1), tz"America/Winnipeg")
2011-06-01T00:00:00-05:00
```

To check if it is serializing properly we expect that the name of the timezone would occur in the serialized stream.

We can see it does for if just serializing a `ZonedDateTime`

```julia
julia> iob = IOBuffer(); serialize(Serializer(iob), zdt); occursin("Winnipeg", String(take!(iob)))
true
```

But that it does not if serializing an array of `ZonedDateTime`s:
```julia
julia> iob = IOBuffer(); serialize(Serializer(iob), [zdt]); occursin("Winnipeg", String(take!(iob)))
false
```

oxinabox:
I think the answer is we should be overloading `write` not `serialize` both `serialize_any` and `serialize_array_values` fall back to `write`

Issue #271, opened 2020-07-08
ZonedDataTime not isbits


Keno:
@oxinabox and I were discussing some performance problems in the presence of large numbers of `ZonedDateTime` objects and I noted that them not being `isbits` is a problem if you put lots of them in a vector, because the GC time taken to mark such vectors will be `O(number of elements)` not `O(number of vectors)`. Now, arguably this is a prime use case for a partially pooled vector that just pools the TimeZone objects, but keeps the raw DateTime inline, but I'm also wondering if `ZonedDateTime` itself couldn't be made to be `isbits` and not have this issue in the first place. In particular, it seems like there's probably only ever gonna be a small handful of TimeZones globally in the system, so we could just globally intern them and only refer to them by small integer identifiers (reconstituting them into the full object on access). Does that sound like a feasible thing to do? I think it would dramatically increase the performance of `ZonedDateTime`.

omus:
I have briefly looked into this and I think it's a feasible thing to do. My main concern is how this will work with serialization specifically if a user had defined a custom time zone instead of a predefined time zone.

oxinabox:
Could we fix that by defining custom serialization, so that it still serializes the same as it does now, even though it just stores a integery identifier both pre-serialization and after deserialization ?

kcajf:
https://github.com/kcajf/TimeZones.jl
I have a branch that was working towards this - i.e. generate code for all the timezones, each one taking up just 16 bits (VariableTimeZoneF). I then made VariableTimeZone parametric in the timezone type. It's a bit hacky, and I never had time to fully finish it, but the proof of concept seemed to work. 
See https://github.com/JuliaTime/TimeZones.jl/issues/182 for more discussion too

omus:
I'll be working on this next week.

kcajf:
@omus what did you think of the idea (discussed in #182) of making ZonedDateTime parametric in the timezone type? Too much change? It would allowz for example, `ZonedDateTime{UTC}` to be only 8 bytes

omus:
Interface-wise I like the idea of a parametric `ZonedDateTime`. Adding the type parameter is definitely breaking which we can do but need to do the experimentation to be sure it's with it. The change in #287 doesn't break the external API so it's a good alternative implementation to compare to.

oxinabox:
I thought worth an update on some more thoughts i have had on using ShortStrings.jl
[...]

PR #324, opened 2021-04-09
Introduce a Shortstring based Name type

oxinabox:
A possible partial resolution to #271 
Doesn't solve the `VariableTimeZone`'s having a `Vector` as afield

<s>
At the moment tests are failing.
That probably is not significiant.
just some last things to chase up
e.g. with show overloads
</s>

### DIFF
--- a/src/types/fixedtimezone.jl
+++ b/src/types/fixedtimezone.jl
@@ -30,7 +30,7 @@ const FIXED_TIME_ZONE_REGEX = r"""
 A `TimeZone` with a constant offset for all of time.
 """
 struct FixedTimeZone <: TimeZone
-    name::String
+    name::Name
     offset::UTCOffset
 end
 
--- a/src/types/fixedtimezone.jl
+++ b/src/types/fixedtimezone.jl
@@ -1,3 +1,7 @@
+# Ideally would always use ShortString15, but it's `hash` is broken on 32-bit systems.
+# https://github.com/JuliaString/MurmurHash3.jl/issues/12
+const FixedTimeZoneName = Int === Int64 ? ShortString15 : String
+
 const FIXED_TIME_ZONE_REGEX = r"""
     ^(?|
         Z
@@ -30,7 +34,7 @@ const FIXED_TIME_ZONE_REGEX = r"""
 A `TimeZone` with a constant offset for all of time.
 """
 struct FixedTimeZone <: TimeZone
-    name::Name
+    name::FixedTimeZoneName
     offset::UTCOffset
 end
 
@@ -72,7 +76,7 @@ UTC+15:45:21
 function FixedTimeZone(s::AbstractString)
     s == "Z" && return UTC_ZERO
 
-    m = match(FIXED_TIME_ZONE_REGEX, s)
+    m = match(FIXED_TIME_ZONE_REGEX, String(s))
     m === nothing && throw(ArgumentError("Unrecognized time zone: $s"))
 
     coefficient = m[:sign] == "-" ? -1 : 1
--- a/src/types/name.jl
+++ b/src/types/name.jl
@@ -16,7 +16,16 @@ end
 
 Base.convert(::Type{String}, name::SName) = string(name)
 function Base.convert(::Type{SName}, str::AbstractString)
+    name = try_convert(SName, str)
+    name isa Nothing && DomainError(str, "All timezone name parts must have length < 16")
+    return name
+end
+
+try_convert(::Type{SName}, name::SName) = name
+try_convert(::Type{String}, name::String) = name
+function try_convert(::Type{SName}, str::AbstractString)
     parts = split(str, "/"; limit=3)
+    all(length(parts) < 16) ||return nothing
     return if length(parts) == 3
         SName(parts[1], parts[2], parts[3])
     elseif length(parts) == 2
@@ -25,6 +34,7 @@ function Base.convert(::Type{SName}, str::AbstractString)
         SName(parts[1], ss15"", ss15"")
     end
 end
+
 
 Base.isempty(name::SName) = isempty(name.region)  # region being empty implies all empty
 
--- a/src/types/name.jl
+++ b/src/types/name.jl
@@ -1,0 +1,46 @@
+struct SName
+    region::ShortString15
+    locality1::ShortString15
+    locality2::ShortString15
+end
+
+function Base.print(io::IO, name::SName)
+    print(io, name.region)
+    if !isempty(name.locality1)
+        print(io,"/", name.locality1)
+        if !isempty(name.locality2)
+            print(io,"/", name.locality2)
+        end
+    end
+end
+
+Base.convert(::Type{String}, name::SName) = string(name)
+function Base.convert(::Type{SName}, str::AbstractString)
+    parts = split(str, "/"; limit=3)
+    return if length(parts) == 3
+        SName(parts[1], parts[2], parts[3])
+    elseif length(parts) == 2
+        SName(parts[1], parts[2], ss15"")
+    else
+        SName(parts[1], ss15"", ss15"")
+    end
+end
+
+Base.isempty(name::SName) = isempty(name.region)  # region being empty implies all empty
+
+name_parts(str::AbstractString) = split(str, "/")
+function name_parts(name::SName)
+    # TODO this could be faster by returning an iterator but not really performance critial
+    parts = [name.region]
+    if !isempty(name.locality1)
+        push!(parts, name.locality1)
+        if !isempty(name.locality2)
+            push!(parts, name.locality2)
+        end
+    end
+    return parts
+end
+
+# Short strings are broken on 32bit:
+# TODO: https://github.com/JuliaString/MurmurHash3.jl/issues/12
+const Name = Int === Int32 ? String : SName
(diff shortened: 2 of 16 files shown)


src/types/name.jl
```diff
@@ -0,0 +1,56 @@
...
+        if !isempty(name.locality2)
+            print(io,"/", name.locality2)
+        end
+    end
+end
+
+Base.convert(::Type{String}, name::SName) = string(name)
+function Base.convert(::Type{SName}, str::AbstractString)
+    name = try_convert(SName, str)
+    name isa Nothing && DomainError(str, "All timezone name parts must have length < 16")
```
oxinabox, line 20:
```suggestion
    name isa Nothing && throw(DomainError(str, "All timezone name parts must have length < 16"))
```

test/interpret.jl
```diff
@@ -125,7 +125,7 @@ long = VariableTimeZone("Test/LongGap", [
 ])
 
 # A time zone with an unnecessary transition that typically is hidden to the user
-hidden = VariableTimeZone("Test/HiddenTransition", [
+hidden = VariableTimeZone("Test/Hidden", [
```
oxinabox, line 128: this one was too long before

src/types/name.jl
```diff
@@ -0,0 +1,56 @@
...
+function try_convert(::Type{SName}, str::AbstractString)
+    parts = split(str, "/"; limit=3)
+    all(length(parts) < 16) ||return nothing
+    return if length(parts) == 3
+        SName(parts[1], parts[2], parts[3])
+    elseif length(parts) == 2
+        SName(parts[1], parts[2], ss15"")
+    else
+        SName(parts[1], ss15"", ss15"")
+    end
```
omus, line 35: Using [incomplete initialization](https://docs.julialang.org/en/v1/manual/constructors/#Incomplete-Initialization) could be cleaner

src/types/name.jl
```diff
@@ -0,0 +1,56 @@
...
+function Base.convert(::Type{SName}, str::AbstractString)
+    name = try_convert(SName, str)
+    name isa Nothing && DomainError(str, "All timezone name parts must have length < 16")
+    return name
+end
+
+try_convert(::Type{SName}, name::SName) = name
+try_convert(::Type{String}, name::String) = name
+function try_convert(::Type{SName}, str::AbstractString)
+    parts = split(str, "/"; limit=3)
```
omus, line 27: I'd avoid limiting the `split` here. I'd rather a failure occur here if there are more than 3 parts and not be surprised later in `name_parts`

test/types/zoneddatetime.jl
```diff
@@ -77,13 +77,13 @@ using Dates: Hour, Second, UTM, @dateformat_str
...
 
         # Disambiguating parameters ignored when there is no ambiguity.
-        @test ZonedDateTime(local_dt, warsaw).zone.name == "CET"
-        @test ZonedDateTime(local_dt, warsaw, 0).zone.name == "CET"
-        @test ZonedDateTime(local_dt, warsaw, 1).zone.name == "CET"
-        @test ZonedDateTime(local_dt, warsaw, 2).zone.name == "CET"
-        @test ZonedDateTime(local_dt, warsaw, true).zone.name == "CET"
-        @test ZonedDateTime(local_dt, warsaw, false).zone.name == "CET"
-        @test ZonedDateTime(utc_dt, warsaw, from_utc=true).zone.name == "CET"
+        @test string(ZonedDateTime(local_dt, warsaw).zone.name) == "CET"
```
omus, line 80: Implementing promotion rules should avoid all these changes

src/types/zoneddatetime.jl
```diff
@@ -6,22 +6,20 @@ using Dates: AbstractDateTime, argerror, validargs
 # A `DateTime` that includes `TimeZone` information.
 # """
 
-struct ZonedDateTime <: AbstractDateTime
+struct ZonedDateTime{T<:TimeZone} <: AbstractDateTime
```
omus, line 9: I learned from Intervals.jl changing the type parameters like this should be considered a breaking change. I'd probably punt this as part of this PR

src/types/name.jl
```diff
@@ -0,0 +1,56 @@
+struct SName
+    region::ShortString15
+    locality1::ShortString15
+    locality2::ShortString15
+end
```
omus, line 5: Maybe should be a subtype of `AbstractString`?

src/types/fixedtimezone.jl
```diff
@@ -30,7 +30,7 @@ const FIXED_TIME_ZONE_REGEX = r"""
 A `TimeZone` with a constant offset for all of time.
 """
 struct FixedTimeZone <: TimeZone
-    name::String
+    name::Name
```
omus, line 33: Names here are quite different and we could probably restrict this  further to save on memory

src/types/name.jl
oxinabox: IIRC you can't tell if is isbits field has been initialized if done incompletely.
Since it will be a arbitrary legal value.

src/types/zoneddatetime.jl
oxinabox: It doesn't achieve the goal of making a ZonedDateTime with a FixedTimeZone isbits without this.

src/types/name.jl
oxinabox: I was considering it.
But it is a lot of work to subtype AbstractString. Especially because we need to match equality and thus hash with `String`.
And this is a internal type that is used for an internal field.
And the main operation that matters is comparing for equality with other `SName`s

src/types/fixedtimezone.jl
oxinabox: Can we restrict them to 15 characters outright?

test/types/zoneddatetime.jl
oxinabox: I don't think so. Happy to hear if i am wrong though.
Subtyping `AbstractString` would (but that is more work see https://github.com/JuliaTime/TimeZones.jl/pull/324#discussion_r611010337)
overloading equality would (but that can cause probelms since it would also require making hash agree and that is expensive to do)
AFIACT there is no promotion rule that gets invoked for `==` between `Strign` and a random type.

src/types/name.jl
oxinabox: Yeah you can;'t per the docs on [incomplete initalization](https://docs.julialang.org/en/v1/manual/constructors/#Incomplete-Initialization)

> However, not all object fields are references. Julia considers some types to be "plain data", meaning all of their data is self-contained and does not reference other objects. The plain data types consist of primitive types (e.g. Int) and immutable structs of other plain data types. The initial contents of a plain data type is undefined:

src/types/zoneddatetime.jl
oxinabox: But i could remove it from this PR.
though we would lose many of the benifits.
(we would get most of those benifits from just #327)

though it would make it easier to make a PR that pushed the timezone as a value in the type-parameter.
Which might be the preferred breaking change.

oxinabox:
Tests pass right now.
One concern though is that even if we remove the change to the type-parameter of ZonedDateTime (https://github.com/JuliaTime/TimeZones.jl/pull/324#discussion_r610994246)
I suspect that this will still utterly break deserialization.
I haven't checked yet though.

@omus  managed to make Intervals deserialize happily across similar changes, so maybe that can be done here also?

src/types/name.jl
omus: Why not just use a larger `ShortString` and skip this custom type all together? If `ShortString63` is larger than you want you could always tweak the size by defining a primitive type and use `ShortString{T}`. This gets you the string functionality without you having to write it.

oxinabox: If you think that is viable than sure.
Based on your earlier comments that i reinterated here https://github.com/JuliaTime/TimeZones.jl/issues/271#issuecomment-816623708
I thought it wasn't
In particular due to the BitInterger seralization one (which i now realized has been fixed).

I also though that they would be slower, as i have found that they are much slower than expected once you get large, but it seems that that doesn't really kick in for ShortString63.
It is a bit slower for some operations but  a bit faster for others.

## Benchmarks

### == false
```julia
julia> @btime x==y setup=(x=convert(TimeZones.SName, "America/Winnipeg"); y=convert(TimeZones.SName, "America/Argentina/ComodRivadavia"))
  1.504 ns (0 allocations: 0 bytes)
false

julia> @btime x==y setup=(x=ShortString63("America/Winnipeg"); y=ShortString63("America/Argentina/ComodRivadavia"))
  2.117 ns (0 allocations: 0 bytes)
false
```

### == true
(basically exactly the same as the false case)
```julia

julia> @btime x==y setup=(x=convert(TimeZones.SName, "America/Winnipeg"); y=convert(TimeZones.SName, "America/Winnipeg"))
  1.505 ns (0 allocations: 0 bytes)
true

julia> @btime x==y setup=(x=ShortString63("America/Winnipeg"); y=ShortString63("America/Winnipeg"))
  2.115 ns (0 allocations: 0 bytes)
true
```

### hash
```julia

julia> @btime hash(x)==hash(y) setup=(x=convert(TimeZones.SName, "America/Winnipeg"); y=convert(TimeZones.SName, "America/Argentina/ComodRivadavia"))
  34.256 ns (0 allocations: 0 bytes);
  
julia> @btime hash(x)==hash(y) setup=(x=ShortString63("America/Winnipeg"); y=ShortString63("America/Argentina/ComodRivadavia"));
  27.598 ns (0 allocations: 0 bytes)
```


So I will make that change, since it is simpler

oxinabox:
I have broken this out into https://github.com/JuliaTime/TimeZones.jl/pull/332
and https://github.com/JuliaTime/TimeZones.jl/pull/330

Issue #326, opened 2021-04-12
Should FixedTimeZones have names?


oxinabox:
Relating to #271 , if we removed the `name` field from the `FixedTimeZone` type then it would be `isbits`.

RIght now, as I understand the situation:

 - `FixedTimeZone(name)`: actually discards the name you pass it. It pulls out the key information to determine the offset, and then constructs a new name in standard form of `UTC±HH:MM[:SS]`
 - OTOH `FixedTimeZone(name, offset, dst_offset)` keeps the name. 
 - Names *do* have to match for timezones to be considered equal
    -   `FixedTimeZone("UTC") != FixedTimeZone("Z", 0)`
    - I am not sure this makes sense for `FixedTimeZone`s. It does for VariableTimeZones, since they might vary later even if transitions and offset are same now
 - The name *is* used for displaying
    - an alternative would be to generate the name in standard form of `UTC±HH:MM[:SS]` for purposes of displaying

omus:
As `VariableTimeZone`s use `FixedTimeZone`s as part of the `Transition`s removing the `name` field would result in no longer being able to access the abbreviations names commonly used in North America.

oxinabox:
This is true.
Though the only place this is exposed at all to the user in the public API is: [`show_next_transition`](https://juliatime.github.io/TimeZones.jl/stable/api-public/#TimeZones.show_next_transition)
```
julia> show_next_transition(tz"America/New_York")
Transition Date:   2021-11-07
Local Time Change: 02:00 → 01:00 (Backward)
Offset Change:     UTC-5/+1 → UTC-5/+0
Transition From:   2021-11-07T01:59:59.999-04:00 (EDT)
Transition To:     2021-11-07T01:00:00.000-05:00 (EST)
```
would instread show up as:

```
julia> show_next_transition(tz"America/New_York")
Transition Date:   2021-11-07
Local Time Change: 02:00 → 01:00 (Backward)
Offset Change:     UTC-5/+1 → UTC-5/+0
Transition From:   2021-11-07T01:59:59.999-04:00
Transition To:     2021-11-07T01:00:00.000-05:00
```

The benifit is we would be saving 1 pointer (4 or 8 bytes depending on 32 or 64 bit) + about 16 bytes of content from every FixedTimeZone.
and we would save hundreds of pointer from many `VariableTimeZone`.
```julia
julia> length(tz"America/New_York".transitions)
237

julia> length(tz"America/Winnipeg".transitions)
187

[...]

PR #327, opened 2021-04-12
make FixedTimeZones be isbits by using ShortStrings

oxinabox:
Contributes towards #271 ,
following up on just this part of https://github.com/JuliaTime/TimeZones.jl/pull/324#discussion_r610994546

requires https://github.com/JuliaString/ShortStrings.jl/pull/46

Is  less extreme than https://github.com/JuliaTime/TimeZones.jl/issues/326

### DIFF
--- a/src/types/fixedtimezone.jl
+++ b/src/types/fixedtimezone.jl
@@ -1,3 +1,7 @@
+# Ideally would always use ShortString15, but it's `hash` is broken on 32-bit systems.
+# https://github.com/JuliaString/MurmurHash3.jl/issues/12
+const FixedTimeZoneName = Int === Int64 ? ShortString15 : String
+
 const FIXED_TIME_ZONE_REGEX = r"""
     ^(?|
         Z
@@ -30,7 +34,7 @@ const FIXED_TIME_ZONE_REGEX = r"""
 A `TimeZone` with a constant offset for all of time.
 """
 struct FixedTimeZone <: TimeZone
-    name::String
+    name::FixedTimeZoneName
     offset::UTCOffset
 end
 
@@ -72,7 +76,7 @@ UTC+15:45:21
 function FixedTimeZone(s::AbstractString)
     s == "Z" && return UTC_ZERO
 
-    m = match(FIXED_TIME_ZONE_REGEX, s)
+    m = match(FIXED_TIME_ZONE_REGEX, String(s))
     m === nothing && throw(ArgumentError("Unrecognized time zone: $s"))
 
     coefficient = m[:sign] == "-" ? -1 : 1
--- a/test/types/fixedtimezone.jl
+++ b/test/types/fixedtimezone.jl
@@ -41,4 +41,14 @@
         fixed_tz = FixedTimeZone("UTC")
         @test size(fixed_tz .== fixed_tz) == ()
     end
+
+    @testset "isbits" begin
+        # We are not using ShortStrings on 32 bit due to hash being broken on it.
+        # see https://github.com/JuliaString/MurmurHash3.jl/issues/12
+        if Int === Int64
+            @test isbits FixedTimeZone("0123")
+        else
+            @test_broken isbits FixedTimeZone("0123")
+        end
+    end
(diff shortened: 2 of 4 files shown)


oxinabox:
The CI/Benchmarks job is failing because it is trying to deserialize a something that has had it's type represenation changed.
I am suprised that it records anything as julia objects (rather than just chunks of code that generate julia objects)

src/tzdata/compile.jl
```diff
@@ -697,7 +697,9 @@ function compile(tz_source::TZSource, dest_dir::AbstractString; kwargs...)
     empty!(TIME_ZONE_CACHE)
 
     for (tz, class) in results
-        parts = split(TimeZones.name(tz), '/')
+        # Need to convert all abstract strings to `String`s because ShortString's don't
+        # support split.
+        parts = split(String(TimeZones.name(tz)), '/')
         tz_path = joinpath(dest_dir, parts...)
```
oxinabox, line 703: Actually i fixed that in ShortStrings v0.3.6 so that is fine.
```suggestion
        parts = split(TimeZones.name(tz), '/')
```

test/types/fixedtimezone.jl
```diff
@@ -41,4 +41,14 @@
...
         @test size(fixed_tz .== fixed_tz) == ()
     end
+
+    @testset "isbits" begin
+        # We are not using ShortStrings on 32-bit due to hash being broken on 32-bit.
+        # See https://github.com/JuliaString/MurmurHash3.jl/issues/12
+        if Int === Int64
+            @test isbits FixedTimeZone("0123")
+        else
+            @test_broken isbits FixedTimeZone("0123")
```
oxinabox, line 51:
```suggestion
            @test isbits(FixedTimeZone("0123"))
        else
            @test_broken isbits(FixedTimeZone("0123"))
```

src/types/fixedtimezone.jl
```diff
@@ -72,7 +76,7 @@ UTC+15:45:21
 function FixedTimeZone(s::AbstractString)
     s == "Z" && return UTC_ZERO
 
-    m = match(FIXED_TIME_ZONE_REGEX, s)
+    m = match(FIXED_TIME_ZONE_REGEX, String(s))
```
omus, line 79: Did you see an error before changing this? I'll note `SubString{String}` also works here so we don't always need to call `String`

src/types/fixedtimezone.jl
```diff
@@ -1,3 +1,7 @@
+# Ideally would always use ShortString15, but it's `hash` is broken on 32-bit systems.
+# https://github.com/JuliaString/MurmurHash3.jl/issues/12
+const FixedTimeZoneName = Int === Int64 ? ShortString15 : String
```
omus, line 3: I really wish the 32-bit problem was properly fixed... if this is what we need to do then let's be as clear about it as we can.


```suggestion
const FixedTimeZoneName = Sys.WORD_SIZE == 64 ? ShortString15 : String
```

src/types/fixedtimezone.jl
```diff
@@ -30,7 +34,7 @@ const FIXED_TIME_ZONE_REGEX = r"""
 A `TimeZone` with a constant offset for all of time.
 """
 struct FixedTimeZone <: TimeZone
-    name::String
+    name::FixedTimeZoneName
```
omus, line 37: Once ShortStrings.jl this type alias won't survive. May as well try to indicate what it truly is.
```suggestion
    name::_ShortString15
```

src/types/fixedtimezone.jl
oxinabox: Yes, it errors before

Iirc I looked in the docs for regex and it explicitly said it only worked on String. Which is a weird thing for Julia docs to say.
And also weird that it doesn't convert abstract strings to strings for it.

src/types/fixedtimezone.jl
oxinabox: It is now fixed

PR #328, opened 2021-04-13
disable fail-fast on GHA CI

oxinabox:
Fail-fast on GHA doesn't work great.
The benchmark job erroring shouldn't prevent the actual testing jobs from running.
Since the benchmarking job can error if you change how any types are defined https://github.com/JuliaTime/TimeZones.jl/pull/327#issuecomment-818623379

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,7 @@ jobs:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         version:
           - "1.0"  # LTS


omus:
Are you sure fast-fail works this way? Seems strange that it looks at jobs that aren't part of it's on matrix

oxinabox:
No, I am not sure.

omus:
> When set to` true`, GitHub cancels all in-progress jobs if any `matrix` job fails. Default: `true`
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

Seems like this documentation is wrong then. Would be good to open an issue.

omus:
So the solution to this is to actually set `fast-fail: false` on just the `benchmark` job. With this then a failure from the benchmark job will not impact any other running job but jobs with `fast-fail: true` will cause other jobs to be cancelled. I'll make a PR for this.

PR #329, opened 2021-04-14
replace EzXML dependency by regex parsing

laborg:
This PR replaces the XML parsing by using a simple regex leading to the removal of EzXML as a dependency. I understand that that usually regex parsing of XML files is a nogo, but in this case the sources files seem stable enough. 
Based on data from JuliaHub, TimeZones.jl is an indirect dependency of 315 packages, and getting rid of EzXML might be benefitial for those packages.

### DIFF
--- a/src/winzone/WindowsTimeZoneIDs.jl
+++ b/src/winzone/WindowsTimeZoneIDs.jl
@@ -32,8 +32,8 @@ function compile(xml_file::AbstractString)
     for line in readlines(xml_file)
         # Territory "001" is the global default
         occursin("territory=\"001\"", line) || continue
-        win_name = match(r"other=\"(.+?)\"",line)[1]
-        posix_name = match(r"type=\"(.+?)\"",line)[1]
+        win_name = match(r"other=\"(.*?)\"", line)[1]
+        posix_name = match(r"type=\"(.*?)\"", line)[1]
         translation[win_name] = posix_name
     end
 
--- a/src/winzone/WindowsTimeZoneIDs.jl
+++ b/src/winzone/WindowsTimeZoneIDs.jl
@@ -31,6 +31,7 @@ function compile(xml_file::AbstractString)
         posix_name = match(r"type=\"(.+?)\"",line)[1]
         translation[win_name] = posix_name
     end
+
     return translation
 end
 
--- a/src/winzone/WindowsTimeZoneIDs.jl
+++ b/src/winzone/WindowsTimeZoneIDs.jl
@@ -24,6 +24,11 @@ function compile(xml_file::AbstractString)
     translation = Dict{String,String}()
 
     # Get the timezone conversions from the file
+    #
+    # Note: Since the XML file is simplistic enough that we can parse what we need via a
+    # regex we can avoid having an XML package dependency. Additionally, since this XML file
+    # is included as part of the this package we can correct any parsing issues before a
+    # TimeZones.jl release occurs.
     for line in readlines(xml_file)
         # Territory "001" is the global default
         occursin("territory=\"001\"", line) || continue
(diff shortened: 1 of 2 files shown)


oxinabox:
Is this a real issue you are facing, or just a theoretical one?

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -21,22 +20,16 @@ const WINDOWS_XML_FILE = joinpath(WINDOWS_XML_DIR, "windowsZones.xml")
...
 
-    # Dictionary to store the windows to time zone conversions
     translation = Dict{String,String}()
-    for map_zone in map_zones
-        win_name = map_zone["other"]
-        posix_name = map_zone["type"]
+
+    # Get the timezone conversions from the file
+    for line in readlines(xml_file)
+        occursin("001",line) || continue
```
omus, line 28: Let's make this a little more robust against false positives
```suggestion
        # Territory "001" is the global default
        occursin("territory=\"001\"", line) || continue
```

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -21,22 +20,16 @@ const WINDOWS_XML_FILE = joinpath(WINDOWS_XML_DIR, "windowsZones.xml")
...
+        win_name = match(r"other=\"(.+?)\"",line)[1]
+        posix_name = match(r"type=\"(.+?)\"",line)[1]
```
omus, line 30:
```suggestion
        win_name = match(r"other=\"(.*?)\"", line)[1]
        posix_name = match(r"type=\"(.*?)\"", line)[1]
```

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -21,22 +20,16 @@ const WINDOWS_XML_FILE = joinpath(WINDOWS_XML_DIR, "windowsZones.xml")
...
         translation[win_name] = posix_name
     end
```
omus, line 32: Doesn't seem necessary to remove the whitespace here
```suggestion
    end

```

src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -21,22 +20,16 @@ const WINDOWS_XML_FILE = joinpath(WINDOWS_XML_DIR, "windowsZones.xml")
...
-
-    # TODO: Eliminate the Etc/* POSIX names here? See Windows section of `localzone`
 
-    # Dictionary to store the windows to time zone conversions
     translation = Dict{String,String}()
-    for map_zone in map_zones
-        win_name = map_zone["other"]
-        posix_name = map_zone["type"]
+
+    # Get the timezone conversions from the file
```
omus, line 26:
```suggestion
    # Get the timezone conversions from the file
    #
    # Note: Since the XML file is simplistic enough that we can parse what we need via a
    # regex we can avoid having an XML package dependency. Additionally, since this XML file
    # is included as part of the this package we can correct any parsing issues before a
    # TimeZones.jl release occurs.

omus (commented):
Thanks, this is something I've considered doing before

laborg:
First off: Thanks for working on such an essential package!
For me the additional dependency its just a theoretical problem (I was skimming through common Julia packages and have seen a couple of them having non-intuitive dependencies). Due to the popularity of TimeZones.jl it seems right to be picky on adding dependencies.

omus (approved):
As the XML parsing is so limited in scope and also only used on Windows I'm fine with dropping it as a dependency

omus:
It looks like the docs Manifest.toml needs to be updated before merging. I'll try to do that tomorrow

omus:
I've moved these changes to https://github.com/JuliaTime/TimeZones.jl/pull/352 which is a branch I can modify

PR #330, opened 2021-04-14
Use ShortStrings for names of both Fixed and Variable timezones

oxinabox:
This is made on top of #327 
it is based on this comment https://github.com/JuliaTime/TimeZones.jl/pull/324#discussion_r613158657

contributes towards #271 
but doesn't solve it since VariableTimeZones still have Vector fields,
and the ZonedDateTime isnb't parameteric on it's timezone type.


omus (commented):
So much cleaner! I'm feeling good about this. I'll reach out to Scott about getting 32-bit fixed

Issue #318, opened 2020-12-21
Parameterize `ZonedDateTime` by `TimeZone`


omus:
Allows for enforcing that a `Vector` contains `ZonedDateTime`s all using the same `TimeZone`. This idea has been floated around before so I made this issue to centralize the conversation.

omus:
Requested in: https://github.com/JuliaData/Arrow.jl/pull/50#issue-509313802

oxinabox:
For it to be parametric on the *value* of the the timezone (ratehr than the type of the timezone) would require that the TimeZone type `isbits`.
Which is #271

omus:
> For it to be parametric on the _value_ of the the timezone (ratehr than the type of the timezone) would require that the TimeZone type `isbits`.
> Which is #271

So this issue is referring to having something like `ZonedDateTime{tz"America/Winnipeg"}`. Effectively this would mean having a separate type for each time zone making each time zone a singleton. The advantage here is you can use type constraints to enforce all elements of an array to be the same time zone and excise storing that information in the struct. The downside however additional overhead on Julia's type system with introducing 300+ types; complications with custom time zones and possibly serialization

omus:
An idea I'm experimenting with is using an intermediate type for the time zone. The advantage of such an approach is that it allows us to use a time zone as a type parameter without requiring the type storing the time zone information to have to be `isbits` which can be quite challenging.

Essentially this could work like:

```julia
using Dates: DateTime
using TimeZones: TimeZones, TimeZone, @tz_str

struct TZ{name} end
TZ{name}() where {name} = TimeZone(String(name))

macro TZ_str(name)
    :(TZ{$(Expr(:quote, Symbol(name)))})
end

struct ZonedDateTime{TZ}
    utc_instant::DateTime
    # zone::FixedTimeZone
end

function ZonedDateTime(dt::DateTime, tz::TimeZone)
    tz_type = TZ{Symbol(tz.name)}
    ZonedDateTime{tz_type}(dt)
end
TimeZones.timezone(::ZonedDateTime{TZ}) where TZ = TZ()
```

```julia
julia> ZonedDateTime{TZ"America/Winnipeg"}(DateTime(2021))
ZonedDateTime{TZ{Symbol("America/Winnipeg")}}(DateTime("2021-01-01T00:00:00"))

julia> ZonedDateTime(DateTime(2021), tz"America/Winnipeg")
[...]

PR #332, opened 2021-04-14
Make ZonedDateTime parametric on timezone type

oxinabox:
This is just the parametric type bit from https://github.com/JuliaTime/TimeZones.jl/pull/324#pullrequestreview-635205760

Contirbutes to #271  once we sort out the timezone types

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -17,7 +17,7 @@ function Base.:(-)(zdt::ZonedDateTime, p::TimePeriod)
     return ZonedDateTime(DateTime(zdt, UTC) - p, timezone(zdt); from_utc=true)
 end
 
-function broadcasted(::typeof(+), r::StepRange{ZonedDateTime}, p::DatePeriod)
+function broadcasted(::typeof(+), r::StepRange{<:ZonedDateTime}, p::DatePeriod)
     start, step, stop = first(r), Base.step(r), last(r)
 
     # Since the local time + period can result in an invalid local datetime when working with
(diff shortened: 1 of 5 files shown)


oxinabox:
Wow @omus  was right, adding a type parameters really does break a lot of things.
It is generally easy enough to be compatible with both but a log ot `Vector{ZonedDateTime}` need to become `Vector{<:ZonedDateTime}` (which works with both).

The other thing we should maybe do is add a format versions to the compiled dir file path.
This wouldn't matter for the end user since they would only ever see 1 version since the `DEPS_DIR` is per version of TimeZones.jl, but for anyone working on it and changing branches. It is a bit annoying

oxinabox:
We need a promotion rule for different `ZonedDateTime{T}`s. .
The natural one is to go to the unionall `ZonedDateTime`, it is what owuld be returned if we were not overriding it in:
https://github.com/JuliaTime/TimeZones.jl/blob/906cea08c2ca4a5c76b12e14fb60d5df7990fbf7/src/types/zoneddatetime.jl#L163-L165

Though an alternative might be to all `astimezone` on all of them, so they all become Variable, or they all become Fixed.
I am not sure on implictaions of that.

but if we don't have a proimotion rule the following happens:

```julia
julia> [ZonedDateTime(Date(2000), tz"America/Winnipeg"), ZonedDateTime(Date(2001), tz"UTC")]
ERROR: no promotion exists for ZonedDateTime{VariableTimeZone} and ZonedDateTime{FixedTimeZone}
Stacktrace:
 [1] error(::String, ::Type, ::String, ::Type)
   @ Base ./error.jl:42
 [2] promote_rule(#unused#::Type{ZonedDateTime{VariableTimeZone}}, #unused#::Type{ZonedDateTime{FixedTimeZone}})
   @ TimeZones ~/JuliaEnvs/BidDatabase.jl/dev/TimeZones/src/types/zoneddatetime.jl:162
 [3] promote_type
   @ ./promotion.jl:233 [inlined]
 [4] promote_typeof(x::ZonedDateTime{VariableTimeZone}, xs::ZonedDateTime{FixedTimeZone})
   @ Base ./promotion.jl:272
 [5] vect(::ZonedDateTime{VariableTimeZone}, ::Vararg{Any, N} where N)
   @ Base ./array.jl:126
 [6] top-level scope
   @ REPL[18]:1
```

omus:
> Though an alternative might be to all astimezone on all of them, so they all become Variable, or they all become Fixed.

I'm definitely against converting the time zones automatically through promotion. I can see having heterogeneous `ZonedDateTime` types in the same `Vector` being useful even if it isn't super efficient. I think promoting to `ZonedDateTime{TimeZone}` seems reasonable. 

A similar issue would also exist for: https://github.com/JuliaTime/TimeZones.jl/issues/318

omus:
Just a thought: what if we introduced a new alternative type to `ZonedDateTime`? That should allow us to experiment with more breaking changes in a more opt-in manner.

oxinabox:
My current preferred solution is to introduce `UTCZonedDateTime`.
It would make the most common case super-lightweight no need to reference timezone at all. and it's the case we can  optimise  interaction with ZonedDateTimes really well.

omus:
> My current preferred solution is to introduce `UTCZonedDateTime`.

Relevant: https://github.com/invenia/Intervals.jl/issues/178#issuecomment-901347323

PR #335, opened 2021-04-15
Introduce a bitstype just for IANA Variable TimeZones

oxinabox:
This is yet another PR relating to #271 
It doesn't solve it totally,
It needs #327  to handling FixedTimeZones, and #332  to make the ZonedDateTime parametric.

It is most similar to https://github.com/JuliaTime/TimeZones.jl/pull/287  / https://github.com/JuliaTime/TimeZones.jl/pull/323
but with rather than a dynamically created lookup table as timezones are used, it has a static (but lazy) look up table.
This table always maps the same integer id, to the same timezone.
This means we don't need to worry about serialization.

Right now this code is kinda gross.
Especially around loading it.
We have 2 caches, and idk we probably only need 1.

### DIFF
--- a/src/discovery.jl
+++ b/src/discovery.jl
@@ -154,15 +154,16 @@ next_transition_instant
 
 function next_transition_instant(zdt::ZonedDateTime)
     tz = zdt.timezone
-    tz isa VariableTimeZone || return nothing
+    tz isa AbstractVariableTimeZone || return nothing
 
+    transits = transitions(tz)
     # Determine the index of the transition which occurs after the UTC datetime specified
     index = searchsortedfirst(
-        tz.transitions, DateTime(zdt, UTC),
+        transits, DateTime(zdt, UTC),
         by=el -> isa(el, TimeZones.Transition) ? el.utc_datetime : el,
     )
 
-    index <= length(tz.transitions) || return nothing
+    index <= length(transits) || return nothing
 
     # Use the UTC datetime of the transition and the offset information prior to the
     # transition to create a `ZonedDateTime` which cannot be constructed with the high-level
@@ -170,8 +171,8 @@ function next_transition_instant(zdt::ZonedDateTime)
     # transition but visually appears to be before the transition. For example in a
     # transition where the clock changes from 01:59 → 03:00 we would return 02:00 where
     # the UTC datetime of 02:00 == 03:00.
-    utc_datetime = tz.transitions[index].utc_datetime
-    prev_zone = tz.transitions[index - 1].zone
+    utc_datetime = transits[index].utc_datetime
+    prev_zone = transits[index - 1].zone
     ZonedDateTime(utc_datetime, tz, prev_zone)
 end
 
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -163,3 +163,7 @@ function last_valid(local_dt::DateTime, tz::VariableTimeZone)
     possible = interpret(local_dt, tz, Local)
     return isempty(possible) ? first(shift_gap(local_dt, tz)) : last(possible)
 end
+
+
+first_valid(dt::DateTime, tz::IANATimeZone, args...) = _do_and_rewrap(first_valid, dt, tz, args...)
+last_valid(dt::DateTime, tz::IANATimeZone, args...) = _do_and_rewrap(last_valid, dt, tz, args...)
--- a/src/types/ianatimezone.jl
+++ b/src/types/ianatimezone.jl
@@ -1,0 +1,136 @@
+
+# This is extremely redundant but still only about 8MB for each of our 2 tables
+# and it avoids any need to write a smarter but more CPU time expensive perfect hash
+# This needs to be big enough to avoid any collisions
+# also extra size is useful because it means we are probably safe if new timezones are added
+const IANA_TABLE_SIZE = 2^20
+
+const IANA_TIMEZONES = Vector{VariableTimeZone}(undef, IANA_TABLE_SIZE)
+
+# TODO: maybe fill this during build(), probably by generating a julia file.
+# That way we can avoid actually instantitating every timezone til it is needed.
+const IANA_NAMES = Vector{String}(undef, IANA_TABLE_SIZE)
+function init_IANA_NAMES!()  # this is run by __init__ (at least for now)
+    for name in timezone_names()
+        # TODO: we should workout how to filter out FixedTimeZones here
+        mod_id = iana_mod_id(name)
+        # Important: Make sure our hash is perfect (even module the table size)
+        isassigned(IANA_NAMES, mod_id) && error("hash collision for $tz, at $mod_id")
+        IANA_NAMES[mod_id] = name
+    end
+    return IANA_NAMES
+end
+
+# have checked that this is perfect
+perfect_hash(tz::VariableTimeZone, h=zero(UInt)) = perfect_hash(tz.name, h)
+function perfect_hash(name::AbstractString, h=zero(UInt))
+    h = hash(:timezone, h)
+    h = hash(name, h)
+    return h
+end
+
+iana_mod_id(str_or_var_tz) = iana_mod_id(perfect_hash(str_or_var_tz))
+iana_mod_id(id::UInt) = mod1(id, IANA_TABLE_SIZE)
+
+function is_standard_iana(str::AbstractString)
+    mod_id = iana_mod_id(str)
+    return isassigned(IANA_NAMES, mod_id) && IANA_NAMES[mod_id] == str
+end
+
+function get_iana_timezone!(str::AbstractString)
+    mod_id = iana_mod_id(str)
+    if isassigned(IANA_TIMEZONES, mod_id)
+        IANA_TIMEZONES[mod_id]
+    else
+        tz_path = joinpath(TZData.COMPILED_DIR, split(str, "/")...)
+        tz, class = deserialize(tz_path)
+        # TODO: maybe here is where we check if it is a FixedTimeZone, and if so don't remember it?
+        if tz isa VariableTimeZone
+            IANA_TIMEZONES[mod_id] = tz
+            return IANATimeZone(perfect_hash(str))
+        else
+            # it is a FixedTimeZone, we are not going to use a IANATimeZone
+            return tz
+        end
+    end
+end
+
+function get_iana_timezone!(id::UInt)
+    mod_id = iana_mod_id(id)
+    if isassigned(IANA_NAMES, mod_id)
+        name = IANA_NAMES[mod_id]
+        return get_iana_timezone!(name)
+    else
+        error(
+            "$id does not correspond to any known IANA timezone. " *
+            "Check you are using the right version of the IANA database.",
+        )
+    end
+end
+
+
+"""
+    IANATimeZone(::AbstractString) <: AbstractVariableTimeZone
+
+A type for representing a standard variable IANA TimeZome from the tzdata.
+Under-the-hood it stores only a unique integer identifier.
+"""
+struct IANATimeZone <: TimeZone
+    # id must be a hash of the corresponding Variable/FixedTimeZone
+    # and it is only possible if `hash` on all timezones in tzdata happens to be perfect
+    # This is the real hash, not the hash modulo IANA_TABLE_SIZE
+    # because that way we can in the future change IANA_TABLE_SIZE and not invalidate old
+    # serialized data.
+    id::UInt
+end
+
+function IANATimeZone(name::AbstractString)
+    return IANATimeZone(perfect_hash(name))
+end
+
+backing_timezone(itz::IANATimeZone) = get_iana_timezone!(itz.id)
+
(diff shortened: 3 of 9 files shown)


src/types/ianatimezone.jl
```diff
@@ -0,0 +1,136 @@
...
+end
+
+function get_iana_timezone!(str::AbstractString)
+    mod_id = iana_mod_id(str)
+    if isassigned(IANA_TIMEZONES, mod_id)
+        IANA_TIMEZONES[mod_id]
+    else
+        tz_path = joinpath(TZData.COMPILED_DIR, split(str, "/")...)
+        tz, class = deserialize(tz_path)
+        # TODO: maybe here is where we check if it is a FixedTimeZone, and if so don't remember it?
```
oxinabox, line 47:
```suggestion
```

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,136 @@
...
+    end
+    return IANA_NAMES
+end
+
+# have checked that this is perfect
+perfect_hash(tz::VariableTimeZone, h=zero(UInt)) = perfect_hash(tz.name, h)
+function perfect_hash(name::AbstractString, h=zero(UInt))
+    h = hash(:timezone, h)
+    h = hash(name, h)
+    return h
```
oxinabox, line 29: I don't think we want to use `hash` for our perfect hash.
Even though we got lucky and it is perfect.
Julia might change it, plus it is not very minimal.

I was hoping that we could reuse it for the `hash` but it seems we can't.

I think I want to play around with `gperf` and generate a more minimal perfect hash function.
which should also let us more confidently handle new timezones being added, so that we can specifically dodge collisions that might be introduced.

src/types/timezone.jl
```diff
@@ -46,7 +46,10 @@ function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
     tz, class = get!(TIME_ZONE_CACHE, str) do
```
oxinabox, line 46: @omus  input on the best way to rewrite this function would apprecated.

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,159 @@
...
+Base.:(==)(a::IANATimeZone, b::TimeZone) = backing_timezone(a) == b
+Base.:(==)(b::TimeZone, a::IANATimeZone) = backing_timezone(a) == b
+
+# TODO: we have the hash, it seems like we should be able to use that to get seeded hash
+Base.hash(a::IANATimeZone, seed::UInt) = hash(backing_timezone(a), seed)
+
+name(a::IANATimeZone) = name(backing_timezone(a))
+transitions(tz::IANATimeZone) = transitions(backing_timezone(tz))
+
+# TODO: should i just make this check the fields of VariableTimeZone and just delegate all?
```
oxinabox, line 126: this should be addressed

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,159 @@
...
+    return IANATimeZone(perfect_hash(name))
+end
+
+backing_timezone(itz::IANATimeZone) = get_iana_timezone!(itz.id)::VariableTimeZone
+
+Base.:(==)(a::IANATimeZone, b::IANATimeZone) = a.id == b.id
+Base.:(==)(a::IANATimeZone, b::TimeZone) = backing_timezone(a) == b
+Base.:(==)(b::TimeZone, a::IANATimeZone) = backing_timezone(a) == b
+
+# TODO: we have the hash, it seems like we should be able to use that to get seeded hash
```
oxinabox, line 120:
```suggestion
# We can't use our perfect hash is as won't agree with the hash of the backing_timezone
```

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,159 @@
...
+      1681, 1681, 1681,    4, 1681,    1,    3,    3,  241,   48,
+         1,   22,   10,    9,   61,   18,   74,    9,   30, 1681,
+      1681, 1681, 1681, 1681, 1681,    4,    2,   18,  167,   48,
+       547,  691,  532,  446,  574,  466,  211,  452,    8,  387,
+       110,  475,  601,  262,   83,  124,  580,  536,   60,    1,
+       719,    2, 1681, 1681, 1681,  291,    1,    1,   27,  424,
+        13,    5,   43,    3,  355,   12,  168,  148,   90,  179,
+         4,    1,  315,    3,    3,    3,    2,   37,    5,   65,
+        22,  320,  245,    1, 1681, 1681, 1681,
+    )
```
oxinabox, line 28: How should i rewrap this?
10 per line?
Up to the line limit?
1 per line?

src/discovery.jl
```diff
@@ -154,24 +154,25 @@ next_transition_instant
 
 function next_transition_instant(zdt::ZonedDateTime)
     tz = zdt.timezone
-    tz isa VariableTimeZone || return nothing
+    tz isa AbstractVariableTimeZone || return nothing
 
+    transits = transitions(tz)
```
oxinabox, line 159: It might be premature optimization to pull this out to be done before.
Idea is to avoid relooking up the `backing_timezone` for `IANATimeZone` multiple times

src/interpret.jl
```diff
@@ -163,3 +163,7 @@ function last_valid(local_dt::DateTime, tz::VariableTimeZone)
     possible = interpret(local_dt, tz, Local)
     return isempty(possible) ? first(shift_gap(local_dt, tz)) : last(possible)
 end
+
+
+first_valid(dt::DateTime, tz::IANATimeZone, args...) = _do_and_rewrap(first_valid, dt, tz, args...)
```
oxinabox, line 168: Probnably `_do_and_rewrap` needs to be renamed

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,159 @@
...
+    len >= 1 && (hval += asso_values[units[1]])
+    len > 0 && (hval += asso_values[units[end]])  # add the last
+    return hval
+end
+
+
+const IANA_TIMEZONES = Vector{VariableTimeZone}(undef, IANA_TABLE_SIZE)
+
+# TODO: maybe fill this during build(), probably by generating a julia file.
+# That way we can avoid actually instantitating every timezone til it is needed.
```
oxinabox, line 51:
```suggestion
```

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,159 @@
...
+const IANA_NAMES = Vector{String}(undef, IANA_TABLE_SIZE)
+function init_IANA_NAMES!()  # this is run by __init__ (at least for now)
+    for name in timezone_names()
+        # TODO: we should workout how to filter out FixedTimeZones here
```
oxinabox, line 55:
```suggestion
```

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,159 @@
...
+        id = perfect_hash(name)
+        # Important: Make sure our hash is perfect (even module the table size)
```
oxinabox, line 57:
```suggestion
        # Important: this line make's sure our hash is indeed perfect
```

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,157 @@
...
+
+
+const IANA_TIMEZONES = Vector{VariableTimeZone}(undef, IANA_TABLE_SIZE)
+
+# TODO: maybe fill this during build(), probably by generating a julia file.
+const IANA_NAMES = Vector{String}(undef, IANA_TABLE_SIZE)
+function init_IANA_NAMES!()  # this is run by __init__ (at least for now)
+    for name in timezone_names()
+        id = perfect_hash(name)
+        # Important: this line make's sure our hash is indeed perfect
```
oxinabox, line 55:
```suggestion
        # Important: this line makes sure our hash is indeed perfect
```

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,157 @@
...
+    len >= 2 && (hval += asso_values[units[2] + 1])
+    len >= 1 && (hval += asso_values[units[1]])
+    len > 0 && (hval += asso_values[units[end]])  # add the last
+    return hval
+end
+
+
+const IANA_TIMEZONES = Vector{VariableTimeZone}(undef, IANA_TABLE_SIZE)
+
+# TODO: maybe fill this during build(), probably by generating a julia file.
```
oxinabox, line 50: Or maybe it is fine, since it only queries what files exist, using `timezone_names` rather than reading them.
Maybe `timezone_names` (or `__init__`) should do the validation that TimeZones.jl has build properly and has the `TZDATA.COMPILED_DIR`.
Which would maybe clean up `TimeZone` a bit`.

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,157 @@
...
+        return name(tz)
+    elseif s == :transitions
+        return transitions(tz)
+    else
+        return getfield(tz, s)
+    end
+end
+function Base.hasproperty(tz::IANATimeZone, s::Symbol)
+    return s === :name || s === :transitions || hasfield(IANATimeZone, s)
+end
```
oxinabox, line 136:
```suggestion
if isdefined(Base, :hasproperty)
    function Base.hasproperty(tz::IANATimeZone, s::Symbol)
        return s === :name || s === :transitions || hasfield(IANATimeZone, s)
    end
end
```

src/types/ianatimezone.jl
```diff
@@ -0,0 +1,159 @@
...
+    len >= 19 && (hval += asso_values[units[19]])
+    len >= 12 && (hval += asso_values[units[12]])
+    len >= 11 && (hval += asso_values[units[11]])
+    len >= 9 && (hval += asso_values[units[9] + 1])
+    len >= 8 && (hval += asso_values[units[8]])
+    len >= 6 && (hval += asso_values[units[6] + 1])
+    len >= 4 && (hval += asso_values[units[4]])
+    len >= 2 && (hval += asso_values[units[2] + 1])
+    len >= 1 && (hval += asso_values[units[1]])
+    len > 0 && (hval += asso_values[units[end]])  # add the last
```
iamed2, line 43:
```suggestion
    @inbounds begin
        len >= 19 && (hval += asso_values[units[19]])
        len >= 12 && (hval += asso_values[units[12]])
        len >= 11 && (hval += asso_values[units[11]])
        len >= 9 && (hval += asso_values[units[9] + 1])
        len >= 8 && (hval += asso_values[units[8]])
        len >= 6 && (hval += asso_values[units[6] + 1])
        len >= 4 && (hval += asso_values[units[4]])
        len >= 2 && (hval += asso_values[units[2] + 1])
        len >= 1 && (hval += asso_values[units[1]])
        len > 0 && (hval += asso_values[units[end]])  # add the last
    end
```
This saves some time and simplifies the native code but relies on each code unit being in the range `1:125`. 

Might be worth it just to pad the array of `asso_values` to cover the domain of `UInt8`

oxinabox: > This saves some time and simplifies the native code but relies on each code unit being in the range 1:125

I tried this and didn't notice any improvement.
How were you benchmarking it?

> Might be worth it just to pad the array of asso_values to cover the domain of UInt8

I tried this and it made it *way* worse. I think because the tuple gets too lange for julia to be willing to do some optimizations.

iamed2:
```
julia> @btime for x in ["America/Winnipeg", "Etc/UTC", "Canada/Central"]
           perfect_hash(x)
       end
  71.438 ns (1 allocation: 112 bytes)

julia> @btime for x in ["America/Winnipeg", "Etc/UTC", "Canada/Central"]
           perfect_hash_inbounds(x)
       end
  66.137 ns (1 allocation: 112 bytes)
```

oxinabox: running `@btime` on the whole testset in `test/types/ianatimezone.jl`
With the benchmark

```julia
@btime for x in ("America/Winnipeg", "Etc/UTC", "Canada/Central")
                TimeZones.perfect_hash(x)
             end
```
Noting that i got rid of the vector in the benchmark code since that was dominating.
In julia 1.6

 - current code without the inbounds annotation|:   38.097 ns (0 allocations: 0 bytes)
 - with inbounds: 33.217 ns (0 allocations: 0 bytes)
 - with inbounds + make assoc_values bigger: 79.305 ns (0 allocations: 0 bytes)
 - with inbounds + prechecking: 33.465 ns (0 allocations: 0 bytes)
 
 
 Looks like we have a winner with the prechecking since that is safe.
 I will push that shortly

oxinabox: - with inbounds and assoc to 127: 23.236 ns (0 allocations: 0 bytes) (Not safe)
 - with inbounds and checking based on bitmask: 47.035 ns (0 allocations: 0 bytes)
 - with inbounds and checking based in <: 36.359 ns (0 allocations: 0 bytes)
 - with inbounds and assoc to 255: 44.585 ns (0 allocations: 0 bytes)

oxinabox: Hmm i don't trust my benchmarks
Timing for 127 + checking based on <: sometimes is 36ns sometimes is 48ns.

I think we should do that though, and i will push that code

Issue #342, opened 2021-05-22
UndefRefError: Race Condition parsing TimeZone from multiple threads: needs coordination


NHDaly:
The `TIME_ZONE_CACHE` is not currently thread safe, so parsing a `TimeZone` object via the following constructor is currently a race condition, and can lead to crashes and errors:

https://github.com/JuliaTime/TimeZones.jl/blob/906cea08c2ca4a5c76b12e14fb60d5df7990fbf7/src/types/timezone.jl#L1

https://github.com/JuliaTime/TimeZones.jl/blob/906cea08c2ca4a5c76b12e14fb60d5df7990fbf7/src/types/timezone.jl#L43-L47

The cache either needs to be surrounded by a lock (though this would serialize all TimeZone constructors and probably be worse than having no cache at all), or replicated via thread-local caches.

I would probably suggest a thread-local cache? But interested in other ideas! :)

Maybe just something like this?:
```julia
const THREADLOCAL_TIME_ZONE_CACHES = Vector{Dict{String,Tuple{TimeZone,Class}}}()

function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
    tz, class = get!(THREADLOCAL_TIME_ZONE_CACHES[Threads.threadid()], str) do
    end
end

function __init__()
    append!(THREADLOCAL_TIME_ZONE_CACHES, [Dict{String,Tuple{TimeZone,Class}}() for _ in 1:Threads.nthreads()]
end
```

KristofferC:
As a note, it is generally better to allocate the object on the thread that it will be used (see https://github.com/JuliaWeb/HTTP.jl/pull/704 and https://github.com/JuliaLang/julia/blob/e4fcdf5b04fd9751ce48b0afc700330475b42443/stdlib/Random/src/RNGs.jl#L369-L385 for examples).

omus:
Thanks for filing the issue and the suggestions. Feel free to make a PR, otherwise I'll try to get to this during the week.

NHDaly:
K, this is definitely not done, but i started a PR here:
https://github.com/JuliaTime/TimeZones.jl/pull/344

@omus  maybe you can pick it up from here? ❤️ 

🙏 @KristofferC thanks for the examples, that's good to see. 👍

PR #344, opened 2021-06-01
Make Time Zone Cache thread-safe with lazily-allocated thread-local caches

NHDaly:
Make Time Zone Cache thread-safe with lazily-allocated thread-local caches. 
Each OS-thread has its own thread-local Time Zone Cache, which makes it thread-safe to construct TimeZones simultaneously from multiple threads.

Fixes https://github.com/JuliaTime/TimeZones.jl/issues/342.

------

I coded this through the GitHub UI during a meeting... so I haven't tested it locally. Sorry about that, but I just wanted to get this started. :)


src/types/timezone.jl
```diff
@@ -1,4 +1,30 @@
...
+    return TZ_CACHE
+end
+@noinline _tz_cache_length_assert() =  @assert false "0 < tid <= length(THREADLOCAL_TIME_ZONE_CACHES)"
+
+module __InitThreadLocalCaches  # This module is only here to house the __init__() function.
+using ..TimeZones: THREADLOCAL_TIME_ZONE_CACHES
+function __init__()
+    resize!(empty!(THREADLOCAL_TIME_ZONE_CACHES), Threads.nthreads()) # ensures that we didn't save a bad object
+end
+end
```
NHDaly, line 25: This pattern is a bit unconventional, and i don't know if there are any negative consequences to it?

There's already an `__init__()` for the module, but it seems annoying to put the cache initialization there. So i created a new module just to house the `__init__` closer to the definition.

Definitely happy to move this - lemme know what you prefer :)

NHDaly (commented):
Okay @omus  i think this is actually ready to review!

I've gotten the tests to pass locally. :) Thanks in advance!

src/types/timezone.jl
omus: Probably a cleaner compromise would be to call a `_tz_cache_init` function defined here from root `__init__`

src/types/timezone.jl
```diff
@@ -1,4 +1,30 @@
...
+        @inbounds THREADLOCAL_TIME_ZONE_CACHES[tid] = TZ_CACHE
+    end
+    return TZ_CACHE
+end
+@noinline _tz_cache_length_assert() =  @assert false "0 < tid <= length(THREADLOCAL_TIME_ZONE_CACHES)"
+
+module __InitThreadLocalCaches  # This module is only here to house the __init__() function.
+using ..TimeZones: THREADLOCAL_TIME_ZONE_CACHES
+function __init__()
+    resize!(empty!(THREADLOCAL_TIME_ZONE_CACHES), Threads.nthreads()) # ensures that we didn't save a bad object
```
omus, line 23: Why not just pre-allocated the empty dicts as part of the initialization?

omus (commented):
It would be great to make a test to ensure we don't break thread-safety again

src/types/timezone.jl
NHDaly: Yeah that's what I proposed originally in my comment, too, but @kristofferC said not to in his comment here:
https://github.com/JuliaTime/TimeZones.jl/issues/342#issuecomment-848501176

He says it's better to allocate the object on the thread where it will be used. Maybe it's for like memory locality or something? 🤔 

So anyway, i copied from the examples he posted. @kristofferC: Did I copy it right?

src/types/timezone.jl
NHDaly: agreed - i'll change that.

src/types/timezone.jl
omus: The code you pasted did the allocation on one thread. Something like this should be fine:

```julia
@threads for i in 1:nthreads()
    THREADLOCAL_TIME_ZONE_CACHES[i] = Dict{String,Tuple{TimeZone,Class}}()
end

**Comment by NHDaly** [comment]
> It would be great to make a test to ensure we don't break thread-safety again

Agreed, but i couldn't figure out how to write a test like that. It seems to be a pretty nondeterministic failure, and I didn't manage to get it to reproduce in a small unit test. Open to suggestions from others!!
@comnik: Are you able to suggest anything?

**Comment by NHDaly** [comment]
Okay, I've added a small test that fails about 75% of the time for me, and takes around 10seconds to run. Hopefully that's an acceptable tradeoff! You might have some ideas for how to increase the likelihood of failure since you know the code better, @omus

File: src/types/timezone.jl (hunk 2 above)
Comment (NHDaly) [reply src/types/timezone.jl:22]: Oh, interesting. Seems reasonable. In @kristofferC's example, his PR was explicitly titled "lazily allocate the threaded buffers and allocate them on the thread that will access it," so i figured there might be benefit to both aspects? What you suggest seems cleaner to me, too.

**Comment by omus** [comment]
Thanks! I'll try to do some performance testing against #345 tonight

File: src/types/timezone.jl [hunk 3]
```diff
@@ -1,4 +1,28 @@
-const TIME_ZONE_CACHE = Dict{String,Tuple{TimeZone,Class}}()
+# ---------- Time Zone Cache setup ----------------------------------------------------------------
+# Thread-local timezone caches to support multithreaded tasks simultaneously constructing TimeZones.
+# See https://github.com/JuliaTime/TimeZones.jl/issues/342 for more information.
+
+const THREADLOCAL_TIME_ZONE_CACHES = Dict{String,Tuple{TimeZone,Class}}[]
+
+@inline default_tz_cache() = default_tz_cache(Threads.threadid())
+@noinline function default_tz_cache(tid::Int)
```
Sacha0, line 8: Nice! :)

src/types/timezone.jl
```diff
@@ -1,4 +1,28 @@
...
+    0 < tid <= length(THREADLOCAL_TIME_ZONE_CACHES) || _tz_cache_length_assert()
+    if @inbounds isassigned(THREADLOCAL_TIME_ZONE_CACHES, tid)
+        @inbounds TZ_CACHE = THREADLOCAL_TIME_ZONE_CACHES[tid]
+    else
+        TZ_CACHE = eltype(THREADLOCAL_TIME_ZONE_CACHES)()
+        @inbounds THREADLOCAL_TIME_ZONE_CACHES[tid] = TZ_CACHE
+    end
+    return TZ_CACHE
+end
+@noinline _tz_cache_length_assert() =  @assert false "0 < tid <= length(THREADLOCAL_TIME_ZONE_CACHES)"
```
Sacha0, line 18: Also nice outlining of the error path! :)

Sacha0 (approved):
This looks great! Thanks Nathan! :)

@omus, did your testing bear out happily? :)

omus:
> @omus, did your testing bear out happily? :)

Slowly working on them. I have found a [readers-writer lock](https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock) is working pretty well without adding the extra storage requirements introduced here

omus:
I'm good to go ahead with this implementation. Performance comparison can be found [here](https://github.com/JuliaTime/TimeZones.jl/pull/345#issuecomment-858248784). I had other implementations but they were also slower so I left them out. Just need some refactoring here and we can proceed.

src/types/timezone.jl
omus: After more thought the reason they are probably doing the lazy allocation is to avoid latency during package startup

test/thread-safety.jl
```diff
@@ -0,0 +1,51 @@
...
+
+# Test that every Task produced the same result
+allsame(x) = all(y -> y == first(x), x)
+@test allsame(tzs)
+"""
+
+@info "Running Thread Safety tests"
+@testset "Multithreaded TimeZone construction" begin
+    run(`$(Base.julia_cmd()) -t8 --proj -E $(program)`)
+end
```
omus, line 51: Did you manage to produce a failure with this test with the non-thread safe code?

src/types/timezone.jl
```diff
@@ -1,4 +1,28 @@
...
+
+# Called from __init__() to initialize the Thread Local caches at runtime.
+function _tz_cache_init()
+    resize!(empty!(THREADLOCAL_TIME_ZONE_CACHES), Threads.nthreads()) # ensures that we didn't save a bad object
+end
+
+# ---------- End Time Zone Cache setup ------------------------------------------------------------
```
omus, line 25:
```suggestion
const THREAD_TZ_CACHES = Dict{String,Tuple{TimeZone,Class}}[]

# Based upon the thread-safe Global RNG implementation in the Random stdlib:
# https://github.com/JuliaLang/julia/blob/e4fcdf5b04fd9751ce48b0afc700330475b42443/stdlib/Random/src/RNGs.jl#L369-L385
@inline _tz_cache() = _tz_cache(Threads.threadid())
@noinline function _tz_cache(tid::Int)
    0 < tid <= length(THREAD_TZ_CACHES) || _tz_cache_length_assert()
    if @inbounds isassigned(THREAD_TZ_CACHES, tid)
        @inbounds cache = THREAD_TZ_CACHES[tid]
    else
        cache = eltype(THREAD_TZ_CACHES)()
        @inbounds THREAD_TZ_CACHES[tid] = cache
    end
    return cache
end
@noinline _tz_cache_length_assert() = @assert false "0 < tid <= length(THREAD_TZ_CACHES)"

function _reset_tz_cache()
    # ensures that we didn't save a bad object
    resize!(empty!(THREAD_TZ_CACHES), Threads.nthreads())
end
```

src/tzdata/compile.jl
```diff
@@ -694,7 +694,7 @@ function compile(tz_source::TZSource, dest_dir::AbstractString; kwargs...)
     results = compile(tz_source; kwargs...)
 
     isdir(dest_dir) || error("Destination directory doesn't exist")
-    empty!(TIME_ZONE_CACHE)
+    empty!(default_tz_cache())
```
omus, line 697: Re-running compile should invalidate the caches on all threads. I'm wondering though if this is still thread-safe then...
```suggestion
    TimeZones._reset_tz_cache()
```

src/types/timezone.jl
omus: Note there's a reference to `src/build.jl` that still refeences `TIME_ZONE_CACHE`

test/types/timezone.jl
```diff
@@ -2,7 +2,7 @@ using TimeZones: Class
 
 @testset "istimezone" begin
     # Invalidate the cache to ensure that `istimezone` works for non-loaded time zones.
-    empty!(TimeZones.TIME_ZONE_CACHE)
+    empty!(TimeZones.default_tz_cache())
```
omus, line 5: Should be safe as the tests aren't threaded
```suggestion
    TimeZones._reset_tz_cache()
```

omus (changes requested):

test/thread-safety.jl
Sacha0: Possibly relevant: https://github.com/JuliaTime/TimeZones.jl/pull/344#issuecomment-854067027 :)

src/tzdata/compile.jl
Sacha0: It seems like this is the sole outstanding comment, and otherwise this pull request is in good shape? :)

NHDaly: Yeah, i don't know what to do about this... @omus is right that this would no longer be thread safe, and i don't know how to make it thread-safe without _also_ introducing a lock around every access to the tz_cache, which defeats the whole point of this PR, and in which case something more like #345 might make more sense.

@omus: Can you explain more about why this function is invalidating the existing cache? In what way does it make the previously computed values invalid?

Sacha0: Hi gentlemen! What's the path forward here? @omus, any thoughts on Nathan's question just above? :)

NHDaly: I've asked about this line, here:
https://github.com/JuliaTime/TimeZones.jl/commit/182b6599a2dfede2114e74ce4817599ee086935a#r55083604

NHDaly:
Okay! :) Thanks @omus - please take another look. I've resolved the `compile()` cache clearing, and it was actually pretty straightforward 👍 Thanks!

src/tzdata/compile.jl
NHDaly: Done. Thanks!

src/build.jl
```diff
@@ -1,3 +1,4 @@
+using TimeZones: _reset_tz_cache
```
omus, line 1: Unnecessary as this is the same module
```suggestion
```

src/types/timezone.jl
```diff
@@ -1,4 +1,28 @@
-const TIME_ZONE_CACHE = Dict{String,Tuple{TimeZone,Class}}()
+# Thread-local TimeZone cache, which caches time zones _per thread_, allowing thread-safe
+# caching. Note that this means the cache will grow in size, and may store redundant objects
+# accross multiple threads, but this extra space usage allows for fast, lock-free access
+# to the cache, while still being thread-safe.
+const THREAD_TZ_CACHES = Dict{String,Tuple{TimeZone,Class}}[]
```
omus, line 5: Mainly for me, I just want to make it very obvious this now a vector:
```suggestion
const THREAD_TZ_CACHES = Vector{Dict{String,Tuple{TimeZone,Class}}}()
```

omus (commented):
Minor changes. Mainly waiting on CI results

src/build.jl
NHDaly: oh nice, thanks

src/types/timezone.jl
NHDaly: Awesome, yeah, thanks 👍

test/runtests.jl
```diff
@@ -72,6 +72,7 @@ include("helpers.jl")
     include("rounding.jl")
     include("parse.jl")
     include("plotting.jl")
+    include("thread-safety.jl")
```
omus, line 75:
```suggestion
    VERSION >= v"1.3" && include("thread-safety.jl")
```

omus:
I think we'll need to do some special casing for Julia 1.0 support

omus (approved):
Will merge in a couple of hours and tag a release

omus:
I was waiting to tag this due to https://github.com/JuliaTime/TimeZones.jl/pull/356 but I think I'll go ahead and make a release anyway. We can always make more :)

NHDaly:
> I was waiting to tag this due to #356 but I think I'll go ahead and make a release anyway. We can always make more :)

Yeah agreed, that sounds like the right move to me, too! :)

NHDaly:
Thanks Curtis! :)

Sacha0:
🎉 ! :)

Issue #343, opened 2021-05-26
Package artifacts and distribute using our standard architecture


ViralBShah:
I wonder if it would be more robust to have the artifacts packaged up as a JLL and distributed like other binary artifacts.

omus:
The artifacts used by this package are just the tzdata archives provided by IANA. Currently TimeZones.jl extracts the tzsource files, compiles this information into the internal TimeZones.jl representation, and the serializes that information to disk during package build time. If we introduce a JLL I think it makes the most sense to store the serialized internal representation as the JLL artifact. Mainly the only thing holding up this work is picking a serialization method that works across Julia versions.

KristofferC:
One thing that could be done is to do the compilation in `__init__` instead and use https://github.com/JuliaPackaging/Scratch.jl to store the resulting files (instead of storing them inside the package folder). That would make the package work with e.g. PackageCompiler.

StefanKarpinski:
There's not really any reason for this to be a JLL, is there? It's not platform specific and doesn't require a fancy cross-compilation toolchain, it's just an artifact with a build step, right?

ViralBShah:
I meant "bundle it in a standard way and distribute using our infrastructure, and make it more packagecompiler friendly".

StefanKarpinski:
Might it be good enough to include a script that can be run when making a release that downloads the timezone files, processes them, bundles them up the way the package wants them, pushes that to GitHub along with the release and generates the corresponding `Artifacts.toml` file? That way the user can just download ready-to-use artifacts that are already in the right format?

staticfloat:
Is the problem that the serialization format is Julia-version-specific?

ViralBShah:
Yes: https://github.com/JuliaTime/TimeZones.jl/issues/343#issuecomment-848793417

DilumAluthge:
Can it be serialized to e.g. JSON? And then we can use JSON3 and StructTypes to deserialize it directly back to Julia types?

KristofferC:
Use Scratch space as a cache and serialize it during runtime to a folder based on the julia version?

ViralBShah:
Seems like we should do that right away. I do wonder if different artifacts can be made for every Julia version.

KristofferC:
You can just prefix the scratch folder with the julia version for example.

omus:
[...]

PR #347, opened 2021-06-09
Use `Scratch` for cached timezone data.

### LINKED ISSUE #338: Avoid defining paths at compile time
Such as https://github.com/JuliaTime/TimeZones.jl/blob/781e59a2075a24651d6ae70491c4b87392640e51/src/TimeZones.jl#L33-L34. It makes the package incompatible with PackageCompiler.

MichaelHatherly:
Fixes #338, based on https://github.com/JuliaTime/TimeZones.jl/issues/343#issuecomment-848823981 by KristofferC.

At module init runs the current `build` to generate
the cached data, and stores it in a scratchspace instead of the
deps folder. Associated globals that were defined at top-level are
now set in `__init__` (and `_init` for submodules) to allow for
`PackageCompiler` support. Scratchspace is linked to the Julia and
package version so should rebuild when those change for users.

The `deps/build.jl` now just imports the package which triggers
a build of the cache if needed so this shouldn't really impact
`using TimeZones` much as far as I could tell.

A required PR for `Scratch` to expand it's version support is at
https://github.com/JuliaPackaging/Scratch.jl/pull/24. I've enabled
support going back to 1.0 for that package to match what `TimeZones`
says it supports, but it turns out `EzXML` is bounded at `1.3` so
I couldn't test down to 1.0-1.2.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,1 +1,3 @@
-using TimeZones
+using TimeZones: build
+
+build()
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,3 +1,1 @@
-using TimeZones: build
-
-build()
+using TimeZones
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -50,7 +50,7 @@ const PKG_VERSION = VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__), "Pr
 const DEPS_DIR = Ref{String}()
 
 function __init__()
-    DEPS_DIR[] = @get_scratch!("timezones-$PKG_VERSION")
+    DEPS_DIR[] = @get_scratch!("timezones-$VERSION-$PKG_VERSION-$(TZData.tzdata_version())")
     isdir(DEPS_DIR[]) || mkpath(DEPS_DIR[])
     TZData._init()
     @static Sys.iswindows() && WindowsTimeZoneIDs._init()
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -54,12 +54,6 @@ function __init__()
     TZData._init()
     @static Sys.iswindows() && WindowsTimeZoneIDs._init()
     if !isdir(joinpath(DEPS_DIR, "compiled", string(VERSION)))
-        cd(DEPS_DIR) do
-            for (path, contents) in DEPS_CONTENTS
-                mkpath(dirname(path))
-                write(path, contents)
-            end
-        end
         build()
     end
 
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -47,13 +47,14 @@ end
 abstract type Local <: TimeZone end
 
 const PKG_VERSION = VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__), "Project.toml"))["version"])
+const DEPS_DIR = Ref{String}()
 
 function __init__()
-    global DEPS_DIR = @get_scratch!("timezones-$PKG_VERSION")
-    isdir(DEPS_DIR) || mkpath(DEPS_DIR)
+    DEPS_DIR[] = @get_scratch!("timezones-$PKG_VERSION")
+    isdir(DEPS_DIR[]) || mkpath(DEPS_DIR[])
     TZData._init()
     @static Sys.iswindows() && WindowsTimeZoneIDs._init()
-    if !isdir(joinpath(DEPS_DIR, "compiled", string(VERSION)))
+    if !isdir(joinpath(DEPS_DIR[], "compiled", string(VERSION)))
         build()
     end
 
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -49,10 +49,11 @@ abstract type Local <: TimeZone end
 const PKG_VERSION = VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__), "Project.toml"))["version"])
 
 function __init__()
-    global DEPS_DIR = @get_scratch!("timezones-$VERSION-$PKG_VERSION")
+    global DEPS_DIR = @get_scratch!("timezones-$PKG_VERSION")
+    isdir(DEPS_DIR) || mkpath(DEPS_DIR)
     TZData._init()
-    Sys.iswindows() && WindowsTimeZoneIDs._init()
-    if isempty(readdir(DEPS_DIR))
+    @static Sys.iswindows() && WindowsTimeZoneIDs._init()
+    if !isdir(joinpath(DEPS_DIR, "compiled", string(VERSION)))
         cd(DEPS_DIR) do
             for (path, contents) in DEPS_CONTENTS
                 mkpath(dirname(path))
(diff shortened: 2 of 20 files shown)


src/TimeZones.jl
```diff
@@ -30,14 +32,36 @@ export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDate
...
 # TimeZone types used to disambiguate the context of a DateTime
 # abstract type UTC <: TimeZone end  # Already defined in the Dates stdlib
 abstract type Local <: TimeZone end
 
+const pkg_version = VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__), "Project.toml"))["version"])
+
 function __init__()
+    global DEPS_DIR = @get_scratch!("timezones-$VERSION-$pkg_version")
+    TZData._init()
+    Sys.iswindows() && WindowsTimeZoneIDs._init()
```
omus, line 54:
```suggestion
    @static Sys.iswindows() && WindowsTimeZoneIDs._init()
```

src/TimeZones.jl
```diff
@@ -30,14 +32,36 @@ export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDate
...
+    if isempty(readdir(DEPS_DIR))
+        cd(DEPS_DIR) do
+            for (path, contents) in DEPS_CONTENTS
+                mkpath(dirname(path))
+                write(path, contents)
+            end
+        end
+        build()
+    end
```
omus, line 63: I suspect it would be cleaner to just have the build process write the scratch directory?

src/TimeZones.jl
```diff
@@ -30,14 +32,36 @@ export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDate
...
+        include_dependency(joinpath(deps, path))
+        DEPS_CONTENTS[path] = read(joinpath(deps, path), String)
+    end
+end
 
 # TimeZone types used to disambiguate the context of a DateTime
 # abstract type UTC <: TimeZone end  # Already defined in the Dates stdlib
 abstract type Local <: TimeZone end
 
+const pkg_version = VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__), "Project.toml"))["version"])
```
omus, line 49:
```suggestion
const PKG_VERSION = VersionNumber(TOML.parsefile(joinpath(dirname(@__DIR__), "Project.toml"))["version"])
```

omus (commented):
Thanks for this!

src/TimeZones.jl
```diff
@@ -30,14 +32,36 @@ export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDate
...
+
 function __init__()
+    global DEPS_DIR = @get_scratch!("timezones-$VERSION-$pkg_version")
```
omus, line 52: May be able to use the tzdata version

MichaelHatherly: Can do, is the cache folder structure likely to remain the same between package versions? If not then using just `tzdata_version` as the scratch key may cause issues, perhaps.

src/TimeZones.jl
MichaelHatherly: Yes, I'll move it into `build()` instead.

src/TimeZones.jl
MichaelHatherly: Const-prop should handle that one fine I'd assume and remove the branching by itself. I can add a `@static` anyway if it's preferred though.

omus:
This change will require us to drop support for Julia versions below 1.5. That's not a deal breaker but annoying as I want to continue to support the 1.0 LTS while it's the official LTS. The changes for dropping 1.0 support can be done in another PR. We'll need to at least disable the 1.0 CI jobs though so we can validate the changes.

MichaelHatherly:
I'd like to retain LTS support as well, hence https://github.com/JuliaPackaging/Scratch.jl/pull/24 should be completed before finishing off this one, so that we can just use a version of `Scratch` that supports Julia `1+`.

MichaelHatherly:
@omus the CI failure is a bit of a mystery to me.. I've adjusted some paths and checks so maybe it'll get by this time.

src/tzdata/TZData.jl
```diff
@@ -14,11 +14,17 @@ end
...
 
-const ARCHIVE_DIR = joinpath(DEPS_DIR, "tzarchive")
-const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
-const COMPILED_DIR = joinpath(DEPS_DIR, "compiled", string(VERSION))
+function _init()
+    global ARCHIVE_DIR = joinpath(TimeZones.DEPS_DIR, "tzarchive")
+    global TZ_SOURCE_DIR = joinpath(TimeZones.DEPS_DIR, "tzsource")
+    global COMPILED_DIR = joinpath(TimeZones.DEPS_DIR, "compiled", string(VERSION))
+    global ACTIVE_VERSION_FILE = joinpath(TimeZones.DEPS_DIR, "active_version")
+    global LATEST_FILE = joinpath(TimeZones.DEPS_DIR, "latest")
```
omus, line 22: I think I'd prefer if these were all global constants and we used `Ref{String}()`

src/tzdata/build.jl
```diff
@@ -43,11 +43,7 @@ function build(
             # Retrieve the current latest version the cached latest has expired
             if latest_version === nothing
                 latest_version = last(tzdata_versions())
-                tzdata_hash = artifact_hash("tzdata$latest_version", ARTIFACT_TOML)
-
-                if tzdata_hash === nothing
-                    error("Latest tzdata is $latest_version which is not present in the Artifacts.toml")
-                end
+                @artifact_str "tzdata$latest_version"
```
omus, line 46: Did this need to be changed?

deps/build.jl
```diff
@@ -1,3 +1 @@
-using TimeZones: build
-
-build()
+using TimeZones
```
omus, line 1: Hmmm, if someone manually runs `] build` we should probably re-build the scratch directory.

src/TimeZones.jl
omus: It's probably safest to include Julia version, package version, and tzdata version. It also just occurred to me that users are allowed to [re-compile the time zones to support dates in the far future](https://juliatime.github.io/TimeZones.jl/stable/faq/#future_tzs-1). I'll need to think more about this.

src/tzdata/build.jl
MichaelHatherly: It should have the same effect of throwing an error when the particular artifact does not exist, but by using `@artifact_str` the `Artifact.toml` contents gets embedded rather than looked up at runtime in the file, which may not exist if it's been relocated.

src/tzdata/TZData.jl
MichaelHatherly: Are any of these constants regarded as public? It'll of course be a breaking change if that's the case unfortunately.

andreasnoack:
Are there any unaddressed comments here or is this good to go?

MichaelHatherly:
> Are there any unaddressed comments here or is this good to go?

The failure in benchmarks is the current blocker here, still trying to pin it down.

MichaelHatherly:
Closing in favour of #348.

Issue #350, opened 2021-07-05
Should UNICODE_CLDR_VERSION be bumped?


LightAndLight:
A new Unicode CLDR version (`release-38-1`) was added to Artefacts.toml in [this commit](https://github.com/JuliaTime/TimeZones.jl/commit/faa208e8a7c4f6f25b9b50668eacc03a4b3e0900) but `UNICODE_CLDR_VERSION` ([here](https://github.com/JuliaTime/TimeZones.jl/blob/master/src/winzone/WindowsTimeZoneIDs.jl#L10)) remains at `"release-37"`.

Should `UNICODE_CLDR_VERSION` be bumped?

omus:
Yes it should be updated. I'll note `release-39` is out so we'll just jump to that.

PR #351, opened 2021-07-05
Update to Unicode CLDR 39 / Use specified CLDR version

omus:
Fixes https://github.com/JuliaTime/TimeZones.jl/issues/350. During updating the CLDR release I noticed we were still embedding the Windows translation files were still begin embedded since https://github.com/JuliaTime/TimeZones.jl/pull/49. Since the issue with throttling had to do with downloading directly from unicode.org we should no longer have this problem since we're downloading from GitHub.


src/winzone/WindowsTimeZoneIDs.jl
```diff
@@ -40,37 +50,20 @@ function compile(xml_file::AbstractString)
...
-
-    if !isfile(xml_file)
-        if isfile(fallback_xml_file) && !force
-            cp(fallback_xml_file, xml_file)
+    if !isfile(xml_file) || force
+        @info "Downloading Windows to POSIX timezone ID XML version: $UNICODE_CLDR_VERSION"
+        @static if VERSION >= v"1.3"
+            artifact_dir = @artifact_str "unicode-cldr-$UNICODE_CLDR_VERSION"
+            xml_file = joinpath(artifact_dir, WINDOWS_ZONE_FILE)
+            cp(xml_file, WINDOWS_XML_FILE)
```
omus, line 59:
```suggestion
            cp(xml_file, WINDOWS_XML_FILE, force=true)
```

PR #354, opened 2021-08-10
WIP: Make FixedTimeZone isbits

Wynand:
This switches FixedTimeZone.name to an InlineString instead of a String, and doing that makes it an isbits type

As a side effect this also limits the number of bytes in a FixedTimeZone name

Usually timezones are named after locations, and the longest location name currently is Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu.  This is in New Zealand, so if a fixed time zone were created for it then it would be `Oceania/Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu`

That is still only 93 characters, so the 255 character limit should be fine for a while, and we could switch to InlineString127 if needed as well

Related issue: https://github.com/JuliaTime/TimeZones.jl/issues/271


Wynand:
Depends on: https://github.com/JuliaData/WeakRefStrings.jl/pull/78

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.5.6"
+version = "1.6.0"
```
Wynand, line 4: A bug fix might be more appropriate

omus:
Thanks, this seems promising. Out of curiosity do you also want `VariableTimeZone` to be an `isbitstype` or is your priority just `FixedTimeZone`?

Wynand:
> Thanks, this seems promising. Out of curiosity do you also want `VariableTimeZone` to be an `isbitstype` or is your priority just `FixedTimeZone`?

No problem!  

For now I'm focusing on `FixedTimeZone` since `VariableTimeZone` still has the `transitions` vector, however `Transition` will also be an `isbitstype` once `FixedTimeZone` is an `isbitstype`, which should speed up garbage collection

nickrobinson251:
> That is still only 93 characters, so the 255 character limit should be fine for a while, and we could switch to InlineString127 if needed as well

I think we should probably switch to `InlineString127`
Having every `FixedTimeZone` take up an extra 128 bytes unnecessarily is gonna add up (especially compared to the current size e.g. `sizeof(tz"UTC") == 24`)

I also wonder if there's anyway to use even shorter InlineStrings... 🤔

omus:
> For now I'm focusing on `FixedTimeZone` since `VariableTimeZone` still has the `transitions` vector, however `Transition` will also be an `isbitstype` once `FixedTimeZone` is an `isbitstype`, which should speed up garbage collection

Great! I think using `InlineString`s is a good stepping stone towards that goal.

> I also wonder if there's anyway to use even shorter InlineStrings... 🤔

Funny you should ask... As an experiment I implemented a `TZAbbr` string-type which should be minimize the storage cost for the `Transition` vector: https://github.com/JuliaTime/TimeZones.jl/pull/355

I still think we should proceed with using `InlineString` as proposed in this PR as the `TZAbbr` is more specialized and will take some additional work to integrate.

Wynand:
Worth noting: This will cause issues with projects using CSV.jl until it's next release, since v0.8.5 doesn't support WeakRefStrings 1.2

omus:
> Worth noting: This will cause issues with projects using CSV.jl until it's next release, since v0.8.5 doesn't support WeakRefStrings 1.2

It's doubtful anyone will explicitly require this latest release so falling back to a slightly older TimeZones.jl should be fine

Project.toml
omus: A minor release is appropriate. This also leaves room for backporting bugfixes

src/types/fixedtimezone.jl
```diff
@@ -95,3 +95,20 @@ end
 
 name(tz::FixedTimeZone) = tz.name
 rename(tz::FixedTimeZone, name::AbstractString) = FixedTimeZone(name, tz.offset)
+
+function Base.:(==)(a::FixedTimeZone, b::FixedTimeZone)
+    a.name == b.name && a.offset == b.offset
```
omus, line 100:
```suggestion
    return a.name == b.name && a.offset == b.offset
```

src/types/fixedtimezone.jl
```diff
@@ -95,3 +95,20 @@ end
...
+end
+
+function Base.isequal(a::FixedTimeZone, b::FixedTimeZone)
+    return (
+        isequal(a.name, b.name) &&
+        isequal(a.offset, b.offset)
+    )
```
omus, line 107:
```suggestion
    return isequal(a.name, b.name) && isequal(a.offset, b.offset)
```

src/types/fixedtimezone.jl
```diff
@@ -95,3 +95,20 @@ end
...
+end
+
+function Base.hash(tz::FixedTimeZone, h::UInt)
+    h = hash(tz.name, h)
+    h = hash(tz.offset, h)
+    return h
+end
```
omus, line 114: Including the `Symbol` in the hash ensures the resulting hash differs from other structs which happen to have the same fields.
```suggestion
    h = hash(:timezone, h)
    h = hash(tz.name, h)
    h = hash(tz.offset, h)
    return h
end

```

src/types/timezone.jl
```diff
@@ -43,13 +43,13 @@ TimeZone(::AbstractString, ::Class)
 function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
```
omus, line 43: Let's avoid having to to the conversion multiple times:
```suggestion
TimeZone(str::AbstractString, args...) = TimeZone(String(str), args...)

function TimeZone(str::String, mask::Class=Class(:DEFAULT))
```

omus:
WeakRefStrings.jl has a Julia minimum of 1.3. All of its dependencies are set to use Julia 1 and the PR that bumped up the minimum was: https://github.com/JuliaData/WeakRefStrings.jl/pull/73. From looking over that PR it seems like WeakRefStrings.jl could be set to use `julia = "1"`

Wynand:
Reran with `test.jl` as
```julia
using TimeZones, BenchmarkTools
BenchmarkTools.DEFAULT_PARAMETERS.gcsample = true
BenchmarkTools.DEFAULT_PARAMETERS.seconds = 300
BenchmarkTools.DEFAULT_PARAMETERS.samples = 100


function bm_tzs()
    map( t -> TimeZone(t, TimeZones.Class(:ALL)), timezone_names())
    TimeZones._reset_tz_cache()
    GC.gc()
end

@benchmark bm_tzs()

```

---
Master results:
```
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  278.357 ms … 318.445 ms  ┊ GC (min … max): 23.32% … 19.60%
 Time  (median):     289.653 ms               ┊ GC (median):    22.14%
 Time  (mean ± σ):   291.489 ms ±   8.360 ms  ┊ GC (mean ± σ):  22.02% ±  0.65%

     ▃▁▆ █  ▃ ▆ ▁▃▆▁ ▃   ▁  ▁▃    ▁▁ ▃ ▁                         
  ▄▁▄███▇█▁▄█▇█▇████▇█▄▄▇█▇▄██▇▇▄▁██▄█▁█▄▁▄▇▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▄ ▄
  278 ms           Histogram: frequency by time          318 ms <

 Memory estimate: 45.01 MiB, allocs estimate: 843644.
```
this branch's results:
```
BenchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):  94.168 ms … 100.176 ms  ┊ GC (min … max): 58.06% … 56.36%
 Time  (median):     96.990 ms               ┊ GC (median):    57.91%
 Time  (mean ± σ):   96.926 ms ±   1.017 ms  ┊ GC (mean ± σ):  58.00% ±  0.62%

                     ▂▆  ▆  ▄▂   ▄ █ ▆▂                         
  ▄▁▁▁▄▁▁▁▁▁▆▁▁▆▁▆▆▄▄███▆██▁██▆▆████▄██▆▄█▆▆▆▄▆▄▄▄▁▁▁▁▄▁▄▁▁▁▁▄ ▄
  94.2 ms         Histogram: frequency by time         99.7 ms <

 Memory estimate: 9.38 MiB, allocs estimate: 54047.
```

src/types/fixedtimezone.jl
```diff
@@ -95,3 +95,18 @@ end
...
+function Base.isequal(a::FixedTimeZone, b::FixedTimeZone)
+    return isequal(a.name, b.name) && isequal(a.offset, b.offset)
+end
+
+function Base.hash(tz::FixedTimeZone, h::UInt)
+    h = hash(:timezone, h)
+    h = hash(tz.name, h)
+    h = hash(tz.offset, h)
+    return h
+end
```
oxinabox, line 112: Why do we need to overload == and hash?

I expected making it isbits would decrease that need.

src/types/timezone.jl
```diff
@@ -64,16 +64,18 @@ US/Pacific (UTC-8/UTC-7)
...
 
-function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
+TimeZone(str::AbstractString, args...) = TimeZone(String(str), args...)
+
+function TimeZone(str::String, mask::Class=Class(:DEFAULT))
     # Note: If the class `mask` does not match the time zone we'll still load the
     # information into the cache to ensure the result is consistent.
-    tz, class = get!(_tz_cache(), str) do
-        tz_path = joinpath(TZData.COMPILED_DIR, split(str, "/")...)
+    tz, class = get!(_tz_cache(), String(str)) do
```
oxinabox, line 72: Is this required?

src/types/timezone.jl
```diff
@@ -64,16 +64,18 @@ US/Pacific (UTC-8/UTC-7)
...
+        tz_path = joinpath(TZData.COMPILED_DIR, split(String(str), "/")...)
```
oxinabox, line 73: An issue should be openned on InlineStrings.jl if this is required

src/types/timezone.jl
```diff
@@ -64,16 +64,18 @@ US/Pacific (UTC-8/UTC-7)
...
 
         if isfile(tz_path)
             open(deserialize, tz_path, "r")
-        elseif occursin(FIXED_TIME_ZONE_REGEX, str)
-            FixedTimeZone(str), Class(:FIXED)
+        elseif occursin(FIXED_TIME_ZONE_REGEX, String(str))
+            FixedTimeZone(String(str)), Class(:FIXED)
```
oxinabox, line 78: Is this required?
It seems like it is just going to be converted to a InlineString anyway so why convert to a String first?

oxinabox:
> Usually timezones are named after locations, and the longest location name currently is Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu. This is in New Zealand, so if a fixed time zone were created for it then it would be Oceania/Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu

> That is still only 93 characters, so the 255 character limit should be fine for a while, and we could switch to InlineString127 if needed as well


I think we can do better.

VariableTimeZones are named after locations.
FixedTimeZones have codes like UTC+0800
See the FIXED_TIMEZONE_REGEX.

I found before they all fit a ShortString15.

Wynand:
> I found before they all fit a ShortString15

Switched to InlineString15, here are the results:

```
enchmarkTools.Trial: 100 samples with 1 evaluation.
 Range (min … max):   97.166 ms … 198.882 ms  ┊ GC (min … max): 60.96% … 29.90%
 Time  (median):     105.430 ms               ┊ GC (median):    60.54%
 Time  (mean ± σ):   108.828 ms ±  13.095 ms  ┊ GC (mean ± σ):  60.96% ±  3.95%

  ▁▃██▃▃▆▆ ▃   ▆ ▃   ▁▁▁▃                ▁                       
  ████████▇█▇▇▄█▄█▇▇▄████▄▁▄▁▄▁▁▄▁▁▇▁▄▁▁▄█▁▇▁▄▇▁▁▄▄▇▁▁▁▄▁▁▁▁▁▁▇ ▄
  97.2 ms          Histogram: frequency by time          135 ms <

 Memory estimate: 5.04 MiB, allocs estimate: 53858.
```

src/types/timezone.jl
Wynand: Removed pretty much all the changes that cast to strings, seems like it's handled better now

src/types/timezone.jl
Wynand: removed

src/types/timezone.jl
Wynand: removed

Wynand:
I can't think of any more code changes or cleanup needed, so if everything looks good this can be merged

src/types/fixedtimezone.jl
```diff
@@ -95,3 +95,8 @@ end
 
 name(tz::FixedTimeZone) = tz.name
 rename(tz::FixedTimeZone, name::AbstractString) = FixedTimeZone(name, tz.offset)
+
+function Base.isequal(a::FixedTimeZone, b::FixedTimeZone)
+    return isequal(a.name, b.name) && isequal(a.offset, b.offset)
+end
```
oxinabox, line 101: Why do we need to overload this?
(sorry I didn't mention it before)

Wynand:
> Looks like this PR effectively lower-bounds Julia to 1.3 through the InlineStrings dependency, so it seems it has the same problems as the old WeakRefStrings. This means that CI won't run as-is.

I've switched to ShortStrings instead, which does support Julia 1.0.  The only quirk so far is with regex matching, but I've also only tested locally with Julia 1.6

oxinabox:
> I'm okay with that. We should drop 1.0 support in a separate PR and be sure to remove all of the VERSION specific code.

I will make that PR presently

PR #356, opened 2021-08-19
Reduce memory & redundant work for concurrent TimeZones construction

NHDaly:
This commit improves the thread-local caching scheme introduced in
#344, by sharing TimeZones
across _all_ thread-local caches (so there's only ever one TimeZone
object instance created per process), and reduces redundant work caused
by multiple concurrent Tasks starting to construct the same TimeZone at
the same time.

Before this commit, multiple Tasks that try to construct the same
TimeZone on different threads would have constructed the same object
multiple times. Whereas now, they would share it, via the "Future".

------------------------------------------------------------

It's difficult to show this effect, but one way to show it is by logging
when a TimeZone is constructed, via this diff:
```julia
diff --git a/src/types/timezone.jl b/src/types/timezone.jl
index 25d36c3..1cea69e 100644
--- a/src/types/timezone.jl
+++ b/src/types/timezone.jl
@@ -68,6 +68,7 @@ function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
     tz, class = get!(_tz_cache(), str) do
+        Core.println("CONSTRUCTING $str")
         tz_path = joinpath(TZData.COMPILED_DIR, split(str, "/")...)

         if isfile(tz_path)
```

Before this commit, you can see that every thread constructs the object
twice - once before the clear and once after (total of 8 times):
```julia
julia> Threads.nthreads()
4

julia> TimeZones.TZData.compile();

julia> foo() = begin
           @sync for i in 1:20
               if (i == 10)
                   @info "---------clear-------"
                   TimeZones.TZData.compile()
               end
               Threads.@Spawn begin
                   TimeZone("US/Eastern", TimeZones.Class(:LEGACY))
               end
           end
           @info "done"
       end
foo (generic function with 1 method)

julia> @time foo()
[ Info: ---------clear-------
CONSTRUCTING US/Eastern
CONSTRUCTING US/Eastern
CONSTRUCTING US/Eastern
CONSTRUCTING US/Eastern
CONSTRUCTING US/Eastern
CONSTRUCTING US/Eastern
CONSTRUCTING US/Eastern
CONSTRUCTING US/Eastern
[ Info: done
  0.391298 seconds (1.51 M allocations: 64.544 MiB, 2.46% gc time, 0.00% compilation time)
```

After this commit, it's constructed only twice - once before the clear
and once after:
```julia
julia> @time foo()
[ Info: ---------clear-------
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: done
  0.414059 seconds (1.46 M allocations: 61.972 MiB, 4.55% gc time)
```

------------------------------------------------------------------

Finally, the other problem this avoids is if we ever accidentally
introduce a Task yield inside the constructor, which could happen if we
used `@info` instead of `Core.println()`, then without this PR, the old
code could potentially do _redundant work_ to construct the TimeZone
multiple times - even on the same thread - since each Task's constructor
would see that there's no TZ in the cache, start the work, and then
yield to the next Task, which would do the same, until finally they all
report their work into the cache, overwriting each other.

This is what happens if we use `@info` in the above diff, instead:

Before this commit - the results are nondeterministic, but in this run,
you can see it redundantly constructed the value all 20 times!:
```julia
julia> @time foo()
[ Info: ---------clear-------
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: done
  0.494492 seconds (1.55 M allocations: 66.754 MiB, 16.67% gc time)
```

After this commit, just the two we expect. 😊 :
```julia
julia> @time foo()
[ Info: ---------clear-------
[ Info: CONSTRUCTING US/Eastern
[ Info: CONSTRUCTING US/Eastern
[ Info: done
  0.422677 seconds (1.47 M allocations: 62.228 MiB, 4.66% gc time)
```


NHDaly:
Sorry the diff is hard to read; it'll be easier after https://github.com/JuliaTime/TimeZones.jl/pull/344/ is merged.

The entirety of this PR is just in this single commit, here: bd68cf9
You can view the diff by clicking that. :)

----------

I think this PR is maybe worth merging separately, since it's nice to see the performance difference and keep it separate from the correctness of the previous PR. Also, we'll want to measure the CPU perf impact to make sure that it's okay before merging this one. I'm not sure how you were measuring it, @omus - can you run that measurement again, for consistency? Thanks!

----------

CC: @sacha0, @omus.

src/types/timezone.jl
```diff
@@ -19,10 +25,22 @@ const THREAD_TZ_CACHES = Vector{Dict{String,Tuple{TimeZone,Class}}}()
...
 end
+# ensures that we didn't save a bad object
+function _reset_tz_cache()
+    # Since we use thread-local caches, we spawn a task on _each thread_ to clear that
+    # thread's local cache.
+    Threads.@threads for i in 1:Threads.nthreads()
+        @assert Threads.threadid() === i "TimeZones.TZData.compile() must be called from the main, top-level Task."
+        empty!(_tz_cache())
+    end
+    @lock tz_cache_mutex begin
```
omus, line 39: Using the function `lock` may avoid the compat issue

NHDaly: Yeah, that's a good suggestion. We've been using `Base.@lock` instead internally, to avoid the potential overhead from the closure..

And indeed, we need to mutate a local variable inside the locked body down below:
https://github.com/JuliaTime/TimeZones.jl/blob/6c0d3ead7640f7818b7cbea9dd3596423f73c225/src/types/timezone.jl#L94-L103

So using `Base.lock(l) do ...` would end up allocating on every call. 😕 👎 😞 


So I'd probably opt to keep it how it currently is. Do you have any preference on that? Should I file a PR to add `Base.@lock` to compat?

Ref: https://github.com/JuliaLang/julia/issues/36441 and https://github.com/JuliaLang/julia/pull/39588

src/types/timezone.jl
```diff
@@ -4,6 +4,12 @@
 # to the cache, while still being thread-safe.
 const THREAD_TZ_CACHES = Vector{Dict{String,Tuple{TimeZone,Class}}}()
 
+# Holding a lock during construction of a specific TimeZone prevents multiple Tasks (on the
+# same or different threads) from attempting to construct the same TimeZone object, and
+# allows them all to share the result.
+const tz_cache_mutex = ReentrantLock()
```
omus, line 10: Let's keep our constants consistent
```suggestion
const TZ_CACHE_MUTEX = ReentrantLock()
```

src/types/timezone.jl
omus: Lets add `@lock` to Compat.jl so everyone can use this

src/types/timezone.jl
```diff
@@ -19,10 +25,22 @@ const THREAD_TZ_CACHES = Vector{Dict{String,Tuple{TimeZone,Class}}}()
 end
 @noinline _tz_cache_length_assert() = @assert false "0 < tid <= length(THREAD_TZ_CACHES)"
 
-function _reset_tz_cache()
-    # ensures that we didn't save a bad object
+function _tz_cache_init()
```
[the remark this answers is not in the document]
omus: Lets do this

src/types/timezone.jl
```diff
@@ -68,20 +86,40 @@ function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
...
+        future = @lock tz_cache_mutex begin
+            get!(TZ_CACHE_FUTURES, str) do
+                constructing = true
+                Channel{Tuple{TimeZone,Class}}(1)
+            end
+        end
+        if constructing
+            tz_path = joinpath(TZData.COMPILED_DIR, split(str, "/")...)
+
+            tz = if isfile(tz_path)
```
omus, line 106: Returned value is a tuple not just a timezone
```suggestion
            t = if isfile(tz_path)
```

omus (commented):
Sorry about being slow to review. Overall this is good just needs a couple of touch ups.

Just wanted to call out that the `TimeZone` mutex changes is essentially equivalent to a readers/writers lock similar to this implementation: https://github.com/JuliaTime/TimeZones.jl/pull/345/commits/6522f7a00a2eb21ce8772ed6ac5ea437f41cbc06. I'm just calling this out as a possible cleanup to this implementation in the future if a well tested readers/writer lock package emerges.

src/types/timezone.jl
NHDaly: Thanks @omus, I've filed https://github.com/JuliaLang/Compat.jl/issues/762 to track that. I can probably do it myself, but don't want to block this PR since it's already taken me so long to get back to it! 😅 😊

src/types/timezone.jl
NHDaly: Done. thanks!

src/types/timezone.jl
NHDaly: Oh, but actually, even better: now that you've lower-bounded to julia 1.6+, i can drop this compat thing entirely. 💪 🎉

NHDaly (commented):
Thanks for the review, @omus! Sorry it was me who took so long to get back to it this time 😁 

Hope you've had a nice holiday! ⭐

NHDaly:
> Just wanted to call out that the TimeZone mutex changes is essentially equivalent to a readers/writers lock similar to this implementation: 6522f7a. I'm just calling this out as a possible cleanup to this implementation in the future if a well tested readers/writer lock package emerges.

FWIW, i still think this will be more performant than a reader/writer lock could be, since this kind of cache locking takes advantage of the cache being append-only, so that you don't need to lock _at all_ on cache hits. I could imagine we could separate out this kind of thread-safe shared cache into a separate package though, so that others could benefit from this implementation 👍

src/types/timezone.jl
NHDaly: huh, i guess i forgot to push after my last commit. Updated now, finally. Resolving

src/types/timezone.jl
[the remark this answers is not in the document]
NHDaly: finally pushed, resolving

omus:
Kick starting CI

src/types/timezone.jl
```diff
@@ -68,20 +86,40 @@ function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
...
+            elseif occursin(FIXED_TIME_ZONE_REGEX, str)
+                FixedTimeZone(str), Class(:FIXED)
+            elseif !isdir(TZData.COMPILED_DIR) || isempty(readdir(TZData.COMPILED_DIR))
+                # Note: Julia 1.0 supresses the build logs which can hide issues in time zone
+                # compliation which result in no tzdata time zones being available.
+                throw(ArgumentError(
+                    "Unable to find time zone \"$str\". Try running `TimeZones.build()`."
+                ))
+            else
+                throw(ArgumentError("Unknown time zone \"$str\""))
```
omus, line 117: Exceptions while constructing will cause threads to be blocked upon waiting for a channel that will never be populated

omus (commented):
There definitely seems to be an issue with this implementation as the CI consistently stalls out and I see the same thing when running the tests locally. Some additional work needs to be done here before we can proceed

omus:
@NHDaly sorry for not giving this the attention it deserves. Been swamped lately. I'll try to be faster on additional reviews

NHDaly:
No worries at all - i fully believe that there are bugs in the implementation here.

We actually split this out into its own package, here: https://github.com/NHDaly/MultiThreadedCaches.jl

We found several bugs in the implementation while extracting it out, so the code is much cleaner now.  I think this PR can wait until we scrub out all this and use that package instead! :)

It's also worth doing another comparison against a Read/Write Lock and also against a ConcurrentHashTable.

Feel free to review that package directly, if you'd like - most of the code was added in: https://github.com/NHDaly/MultiThreadedCaches.jl/pull/1, plus a couple follow ups.

Thanks for your patience @omus, sorry for the delay on my end. ALSO i'm about to go on vacation for three weeks, so 👋 we can talk about this again soon! 😅

PR #360, opened 2021-10-07
Rebuild TZData before benchmarks

Wynand:
This recompiles TZData before running benchmarks to ensure benchmarks still run when the serialized TZdata is changed


Wynand:
This should resolve the benchmarking issue: https://github.com/JuliaTime/TimeZones.jl/pull/360

I've tested running the CI code locally, replicated the issue, and fixed it by using this code as the base.  I'll add more details on how I tested it in this MR

PR #367, opened 2021-11-10
Update ZonedDateTime constructor to clarify behaviour when from_utc=false

iamed2:
I got caught by this and almost assumed something incorrect about the downstream CloudWatchLogs.jl library. The "instead of in local time" clause implied to me that when `from_utc=false` the `DateTime` would be _converted_ from `localzone()` into the specified `TimeZone`, but that is not the case.

### DIFF
--- a/src/types/zoneddatetime.jl
+++ b/src/types/zoneddatetime.jl
@@ -28,9 +28,10 @@ end
     ZonedDateTime(dt::DateTime, tz::TimeZone; from_utc=false) -> ZonedDateTime
 
 Construct a `ZonedDateTime` by applying a `TimeZone` to a `DateTime`. When the `from_utc`
-keyword is true the given `DateTime` is assumed to be in UTC instead of in local time and is
-converted to the specified `TimeZone`.  Note that when `from_utc` is true the given
-`DateTime` will always exists and is never ambiguous.
+keyword is true the given `DateTime` is assumed to be in UTC and is converted to the specified
+`TimeZone`.  When `from_utc` is false the given `DateTime` is assumed to already be in the
+specified `TimeZone`. Note that when `from_utc` is true the given `DateTime` will always exist
+and is never ambiguous.
 """
 function ZonedDateTime(dt::DateTime, tz::VariableTimeZone; from_utc::Bool=false)
     # Note: Using a function barrier which reduces allocations


omus:
>  The "instead of in local time" clause implied to me that when `from_utc=false` the `DateTime` would be _converted_ from `localzone()` into the specified `TimeZone`, but that is not the case.

Definitely seems to be a perspective issue. The original documentation was written with the idea that it was known that internally a UTC datetime is stored. The primary reason `from_utc` is for performance reasons so using the term "convert" could been interpreted that using `from_utc=true` is slower.

src/types/zoneddatetime.jl
```diff
@@ -28,9 +28,10 @@ end
     ZonedDateTime(dt::DateTime, tz::TimeZone; from_utc=false) -> ZonedDateTime
 
 Construct a `ZonedDateTime` by applying a `TimeZone` to a `DateTime`. When the `from_utc`
-keyword is true the given `DateTime` is assumed to be in UTC instead of in local time and is
-converted to the specified `TimeZone`.  Note that when `from_utc` is true the given
-`DateTime` will always exists and is never ambiguous.
+keyword is true the given `DateTime` is assumed to be in UTC and is converted to the specified
+`TimeZone`.  When `from_utc` is false the given `DateTime` is assumed to already be in the
+specified `TimeZone`. Note that when `from_utc` is true the given `DateTime` will always exist
+and is never ambiguous.
```
omus, line 34: Better? 
```suggestion
keyword is `true` the given `DateTime` is declared to be in UTC but will be displayed in the specified
`TimeZone`.  When `from_utc` is `false` the given `DateTime` is declared to be in local time, or in the
specified `TimeZone`. Note that when `from_utc` is `true` the provided UTC `DateTime` will always exist
and is never ambiguous.
```

iamed2: I think "will be displayed" leaks the abstraction; to the user, they don't need to know that the underlying storage is in UTC. I can see why you would want to avoid "converted" though. I'm struggling to come up with good wording.

I think the phrase "local time" should be avoided to avoid confusion with `localzone()`/the OS time zone.

omus: > I think "will be displayed" leaks the abstraction; to the user, they don't need to know that the underlying storage is in UTC. I can see why you would want to avoid "converted" though. I'm struggling to come up with good wording.

I'm definitely up for suggestions. I think this part of the documentation has been hard to get right.

> I think the phrase "local time" should be avoided to avoid confusion with `localzone()`/the OS time zone.

The phrase "local time" is something the [IANA documentation](https://data.iana.org/time-zones/tz-how-to.html) uses and isn't quite perfect as really it just refers to only the time component. Maybe "local timestamp" or "local `DateTime`" is an improvement.

I suspect we'll have to iterate on this a little bit. Thank you for opening the PR :)

PR #369, opened 2021-12-02
Set minimum supported version to Julia 1.6

omus:
Now that [Julia 1.6 is the LTS](https://discourse.julialang.org/t/julia-v1-7-0-has-been-released-and-v1-6-is-now-lts/72336) we can drop support for earlier versions of Julia.


omus:
I'll merge this tomorrow unless there are objections

PR #370, opened 2021-12-02
Benchmark failures should not cancel other jobs

omus:
Replaces: https://github.com/JuliaTime/TimeZones.jl/pull/328

> When set to `true`, GitHub cancels all in-progress jobs if any `matrix` job fails. Default: `true`

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

By setting `fail-fast: false` on the benchmark job then a failure there does not cancel the other in-progress jobs but other jobs with `fail-fast: true` can cause the benchmark to be cancelled.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -75,7 +75,7 @@ jobs:
     name: Benchmarks
     runs-on: ubuntu-latest
     strategy:
-      fast-fail: false
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


omus:
My issue was that it's `fail-fast` and not `fast-fail` 🤦

PR #384, opened 2022-07-07
Use `Scratch` to hold downloads and compiled tzdata

staticfloat:
This should address many of the problems listed in #343.  It still contains a `Pkg.build()` step to do the serialization, but it stores things in a scratch space, rather than in the package directory, and it marks the default tzdata as an eager artifact, rather than a lazy one, so it should download along with the rest of the artifacts during package installation.

The one thing I'm not sure about is whether the "version" that it checks against the IANA servers is properly cached; I see many different ways that the version can be queried, and I'm not sure if the "happy path" still makes a request against the IANA servers.  I would prefer that it doesn't, and no network accesses outside of the normal Pkg flow occur, as that helps this package be used in restrictive environments where only an internal PkgServer is whitelisted.


staticfloat:
As a side note, I did experiment with eliminating the local serialization step completely; I used `JLD2` as a substitute, which seems to work, although it does add a few KB to each serialized file.

We could have a GitHub Actions job that automatically serializes all tzdata sets and uploads them to a GitHub release or similar, and opens a pull request if a new tzdata set is added without the paired serialized artifact.  If someone wants to do this, I'd be happy to guide them through it, but I can't do it myself at the moment.

omus:
Thanks for this @staticfloat. I'll give this a thorough review tomorrow but I'm already sure I'd like to pair this with the TZJFile format introduced in #380.

src/tzdata/build.jl
```diff
@@ -36,32 +36,33 @@ function build(
 
     artifact_dir = @artifact_str "tzdata$version"
 
-    if !isempty(tz_source_dir)
+    if source_dir !== nothing
```
omus, line 39:
```suggestion
    if tz_source_dir !== nothing
```

src/tzdata/build.jl
```diff
@@ -36,32 +36,33 @@ function build(
...
         @info "Installing $version tzdata region data"
         regions = union!(intersect(regions, readdir(artifact_dir)), CUSTOM_REGIONS)
         for region in setdiff(regions, CUSTOM_REGIONS)
-            cp(joinpath(artifact_dir, region), joinpath(tz_source_dir, region), force=true)
+            cp(joinpath(artifact_dir, region), joinpath(source_dir, region), force=true)
```
omus, line 43:
```suggestion
            cp(joinpath(artifact_dir, region), joinpath(tz_source_dir, region), force=true)
```

src/tzdata/build.jl
```diff
@@ -36,32 +36,33 @@ function build(
...
+        for region in CUSTOM_REGIONS
+            cp(joinpath(dirname(dirname(@__DIR__)), "deps", "custom_tzsource_regions", region),
+               joinpath(source_dir, region), force=true)
         end
     end
-    if !isempty(compiled_dir)
+    if compiled_dir !== nothing
         @info "Converting tz source files into TimeZone data"
-        tz_source = TZSource(joinpath.(tz_source_dir, regions))
+        tz_source = TZSource(joinpath.(source_dir, regions))
```
omus, line 53:
```suggestion
        tz_source = TZSource(joinpath.(tz_source_dir, regions))
```

omus (commented):
Overall seems good to me. I've modified some of the names to be more consistent. e.g. mixing "compiled"/"serialized"/"serialized cache" or "tz_source"/"source". I'll mention at first I was thinking using `Ref` constants for the directories was the correct approach but the use of functions reduces the chance of mutations there so I think it's the right direction. I'll make a follow up PR to rename "custom_tzsource_regions" and I think we can merge this. I'll integrate defaulting to the TZJFile format before making a release though

omus:
I should call out that this could be a breaking change if people were referencing the now removed constants. I'm not too worried about this but I may use some `global`s defined at in `__init__` to ensure this is non-breaking.

src/tzdata/TZData.jl
```diff
@@ -2,18 +2,18 @@ module TZData
...
+using ...TimeZones: scratch_dir
 
 # Note: The tz database is made up of two parts: code and data. TimeZones.jl only requires
 # the "tzdata" archive or more specifically the "tz source" files within the archive
 # (africa, australasia, ...)
 
-const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
-const COMPILED_DIR = joinpath(DEPS_DIR, "compiled", string(VERSION))
+tz_source_dir() = scratch_dir("tzsource")
+serialized_cache_dir() = scratch_dir("serialized", string(VERSION))
```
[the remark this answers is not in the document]
omus: ❗ We'll need to use different directories for each tzdata version. Otherwise having multiple versions of TimeZones installed at the same time with different tzdata versions will interfere with each other

omus:
> ❗ We'll need to use different directories for each tzdata version. Otherwise having multiple versions of TimeZones installed at the same time with different tzdata versions will interfere with each other
– https://github.com/JuliaTime/TimeZones.jl/pull/384#discussion_r916476283

With this in mind I think it makes sense to merge #385 first as that will eliminate us having to specify the Julia version number to ensure serialization compatibility.

omus:
> I'll mention at first I was thinking using `Ref` constants for the directories was the correct approach but the use of functions reduces the chance of mutations there so I think it's the right direction

After sleeping on this there may be an advantage in the `Ref` approach under certain conditions. For example if we continue to allow specifying the tzdata version used via the env variable `JULIA_TZ_VERSION` we may want to reference the path to the tzdata version loaded into memory currently instead just reflecting the current `JULIA_TZ_VERSION` value. Using a `Ref` would allow us to update this directory under specific conditions only.

src/tzdata/build.jl
```diff
@@ -36,14 +36,19 @@ function build(
...
+    if tz_source_dir !== nothing
         @info "Installing $version tzdata region data"
         regions = union!(intersect(regions, readdir(artifact_dir)), CUSTOM_REGIONS)
         for region in setdiff(regions, CUSTOM_REGIONS)
             cp(joinpath(artifact_dir, region), joinpath(tz_source_dir, region), force=true)
         end
+        # Copy over our 'custom regions' from `deps/custom_tzsource_regions`
+        custom_tz_source_dir = joinpath(dirname(dirname(@__DIR__)), "deps", "custom_tzsource_regions")
+        for region in CUSTOM_REGIONS
+            cp(joinpath(custom_tz_source_dir, region), joinpath(source_dir, region), force=true)
```
omus, line 48:
```suggestion
            cp(joinpath(custom_tz_source_dir, region), joinpath(tz_source_dir, region), force=true)
```

src/tzdata/compile.jl
```diff
@@ -719,7 +719,6 @@ function compile(tz_source::TZSource, dest_dir::AbstractString; kwargs...)
 end
 
 # TODO: Deprecate?
-function compile(tz_source_dir::AbstractString=TZ_SOURCE_DIR, dest_dir::AbstractString=COMPILED_DIR; kwargs...)
-    tz_source_paths = joinpath.(tz_source_dir, readdir(tz_source_dir))
-    compile(TZSource(tz_source_paths), dest_dir; kwargs...)
+function compile(source_dir::AbstractString=tz_source_dir(), dest_dir::AbstractString=compiled_dir(); kwargs...)
+    compile(TZSource(readdir(source_dir; join=true)), dest_dir; kwargs...)
```
omus, line 723:
```suggestion
function compile(tz_source_dir::AbstractString=tz_source_dir(), dest_dir::AbstractString=compiled_dir(); kwargs...)
    compile(TZSource(readdir(tz_source_dir; join=true)), dest_dir; kwargs...)
```

omus:
Alright in order to get things moving I'll merge this as it exists now and follow up on the issues I pointed out in a follow up PR.

src/tzdata/build.jl
```diff
@@ -21,9 +22,8 @@ const REGIONS = [STANDARD_REGIONS; LEGACY_REGIONS]
 function build(
     version::AbstractString,
     regions::AbstractVector{<:AbstractString},
-    tz_source_dir::AbstractString="",
-    compiled_dir::AbstractString="";
-    verbose::Bool=false,
+    tz_source_dir::Union{AbstractString,Nothing}=nothing,
+    compiled_dir::Union{AbstractString,Nothing}=nothing,
```
omus, line 26:
```suggestion
    tz_source_dir::AbstractString="",
    compiled_dir::AbstractString="",
```

src/tzdata/build.jl
```diff
@@ -36,14 +36,19 @@ function build(
 
     artifact_dir = @artifact_str "tzdata$version"
 
-    if !isempty(tz_source_dir)
+    if tz_source_dir !== nothing
```
omus, line 39:
```suggestion
    # TODO: Deprecate skipping tzdata installation step or use `nothing` instead
    if !isempty(tz_source_dir)
```

src/tzdata/build.jl
```diff
@@ -36,14 +36,19 @@ function build(
...
             cp(joinpath(artifact_dir, region), joinpath(tz_source_dir, region), force=true)
         end
+        # Copy over our 'custom regions' from `deps/custom_tzsource_regions`
+        custom_tz_source_dir = joinpath(dirname(dirname(@__DIR__)), "deps", "custom_tzsource_regions")
+        for region in CUSTOM_REGIONS
+            cp(joinpath(custom_tz_source_dir, region), joinpath(tz_source_dir, region), force=true)
+        end
     end
-    if !isempty(compiled_dir)
+    if compiled_dir !== nothing
```
omus, line 51:
```suggestion

    # TODO: Deprecate skipping conversion step or use `nothing` instead
    if !isempty(compiled_dir)
```

omus (commented):
Although I agree with using `nothing` for this purpose this change could be breaking and doesn't seem worth it at this time. I've left a comment to address this in the future.

omus:
This PR is also running into #386. Looking into this before merging here

omus:
> This PR is also running into #386. Looking into this before merging here

Failure definitely appears unrelated.

omus (approved):
There are a few follow up changes I'd like to make but I think the fastest path forward is to merge this PR as it exists and make the additional changes in another PR

PR #385, opened 2022-07-08
Switch to using TZJFile serialization format

omus:
Utilizes the changes introduced in #380. We should definitely pair this with scratch spaces like in #384

### DIFF
--- a/src/tzdata/TZData.jl
+++ b/src/tzdata/TZData.jl
@@ -2,14 +2,14 @@ module TZData
 
 using LazyArtifacts
 using Printf
-using ...TimeZones: DEPS_DIR
+using ...TimeZones: TZJFile, DEPS_DIR
 
 # Note: The tz database is made up of two parts: code and data. TimeZones.jl only requires
 # the "tzdata" archive or more specifically the "tz source" files within the archive
 # (africa, australasia, ...)
 
 const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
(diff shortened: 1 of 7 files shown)


ViralBShah:
Turns out to be quite a yak shave

src/tzdata/TZData.jl
```diff
@@ -2,14 +2,14 @@ module TZData
...
-using ...TimeZones: DEPS_DIR
+using ...TimeZones: TZJFile, DEPS_DIR
 
 # Note: The tz database is made up of two parts: code and data. TimeZones.jl only requires
 # the "tzdata" archive or more specifically the "tz source" files within the archive
 # (africa, australasia, ...)
 
 const TZ_SOURCE_DIR = joinpath(DEPS_DIR, "tzsource")
-const COMPILED_DIR = joinpath(DEPS_DIR, "compiled", string(VERSION))
+const COMPILED_DIR = joinpath(DEPS_DIR, "compiled", "tzjf")
```
omus, line 12: We should specify the tzjfile version as part of this path. Mainly this only matters when we start to use scratch spaces and have multiple tzjfile versions.

As we write the `DEFAULT_VERSION` I'll just reference that here but I can see a future iteration being more explicit about the compiled version being build and the directory it's output to.

omus:
In my opinion I think we should merge this prior to #384. Mainly because this avoids every populating the scratch directory with the old serialized files.

Issue #386, opened 2022-07-08
Stochastic failure in thread-safety tests


omus:
Noticed this failure on "Julia 1.6 - windows-latest - x64". I'm not seeing the source of the failure so it seemed best to track this in an issue:
```
 Testing Running tests...
[ Info: Installing 2016j tzdata region data
[ Info: Downloading Windows to POSIX timezone ID XML version: release-40
[ Info: Compiling Windows time zone name translation
[ Info: Running Thread Safety tests
Multithreaded TimeZone construction: Error During Test at D:\a\TimeZones.jl\TimeZones.jl\test\thread-safety.jl:65
  Got exception outside of a @test
  failed process: Process(`'C:\hostedtoolcache\windows\julia\1.6.6\x64\bin\julia.exe'-Cnative'-JC:\hostedtoolcache\windows\julia\1.6.6\x64\lib\julia\sys.dll'--depwarn=yes--check-bounds=yes-g1--code-coverage=user--color=yes--startup-file=no--proj-E"using TimeZones
using Test
[ Info: Installing 2016j tzdata region data
[ Info: Converting tz source files into TimeZone data
@assert Threads.nthreads() > 1 \"This system does not support multiple threads, so the thread-safety tests cannot be run.\"
@testset \"Multithreaded TimeZone brute force test\" begin
    function create_zdt(year, month, day, tz_name)
        ZonedDateTime(DateTime(year, month, day), TimeZone(tz_name))
    end
    function cycle_zdts()
        return [
            try
                create_zdt(year, month, day, tz_name)
            catch e
                # Ignore ZonedDateTimes that aren't valid
                e isa Union{ArgumentError,AmbiguousTimeError,NonExistentTimeError} || rethrow()
                nothing
            end
            for year in 2000:2020
            for month in 1:5
            for day in 10:15
            for tz_name in timezone_names()
        ]
    end
    outputs = Channel(Inf)
    @sync begin
        for _ in 1:15
            Threads.@spawn begin
                put!(outputs, cycle_zdts())
            end
        end
    end
    close(outputs)
    tzs = collect(outputs)
    # Test that every Task produced the same result
    allsame(x) = all(y -> y == first(x), x)
    @test allsame(tzs)
end
#----------------------------------------------------
@testset \"Interleaved compile() and TimeZone construction\" begin
    @sync for i in 1:20
        if (i % 5 == 0)
            TimeZones.TZData.compile()
        end
        Threads.@spawn begin
[...]

PR #387, opened 2022-07-08
Run thread-safety checks when Julia is using threads

omus:
Digging into #386. Seems like these failures are less random than they first appeared. Running the thread tests in a separate Julia process result in stack traces being lost and only allow for speculative debugging. By enabling threads on the CI and only running the tests when threads are enabled we should be able to get useful information to address the root of the problem.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -32,6 +32,8 @@ jobs:
             arch: x86
           - os: windows-latest
             arch: x86
+    env:
+      JULIA_NUM_THREADS: auto
     steps:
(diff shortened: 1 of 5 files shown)


omus:
I've subverted the thread-safe cache to validate that these thread-safety tests still fail with this change.

omus:
Tests still pass even without the thread-safe cache. I have managed to produce a thread-safety failure locally so I think the best way to proceed is to merge this PR (with the proper thread-safe cache) and hopefully I can get some additional information in regards to #386

PR #388, opened 2022-07-09
Minor follow up changes to #384

omus:
Some minor changes to #384 which were easiest done with a follow up PR.


src/tzdata/build.jl
```diff
@@ -39,21 +39,19 @@ function build(
...
+        for path in region_paths
+            cp(path, joinpath(tz_source_dir, basename(path)), force=true)
         end
     end
 
     # TODO: Deprecate skipping conversion step or use `nothing` instead
     if !isempty(compiled_dir)
         @info "Converting tz source files into TimeZone data"
-        tz_source = TZSource(joinpath.(tz_source_dir, regions))
+        tz_source = TZSource(readdir(tz_source_dir, join=true))
```
omus, line 54: There are some nuisances to the original code here I'm not sure I still agree with but I'll make some adjustments to keep the original behaviour. I'll be sure to add a detailed comment on the subtleties

Issue #371, opened 2022-01-06
Initialization Error in PackageCompiler app on Windows when relocating


Felix-Gauthier:
There seems to be an initialization error in applications generated through `PackageCompiler.jl` using `TimeZones.jl` as a dependency on Windows when relocating to another computer (i.e. sharing a quick CLI app with a not-yet-converted-to-glorious-julia-prog colleague)

Consider the following:

<details>
<summary>Foo module</summary>

```julia
module Foo

using TimeZones

function julia_main()::Cint
    try
        @info "Hello, World!"
    catch
        Base.invokelatest(Base.display_error, Base.catch_stack())
        return 1
    end
    return 0
end

end # module
```
</details>

Generating the app with `PackageCompiler v2.0.4` and `TimeZones v1.7.0` (n.b. I have not tested with lower versions of each pkg)
```julia
using PackageCompiler; create_app("Foo", "./FooCompiled"; force=true, include_lazy_artifacts=true)
```

I initially thought the error was because of not including lazy artifacts, but I've tested both ways to make sure with same results

<details>
<summary>Error stacktrace</summary>

```
fatal: error thrown and no exception handler available.
InitError(mod=:WindowsTimeZoneIDs, error=Base.IOError(msg="mkdir("C:\\Users\\user\\.julia\\packages\\TimeZones\\ZQ4lc\\deps\\local"; mode=0o777): no such file or directory (ENOENT)", code=-4058))
uv_error at .\libuv.jl:97 [inlined]
#mkdir#10 at .\file.jl:183
mkdir at .\file.jl:176 [inlined]
__init__ at C:\Users\user\.julia\packages\TimeZones\ZQ4lc\src\winzone\WindowsTimeZoneIDs.jl:19
jfptr___init___34981.clone_1 at C:\Users\user\Desktop\FooCompiled\FooCompiled\lib\julia\sys.dll (unknown line)
jl_apply at /cygdrive/c/buildbot/worker/package_win64/build/src\julia.h:1788 [inlined]
jl_module_run_initializer at /cygdrive/c/buildbot/worker/package_win64/build/src\toplevel.c:73
_finish_julia_init at /cygdrive/c/buildbot/worker/package_win64/build/src\init.c:796
jl_init_with_image at /cygdrive/c/buildbot/worker/package_win64/build/src\jlapi.c:74 [inlined]
[...]

PR #389, opened 2022-07-09
Include the tzdata version in our scratch directory structure

omus:
Follow up to #384. Since we're using a scratch directory multiple TimeZones.jl packages using different versions may concurrently be using the scratch directory. If these packages use different tzdata versions we could accidentally use a different tzdata version that what is typically used by that package. To solve this I've included the tzdata version in the scratch directory structure as well as included the tzjfile version. To make this work properly I ended up using `Ref`s to some of the directories so it's possible to update the tzdata version being used.

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -31,15 +31,22 @@ export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDate
     # ranges.jl
     guess
 
-# When we write out things like serialized tzdata representations,
-# do it into a scratchspace.
-scratch_dir(args...) = mkpath(joinpath(@get_scratch!(args[1]), args[2:end]...))
+_scratch_dir() = @get_scratch!("build")
+_tz_source_dir(version::AbstractString) = joinpath(_scratch_dir(), "tzsource", version)
+function _compiled_dir(version::AbstractString)
+    joinpath(_scratch_dir(), "compiled", "tzjf", "v$(TZJFile.DEFAULT_VERSION)", version)
+end
+
+const COMPILED_DIR = Ref{String}()
 
 # TimeZone types used to disambiguate the context of a DateTime
 # abstract type UTC <: TimeZone end  # Already defined in the Dates stdlib
 abstract type Local <: TimeZone end
 
 function __init__()
+    # Write out our compiled tzdata representations into a scratchspace
+    COMPILED_DIR[] = _compiled_dir(tzdata_version())
+
     # Initialize the thread-local TimeZone cache (issue #342)
     _reset_tz_cache()
 
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -37,7 +37,7 @@ function _compiled_dir(version::AbstractString)
     joinpath(_scratch_dir(), "compiled", "tzjf", "v$(TZJFile.DEFAULT_VERSION)", version)
 end
 
-const COMPILED_DIR = Ref{String}()
+const _COMPILED_DIR = Ref{String}()
 
 # TimeZone types used to disambiguate the context of a DateTime
 # abstract type UTC <: TimeZone end  # Already defined in the Dates stdlib
@@ -45,7 +45,7 @@ abstract type Local <: TimeZone end
 
 function __init__()
     # Write out our compiled tzdata representations into a scratchspace
-    COMPILED_DIR[] = _compiled_dir(tzdata_version())
+    _COMPILED_DIR[] = _compiled_dir(tzdata_version())
 
     # Initialize the thread-local TimeZone cache (issue #342)
     _reset_tz_cache()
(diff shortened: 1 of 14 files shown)


test/ci.jl
```diff
@@ -5,8 +5,8 @@ using TimeZones: TZData
 
 @testset "build process" begin
     # Clean out deps directories for a clean re-build
-    compiled_dir = TZData.compiled_dir()
-    tz_source_dir = TZData.tz_source_dir()
+    compiled_dir = TZData._compiled_dir(TZDATA_VERSION)
+    tz_source_dir = TZData._tz_source_dir(TZDATA_VERSION)
```
omus, line 9: I'll take another look at these tests tomorrow. I think there are some assumptions that are no longer true and possibly we can always run these tests instead of just in the CI environment

src/TimeZones.jl
```diff
@@ -43,6 +47,14 @@ function __init__()
     # Initialize the thread-local TimeZone cache (issue #342)
     _reset_tz_cache()
 
+    # Write out our compiled tzdata representations into a scratchspace
+    version = tzdata_version()
+    _COMPILED_DIR[] = _compiled_dir(version)
+
+    # Build the time zone data if required. Should only occur when `JULIA_TZ_VERSION` is
+    # specified we to a version not previously built.
+    isdir(_COMPILED_DIR[]) || build(version)
```
omus, line 56: Added this as otherwise a user who specifies `JULIA_TZ_VERSION` would be required to manually call build if they hadn't already built the tzdata version specified. As an additional benefit this should call build automatically in the cases where `deps/build.jl` is skipped. Possibly we could drop `deps/build.jl` entirely but that may not play nice with PackageCompiler.jl

omus:
Windows failure:
```
     Building TimeZones → `D:\a\TimeZones.jl\TimeZones.jl\deps\build.log`
[ Info: Installing 2016j tzdata region data
[ Info: Converting tz source files into TimeZone data
[ Info: Downloading Windows to POSIX timezone ID XML version: release-40
ERROR: LoadError: InitError: IOError: open("C:\\Users\\runneradmin\\.julia\\scratchspaces\\f269a46b-ccf7-5d73-abea-4c690281aa53\\build\\local\\windowsZones.xml", 769, 33206): no such file or directory (ENOENT)
Stacktrace:
  [1] uv_error
    @ .\libuv.jl:100 [inlined]
  [2] open(path::String, flags::UInt16, mode::UInt64)
    @ Base.Filesystem .\filesystem.jl:122
  [3] sendfile(src::String, dst::String)
    @ Base.Filesystem .\file.jl:994
  [4] cp(src::String, dst::String; force::Bool, follow_symlinks::Bool)
    @ Base.Filesystem .\file.jl:384
  [5] build(xml_file::String; force::Bool)
    @ TimeZones.WindowsTimeZoneIDs D:\a\TimeZones.jl\TimeZones.jl\src\winzone\WindowsTimeZoneIDs.jl:47
  [6] build(version::String; force::Bool)
    @ TimeZones D:\a\TimeZones.jl\TimeZones.jl\src\build.jl:14
  [7] build(version::String)
    @ TimeZones D:\a\TimeZones.jl\TimeZones.jl\src\build.jl:10
  [8] __init__()
    @ TimeZones D:\a\TimeZones.jl\TimeZones.jl\src\TimeZones.jl:56
...
```

src/TimeZones.jl
omus: In https://github.com/JuliaTime/TimeZones.jl/issues/371 we may have seen an issue with PackageCompiler not playing nice with directory creation occurring during the package init. However, the use of scratch spaces may not have that problem and typically directories will be created during `deps/build.jl`

omus: I'm very much of two minds about this:
1. Having this causes the time zone data to be built if it's needed. For most users this means they shouldn't see the "please run `TimeZones.build()`" message which is much nicer in automated environments
2. In the case of a build failure the TimeZones.jl package won't allow you load the package and investigate the failure

I think this choice is important enough that I'll make it into a separate PR

test/ci.jl
omus: Made a couple of small changes to the logic in these tests

src/TimeZones.jl
omus: https://github.com/JuliaTime/TimeZones.jl/pull/390

omus:
Not loving the coverage reduction but I don't think that should hold this up.

StefanKarpinski:
Seems better to use separate scratch spaces for different versions of the tzdata, no? That way the pkg gc can clean up no longer uesd tzdata; otherwise old ztdata will accumulate over time, taking up more and more space.

omus:
> Seems better to use separate scratch spaces for different versions of the tzdata, no? That way the pkg gc can clean up no longer uesd tzdata; otherwise old ztdata will accumulate over time, taking up more and more space.

Good point I wasn't really considering. I can make a follow up issue for that.

PR #390, opened 2022-07-11
Build at package initialization time (if required)

omus:
See https://github.com/JuliaTime/TimeZones.jl/pull/389#discussion_r917343619 for details. PR causes `TimeZones.build` to automatically be called if the desired tzdata version has not previously been built. This should:

- Result in end-users not having to ever run `TimeZones.build()` themselves
- Avoid having automated environments fail in scenarios where `deps/build.jl` was not triggered

The main concerns with this change is that it may not play nicely with PackageCompiler.jl and users may not like that artifact retrieval and some processing could occur at package initialization time.

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -44,11 +44,16 @@ const _COMPILED_DIR = Ref{String}()
 abstract type Local <: TimeZone end
 
 function __init__()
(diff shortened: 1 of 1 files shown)


omus:
I lost track of this PR. Was superseded by #441

PR #391, opened 2022-07-11
Limit imports

omus:
Limits what we import from Scratch.jl and LazyArtifacts.jl


omus:
Failure looks related. I'll look into this later

omus:
No longer needed

PR #413, opened 2022-11-08
Print a better error message when an assertion fails

DilumAluthge:


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.9.0"
+version = "1.10.0"
```
omus, line 4: All version changes for this repo are handled outside of PRs
```suggestion
version = "1.9.0"
```

src/types/timezone.jl
```diff
@@ -17,7 +17,13 @@ const THREAD_TZ_CACHES = Vector{Dict{String,Tuple{TimeZone,Class}}}()
...
     return cache
 end
-@noinline _tz_cache_length_assert() = @assert false "0 < tid <= length(THREAD_TZ_CACHES)"
+@noinline function _tz_cache_length_assert(; tid)
+    msg = "Assertion failed: 0 < tid <= length(THREAD_TZ_CACHES) where " *
+          "tid=$(tid) " *
+          "length(THREAD_TZ_CACHES)=$(length(THREAD_TZ_CACHES)) " *
+          "Threads.nthreads()=$(Threads.nthreads())"
+    throw(ErrorException(msg))
+end
```
omus, line 26: In what scenario did you see the assertion error?

DilumAluthge: I encountered it in PackageCompiler, when trying to build a system image in which TimeZones was one of the indirect dependencies. Unfortunately I wasn't able to get it down to a minimal reproducer.

DilumAluthge:
I don't think the original code exists anymore, so I think this PR is now stale.

PR #415, opened 2022-11-14
CI: test both with and without threading

DilumAluthge:


omus:
Any motivation behind this?

DilumAluthge:
Yeah, I have a hypothesis that the segfaults that we're seeing on Julia nightly only happen when threading is enabled.

omus:
We may want to start testing with threads yet but the original reason for this is no longer happening.

Issue #419, opened 2022-12-02
The time zone "Europe/Amsterdam" is of class `TimeZones.Class(:LEGACY)`


rikhuijzer:
I'm unable to run code which refers to `tz"Europe/Amsterdam"`:

```julia
julia> using TimeZones

julia> tz"Europe/Amsterdam"
ERROR: LoadError: ArgumentError: The time zone "Europe/Amsterdam" is of class `TimeZones.Class(:LEGACY)` which is currently not allowed by the mask: `TimeZones.Class(:FIXED) | TimeZones.Class(:STANDARD)`
Stacktrace:
 [1] TimeZone(str::String, mask::TimeZones.Class)
   @ TimeZones ~/.julia/packages/TimeZones/V28u7/src/types/timezone.jl:89
 [2] TimeZone(str::String)
   @ TimeZones ~/.julia/packages/TimeZones/V28u7/src/types/timezone.jl:70
 [3] var"@tz_str"(__source__::LineNumberNode, __module__::Module, str::Any)
   @ TimeZones ~/.julia/packages/TimeZones/V28u7/src/types/timezone.jl:110
in expression starting at REPL[7]:1
```

This is unexpected since `Europe/Amsterdam` is a commonly used timezone. The release notes of tz show no mention of the deprecation of the timezone, apart from the following note:

> Finish moving to 'backzone' the location-based zones whose
     timestamps since 1970 are duplicates; adjust links accordingly.
     This change ordinarily affects only pre-1970 timestamps, and with
     the new PACKRATLIST option it does not affect any timestamps.
     In this round the affected zones are Antarctica/Vostok,
     Asia/Brunei, Asia/Kuala_Lumpur, Atlantic/Reykjavik,
     Europe/Amsterdam, Europe/Copenhagen, Europe/Luxembourg,
     Europe/Monaco, Europe/Oslo, Europe/Stockholm, Indian/Christmas,
     Indian/Cocos, Indian/Kerguelen, Indian/Mahe, Indian/Reunion,
     Pacific/Chuuk, Pacific/Funafuti, Pacific/Majuro, Pacific/Pohnpei,
     Pacific/Wake and Pacific/Wallis, and the affected links are
     Arctic/Longyearbyen, Atlantic/Jan_Mayen, Iceland, Pacific/Ponape,
     Pacific/Truk, and Pacific/Yap.

[...]

PR #421, opened 2022-12-03
Use backzone file as 'standard' regions

tpgillam:
Pertaining to https://github.com/JuliaTime/TimeZones.jl/issues/419

Needs discussion!

I had to increase the length of the FixedTimeZone name, as there was at least one name in `backzone` of length 18 characters.

### DIFF
--- a/src/tzdata/build.jl
+++ b/src/tzdata/build.jl
@@ -1,7 +1,7 @@
 # The default tz source files we care about. See "ftp://ftp.iana.org/tz/data/Makefile"
 # "PRIMARY_YDATA" for listing of tz source files to include.
 const STANDARD_REGIONS = [
-    "africa", "antarctica", "asia", "australasia",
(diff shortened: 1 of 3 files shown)


omus:
> I had to increase the length of the FixedTimeZone name, as there was at least one name in backzone of length 18 characters.

Interesting. Definitely warrants a test.

src/tzdata/build.jl
```diff
@@ -1,7 +1,7 @@
 # The default tz source files we care about. See "ftp://ftp.iana.org/tz/data/Makefile"
 # "PRIMARY_YDATA" for listing of tz source files to include.
 const STANDARD_REGIONS = [
-    "africa", "antarctica", "asia", "australasia",
+    "africa", "antarctica", "asia", "australasia", "backzone",
```
omus, line 4: More work needs to be done to determine how we should classify "backzone". From what I have read so far it seems like it should probably be treated like a `LEGACY_REGION`

PR #423, opened 2022-12-14
Minimize allocations when unpacking `TimeZone`s from cache

### LINKED ISSUE #422: Creating `TimeZone` from string allocates unnecessarily
```julia
julia> using TimeZones

julia> TimeZone("UTC");  # populate cache

julia> @allocations TimeZone("UTC")  
2
```
`UTC` is a `FixedTimeZone`, which is already in the cache, and a `FixedTimeZone` is `isbits`... but we still allocate twice (when we should not need to allocate at all)

nickrobinson251:
- close #422 

Even though we're careful about caching the creation of TimeZones and made `FixedTimeZone` an `isbitstype`/`allocatedinline`, we _still allocate_  everytime `TimeZone` is called, even if just returning a `FixedTimeZone` from the cache!

I think this happens because the values in the cache are of type `TimeZone`, i.e. it's not known if we're returning a `FixedTimeZone` or a `VariableTimeZone`.

This PR proposes to instead maintain separate caches (per thread) for `FixedTimeZone`s and `VariableTimeZone`s, rather than a single one (per thread) for all `TimeZone`s

Effectively this is some code duplication for a performance improvement. 

`VariableTimeZone` still allocates once, presumably since it's not `isbitstype` (#271).

# Micro benchmarks

#### This PR
```julia
julia> str = "UTC";

julia> TimeZone(str);  # compile and cache

julia> @benchmark TimeZone($str)
BenchmarkTools.Trial: 10000 samples with 987 evaluations.
 Range (min … max):  46.733 ns … 156.619 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     53.994 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   54.265 ns ±   2.880 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                   ▂     █▂     █▃
  ▂▁▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂█▅▄▄▄▅██▅▆▇████▅▄▃▃▃▃▄▄▃▃▄▄▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂ ▃
  46.7 ns         Histogram: frequency by time         61.8 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

#### `master` (v1.9.1)
```julia
julia> @benchmark TimeZone($str)
BenchmarkTools.Trial: 10000 samples with 943 evaluations.
 Range (min … max):  100.168 ns …  1.490 μs  ┊ GC (min … max): 0.00% … 92.27%
 Time  (median):     104.277 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   107.123 ns ± 48.091 ns  ┊ GC (mean ± σ):  1.61% ±  3.33%

             ▁█▆▂
  ▂▃▃▄▃▄▄▅▅▅▇████▅▄▃▃▃▄▅▅▆▇▇▇▅▅▄▃▃▃▂▂▂▂▂▂▁▁▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  100 ns          Histogram: frequency by time          117 ns <

 Memory estimate: 64 bytes, allocs estimate: 2.
```

#### This PR

```julia
julia> str = "America/Winnipeg";

julia> TimeZone(str);   # compile and cache

julia> @benchmark TimeZone($str)
BenchmarkTools.Trial: 10000 samples with 946 evaluations.
 Range (min … max):   98.441 ns …  1.144 μs  ┊ GC (min … max): 0.00% … 90.99%
 Time  (median):     101.039 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   102.185 ns ± 32.794 ns  ┊ GC (mean ± σ):  1.01% ±  2.87%

                ▅▄█▆▁
  ▁▂▃▄▄▆█▇▆▄▄▄▄▇█████▄▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  98.4 ns         Histogram: frequency by time          108 ns <

 Memory estimate: 48 bytes, allocs estimate: 1.
```

#### `master` (v1.9.1)
```julia
julia> @benchmark TimeZone($str)
BenchmarkTools.Trial: 10000 samples with 909 evaluations.
 Range (min … max):  119.179 ns …  1.722 μs  ┊ GC (min … max): 0.00% … 92.34%
 Time  (median):     120.187 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   123.116 ns ± 47.667 ns  ┊ GC (mean ± σ):  1.33% ±  3.17%

  ▁▅██▇▆▄▃▂▂▁▁▁▁▂▃▃▂▃▃▂▁▁▁▁▁▁▁                                 ▂
  ████████████████████████████▇▇███▇▇██▇▇▇▇▇▇▇▇▇█▆▇▅▇▆▆▆▆▆▅▆▅▄ █
  119 ns        Histogram: log(frequency) by time       134 ns <

 Memory estimate: 64 bytes, allocs estimate: 2.
```


nickrobinson251:
```
~/r/TimeZones.jl> TZDATA_VERSION=2016j julia --project=benchmark/ -e 'using PkgBenchmark, TimeZones; export_markdown(stdout, judge(TimeZones, "origin/HEAD", verbose=false))'
PkgBenchmark: Running benchmarks...
[ Info: Installing 2016j tzdata region data
[ Info: Converting tz source files into TimeZone data
PkgBenchmark: creating benchmark tuning file /Users/nickr/repos/TimeZones.jl/benchmark/tune.json...
Tuning 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:36
Benchmarking 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:43
PkgBenchmark: Running benchmarks...
[ Info: Installing 2016j tzdata region data
[ Info: Converting tz source files into TimeZone data
PkgBenchmark: using benchmark tuning data in /Users/nickr/repos/TimeZones.jl/benchmark/tune.json
Benchmarking 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:48
```
# Benchmark Report for *~/repos/TimeZones.jl*

## Job Properties
* Time of benchmarks:
    - Target: 14 Dec 2022 - 18:05
    - Baseline: 14 Dec 2022 - 18:06
* Package commits:
    - Target: 85eb32
    - Baseline: f8f39b
* Julia commits:
    - Target: 36034a
    - Baseline: 36034a
* Julia command flags:
    - Target: None
    - Baseline: None
* Environment variables:
    - Target: None
    - Baseline: None

## Results
A ratio greater than `1.0` denotes a possible regression (marked with :x:), while a ratio less
than `1.0` denotes a possible improvement (marked with :white_check_mark:). Only significant results - results
that indicate possible regressions or improvements - are shown below (thus, an empty table means that all
benchmark results remained invariant between builds).

| ID                                                           | time ratio                   | memory ratio                 |
|--------------------------------------------------------------|------------------------------|------------------------------|
| `["ZonedDateTime", "local", "standard"]`                     | 0.84 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["ZonedDateTime", "range", "VariableTimeZone/DatePeriod"]`  | 0.91 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["arithmetic", "DatePeriod"]`                               | 0.93 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["arithmetic", "TimePeriod"]`                               |                1.09 (5%) :x: |                   1.00 (1%)  |
| `["arithmetic", "broadcast", "FixedTimeZone/TimePeriod"]`    | 0.94 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["arithmetic", "broadcast", "VariableTimeZone/DatePeriod"]` | 0.93 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["interpret", "local", "ambiguous"]`                        | 0.87 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["interpret", "local", "standard"]`                         | 0.89 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["parse", "ISOZonedDateTimeFormat"]`                        | 0.82 (5%) :white_check_mark: | 0.87 (1%) :white_check_mark: |
| `["parse", "issue#25"]`                                      | 0.92 (5%) :white_check_mark: | 0.88 (1%) :white_check_mark: |
| `["transition_range", "local", "ambiguous"]`                 | 0.83 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["transition_range", "local", "non-existent"]`              | 0.79 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["transition_range", "local", "standard"]`                  | 0.81 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["transition_range", "utc"]`                                |                1.43 (5%) :x: |                   1.00 (1%)  |
| `["tryparsenext_fixedtz", "+06"]`                            | 0.94 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["tryparsenext_fixedtz", "+0600"]`                          | 0.94 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["tryparsenext_fixedtz", "+06:00"]`                         | 0.95 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["tryparsenext_fixedtz", "-06"]`                            | 0.93 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["tryparsenext_fixedtz", "-0600"]`                          | 0.95 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["tryparsenext_fixedtz", "UTC"]`                            | 0.73 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["tryparsenext_fixedtz", "Z"]`                              | 0.56 (5%) :white_check_mark: |                   1.00 (1%)  |
| `["tz_data", "parse_components"]`                            | 0.89 (5%) :white_check_mark: | 0.84 (1%) :white_check_mark: |

## Benchmark Group List
Here's a list of all the benchmark groups executed by this job:

- `["ZonedDateTime"]`
- `["ZonedDateTime", "local"]`
- `["ZonedDateTime", "range"]`
- `["arithmetic"]`
- `["arithmetic", "broadcast"]`
- `["interpret", "local"]`
- `["interpret"]`
- `["parse"]`
- `["transition_range", "local"]`
- `["transition_range"]`
- `["tryparsenext_fixedtz"]`
- `["tryparsenext_tz"]`
- `["tz_data"]`

## Julia versioninfo

### Target
```
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  uname: Darwin 21.4.0 Darwin Kernel Version 21.4.0: Mon Feb 21 20:35:58 PST 2022; root:xnu-8020.101.4~2/RELEASE_ARM64_T6000 x86_64 i386
  CPU: Apple M1 Max:
                 speed         user         nice          sys         idle          irq
       #1-10  2400 MHz    1580681 s          0 s    1469385 s   30832641 s          0 s
  Memory: 64.0 GB (2506.71875 MB free)
  Uptime: 446789.0 sec
  Load Avg:  2.73779296875  2.6376953125  2.58837890625
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, westmere)
  Threads: 1 on 10 virtual cores
```

### Baseline
```
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  uname: Darwin 21.4.0 Darwin Kernel Version 21.4.0: Mon Feb 21 20:35:58 PST 2022; root:xnu-8020.101.4~2/RELEASE_ARM64_T6000 x86_64 i386
  CPU: Apple M1 Max:
                 speed         user         nice          sys         idle          irq
       #1-10  2400 MHz    1581658 s          0 s    1469753 s   30837878 s          0 s
  Memory: 64.0 GB (2527.17578125 MB free)
  Uptime: 446854.0 sec
  Load Avg:  2.205078125  2.5078125  2.54345703125
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, westmere)
  Threads: 1 on 10 virtual cores

**Comment by nickrobinson251** [comment]
Re-ran the benchmarks and the ones that are ❌ above are :white_check_mark: on re-run (and others are now not significant), so i think they're very noisy? Especially as most of these don't look like they even hit the codepaths changed in this PR?

Anyway, from this doesn't seem like we can say the two caches cause any performance issue

File: src/types/timezone.jl [hunk 1]
```diff
@@ -67,32 +83,56 @@ TimeZone(::AbstractString, ::Class)
...
+        _check_class(mask, class)
+        return vtz::VariableTimeZone
+    end
+
+    # We need to compute the timezone
+    tz_path = joinpath(_COMPILED_DIR[], split(str, "/")...)
+    if isfile(tz_path)
+        tz, class = open(TZJFile.read, tz_path, "r")(str)::Tuple{TimeZone,Class}
+        if tz isa FixedTimeZone
+            _ftz_cache()[str] = (tz, class)
```
nickrobinson251, line 103: `_ftz_cache()` is now potentially called multiple times in this function, whereas previously the `_tz_cache()` was called once (since we used `get!` to insert a value if not found) ...does this affect thread-safety?

src/types/timezone.jl
```diff
@@ -67,32 +83,56 @@ TimeZone(::AbstractString, ::Class)
...
+        # compliation which result in no tzdata time zones being available.
+        throw(ArgumentError(
+            "Unable to find time zone \"$str\". Try running `TimeZones.build()`."
+        ))
+    else
+        throw(ArgumentError("Unknown time zone \"$str\""))
     end
+end
 
+function _check_class(mask::Class, class::Class)
```
nickrobinson251, line 128: need to pass `str` here too (this doesn't seem covered by tests)

src/types/timezone.jl
```diff
@@ -1,27 +1,43 @@
-# Thread-local TimeZone cache, which caches time zones _per thread_, allowing thread-safe
+# Thread-local TimeZone caches, which caches time zones _per thread_, allowing thread-safe
 # caching. Note that this means the cache will grow in size, and may store redundant objects
 # accross multiple threads, but this extra space usage allows for fast, lock-free access
 # to the cache, while still being thread-safe.
-const THREAD_TZ_CACHES = Vector{Dict{String,Tuple{TimeZone,Class}}}()
+# Use a separate cache for FixedTimeZone (which is `isbits`) so the containers is concretely
+# typed and we avoid allocating a FixedTimeZone every time we get one from the cache.
```
NHDaly, line 6:
```suggestion
# Use a separate cache for FixedTimeZone (which is `isbits`) so the container is concretely
# typed and we avoid allocating a FixedTimeZone every time we get one from the cache.
```

src/types/timezone.jl
```diff
@@ -1,27 +1,43 @@
...
+const THREAD_FTZ_CACHES = Vector{Dict{String,Tuple{FixedTimeZone,Class}}}()
+const THREAD_VTZ_CACHES = Vector{Dict{String,Tuple{VariableTimeZone,Class}}}()
```
NHDaly, line 8: Since you're over here touching all of this code again, we _could_ consider switching to use https://github.com/JuliaConcurrent/MultiThreadedCaches.jl, which wraps up the pattern we used here into a standalone package. And that other package benefits from some eyes from `@tkf` and other engineers at RAI (@kpamnany and others), as well as some later attention to improving perf on accesses.

I don't want to complicate your PR too much, but I had been meaning to come back here and consider replacing the cache here with one from that package. I think that might be a nice cleanup since it would dramatically reduce the duplication here by using the factored out versions! :)

src/types/timezone.jl
```diff
@@ -67,32 +83,56 @@ TimeZone(::AbstractString, ::Class)
 function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
     # Note: If the class `mask` does not match the time zone we'll still load the
     # information into the cache to ensure the result is consistent.
-    tz, class = get!(_tz_cache(), str) do
-        tz_path = joinpath(_COMPILED_DIR[], split(str, "/")...)
+    ftz, class = get(_ftz_cache(), str, (nothing, Class(:NONE)))
+    if ftz !== nothing
+        _check_class(mask, class)
+        return ftz::FixedTimeZone
+    end
```
NHDaly, line 90: 🤔 hrrmm the downside of the https://github.com/JuliaConcurrent/MultiThreadedCaches.jl approach is that it currently _only_ exposes `get!()` as its interface, for reasons that i cannot fully remember..

So the trick you're doing here wouldn't work. Instead, you'd maybe need to figure out which type of TimeZone it is ahead of time? I don't know if that's possible though, just from its name? I guess you could have a third cache that just caches the type of the timezone?

So it'd be something like:
```julia
    type = get!(_tz_type_cache, str) do
        compute_tz_type(str)
    end
    if type === :FixedTimeZone
        tz, class = get!(ftz_cache, str) do
            compute_fixed_timezone(str)
        end
    elseif type === :VariableTimeZone
        tz, class = get!(vtz_cache, str) do
            compute_variable_timezone(str)
        end
    end
```
Something like that? I'm not sure if splitting things up this way is feasible though?

NHDaly (commented):
👍 👍 This is a nice change, @nickrobinson251! It makes sense why it reduces the allocations. 👍 Thanks for investigating this so deeply

NHDaly:
I'm not sure i understand why the VariableTimeZones still alloc when you look them up. I agree with you it is unexpected.. I'm not sure the isbits thing should matter, since we're pointing to the _existing_ vector, not allocating a new one, and Julia is supposed to be able to stack-allocate an immutable struct that contains mutable fields, since like 1.4 or something. So i'm not sure i understand yet; it might be good to look through a profile together, once the rest of the issues in the PR are addressed.

src/types/timezone.jl
nickrobinson251: > figure out which type of TimeZone it is ahead of time?

I think that's what's not easy, because we actually create the TimeZones by reading from a timezone-specific file
https://github.com/JuliaTime/TimeZones.jl/blob/f8f39b3100334a0079f525304f7dddc7656204da/src/tzjfile/read.jl#L52-L78

i.e. we have to read the file to know what type it will be, and we don't want to have to read multiple times

so i think we'd have change that `read` function to return the type, then put that value in a third cache, while holding on to the `tz, class`

something like
```julia
    local tz, class
    type = get!(_tz_type_cache, str) do
        tz, class, type = compute_tz(str)
        return type
    end
    if type === :FixedTimeZone
        tz, class = get!(ftz_cache, str) do
            (tz, class)
        end
    elseif type === :VariableTimeZone
        tz, class = get!(vtz_cache, str) do
            (tz, class)
       end
    end
```

nickrobinson251: hmmmm, this way's not looking nice to me 🤔 

how much benefit do we think MultiThreadedCaches.jl is going to add for us?

nickrobinson251: we could maybe go add `get` and `empty!` to the MultiThreadedCaches.jl API to reduce the necessary changes

omus:
I'll try to take a look soon. I'll call out that I also want to get #382 merged soon as well.

test/helpers.jl
```diff
@@ -1,5 +1,17 @@
 # Utility functions for testing
 
+if VERSION < v"1.9.0-"  # https://github.com/JuliaLang/julia/pull/47367
+    macro allocations(ex)
+        quote
+            Base.Experimental.@force_compile
```
Drvi, line 6: CI failures suggest that Julia 1.6 does not know `Base.Experimental.@force_compile`

nickrobinson251: ah, thanks! in which case we can switch to `while false; end` (https://github.com/JuliaLang/julia/pull/42128/)
```suggestion
            while false; end  # want to force compilation, but v1.6 doesn't have `@force_compile`
```

omus:
I've been a bit swamped this week. Overall, I'm in favour of this change and will definitely merge this before #382

nickrobinson251:
Just following up to make sure this doesn't drop off your radar. Let me know if i can do anything to help, @omus :)

omus:
It turned out #382 became urgent to merge in Julia 1.9 so the original plan had changed. If @nickrobinson251 you have time to rebase this against the latest version of the code that would be great. If not, I'll try to adapt this code myself sometime.

nickrobinson251:
I think this is just a (not-very-simple) rebase away. And I would love, love, love someone to take that on, as I don't have capacity at the moment -- please feel encouraged to do so!

PR #426, opened 2023-01-12
fix 424

lgeissbauer-btig:


lgeissbauer-btig:
We don't need this because of
https://github.com/aviatesk/JET.jl/issues/456

omus:
Do you still want to merge this or should this be closed? I do think the original code is ever so slightly clearer

PR #431, opened 2023-05-02
Update dependencies for `generate_artifacts.jl`

omus:
Some of the dependencies were out of date which was causing issues when trying to run this.


omus:
Tests will be fixed by #432

PR #412, opened 2022-11-06
Modify show for FixedTimeZone

omus:
In https://github.com/JuliaStrings/InlineStrings.jl/pull/52 (InlineStrings@1.3.0) the `repr` method for `InlineString` now shows type information (👍 ). For the `show` method for `FixedTimeZone` this information isn't required as all strings are convered to an `InlineString`. To keep the output of `show` concise we'll just show a simple string instead of including the type information.

PR #432, opened 2023-05-02
Show `FixedTimeZone` name as a regular string

omus:
The change from https://github.com/JuliaStrings/InlineStrings.jl/pull/52 changed how `show` works for `InlineStrings.String15` which broke the tests here. I've updated how we show `FixedTimeZone` so that the name always prints as a regular `String`. This works fine for constructing a `FixedTimeZone` and is more terse this seems like a reasonable change.

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -60,7 +60,8 @@ function Base.show(io::IO, tz::FixedTimeZone)
         std = Dates.value(tz.offset.std)
         dst = Dates.value(tz.offset.dst)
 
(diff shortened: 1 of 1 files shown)


omus:
I did already attempt to fix this in: https://github.com/JuliaTime/TimeZones.jl/pull/412. There must have been another change to `string` which broke the original fix

omus:
The nightly documentation failure is something else. Will tackle that separately

PR #435, opened 2023-05-08
Fix thread-local cache management for Julia 1.9.0

### LINKED ISSUE #429: The per-thread `empty!()` call in `compile()` is broken on julia master.
Julia master recently merged https://github.com/JuliaLang/julia/pull/49094, which moved the interactive threadpool to come _before_ the default threadpool.

This means that the threadids you get from iterating `@threads :static` no longer match up with the indexes.
For example, here is julia started with `JULIA_NUM_THREADS=4,1 julia --proj`:
```julia
julia> VERSION
v"1.10.0-DEV.1118"

julia> Threads.nthreads()
4

julia> Threads.maxthreadid()
5

julia> Threads.@threads :static for i in 1:Threads.nthreads()
           @info i, Threads.threadid()
           @assert Threads.threadid() === i "TimeZones.TZData.compile() must be called from the main, top-level Task."
       end
[ Info: (1, 2)
[ Info: (3, 4)
[ Info: (4, 5)
[ Info: (2, 3)
ERROR: TaskFailedException

    nested task error: AssertionError: TimeZones.TZData.compile() must be called from the main, top-level Task.
    Stacktrace:
     [1] macro expansion
       @ ./REPL[5]:3 [inlined]
     [2] (::var"#18#threadsfor_fun#4"{var"#18#threadsfor_fun#3#5"{UnitRange{Int64}}})(tid::Int64; onethread::Bool)
       @ Main ./threadingconstructs.jl:163
     [3] #18#threadsfor_fun
       @ ./threadingconstructs.jl:130 [inlined]
     [4] (::Base.Threads.var"#1#2"{var"#18#threadsfor_fun#4"{var"#18#threadsfor_fun#3#5"{UnitRange{Int64}}}, Int64})()
       @ Base.Threads ./threadingconstructs.jl:108

...and 3 more exceptions.
```
Using `for i in 1:Threads.maxthreadid()` doesn't work either, because it only iterates in the current thread pool:
```julia
julia> Threads.@threads :static for i in 1:Threads.maxthreadid()
           @info i, Threads.threadid()
       end
[ Info: (4, 4)
[ Info: (5, 5)
[ Info: (1, 2)
[ Info: (3, 3)
[ Info: (2, 2)
```
(note there are two iterations with `threadid() = 2`.)

.........

It doesn't seem like there's any way to run a `Threads.@threads` over the interactive threads, from what I can see. Not 100% sure about that.

----

It might be that we need to either change the way this works so that instead of eagerly emptying all the caches, we do something more cooperative, like setting some "time to reset" bit for each thread, which is lazily checked on the first access on each thread and the cache is cleared before doing anything else? Or maybe triggering a channel that each thread has a task waiting on to clear its cache.

Or maybe we need to make a change in Julia to provide another possible schedule to `@threads`, like `:interactive` to force it to schedule over the interactive threadpool? :/

CC: @omus, @jeffbezanson

### LINKED ISSUE #430: `TimeZones.TZData.compile()` fails on Julia 1.9.0-rc3
Due to https://github.com/JuliaLang/julia/pull/49094, the interactive thread(s) (if you start any) are now thread 1 (and up) and the default threads start after that. That breaks [this assertion](https://github.com/JuliaTime/TimeZones.jl/blob/master/src/tzdata/compile.jl#L701), which I don't think is necessary anyway?

I think this can be fixed with something like:
```julia
function empty_thread_local_caches(both_pools = true)
    tp = Threads.threadpool()
    Threads.@threads :static for i in 1:Threads.nthreads(tp)
        empty!(_tz_cache())
    end
    !both_pools && return
    if tp == :default
        if Threads.nthreads(:interactive) > 0
            fetch(@spawn :interactive empty_thread_local_caches(false))
        end
    else
        fetch(@spawn :default empty_thread_local_caches(false))
    end
end
```

Also, since this package uses thread-local data, you probably want to use `maxthreadid()` [here](https://github.com/JuliaTime/TimeZones.jl/blob/master/src/types/timezone.jl#L24) instead of `nthreads()`.

kpamnany:
Here's an alternative (to https://github.com/JuliaTime/TimeZones.jl/pull/382) way to fix the thread-local caches. It's a bit clunky. And also I think Julia would generally prefer that nobody did thread-local stuff. And it remains racy against concurrent use of the caches (though that's probably not a major concern). But here it is anyway.

Closes #430 
Closes #429

### DIFF
--- a/src/types/timezone.jl
+++ b/src/types/timezone.jl
@@ -19,9 +19,16 @@ const THREAD_TZ_CACHES = Vector{Dict{String,Tuple{TimeZone,Class}}}()
 end
 @noinline _tz_cache_length_assert() = @assert false "0 < tid <= length(THREAD_TZ_CACHES)"
 
+@static if VERSION >= v"1.9.0"
+function _reset_tz_cache()
+    # ensures that we didn't save a bad object
+    resize!(empty!(THREAD_TZ_CACHES), Threads.maxthreadid())
+end
+else # prior to 1.9.0
 function _reset_tz_cache()
     # ensures that we didn't save a bad object
     resize!(empty!(THREAD_TZ_CACHES), Threads.nthreads())
+end
 end
 
 """
--- a/src/tzdata/compile.jl
+++ b/src/tzdata/compile.jl
@@ -689,18 +689,38 @@ function compile(tz_source::TZSource; kwargs...)
     return results
 end
 
+@static if VERSION >= v"1.9.0"
+# As the caches are thread local, we must spawn a task on every thread to
+# empty that thread's cache.
+function empty_thread_local_caches(both_pools = true)
+    tp = Threads.threadpool()
+    Threads.@threads :static for i in 1:Threads.nthreads(tp)
+        empty!(_tz_cache())
+    end
+    !both_pools && return
+    if tp == :default
+        if Threads.nthreads(:interactive) > 0
+            fetch(Threads.@spawn :interactive empty_thread_local_caches(false))
+        end
+    else
+        fetch(Threads.@spawn :default empty_thread_local_caches(false))
+    end
+end
+else # prior to 1.9.0
+function empty_thread_local_caches(both_pools = true)
+    Threads.@threads :static for i in 1:Threads.nthreads()
+        empty!(_tz_cache())
+    end
+end
+end
+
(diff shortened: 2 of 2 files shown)


omus:
Went with #382 instead

PR #436, opened 2023-06-16
Upgrade Documenter to 0.27

### LINKED ISSUE #397: Documentation is out-of-date
https://juliatime.github.io/TimeZones.jl/stable/ currently points to v1.5.2, which is nearly 2-year old.  Also, documentation uses a very old version of `Documenter.jl`: https://github.com/JuliaTime/TimeZones.jl/blob/2d1af8a1b21247887db1d85971520dd150169dfb/docs/Project.toml#L6

mortenpi:
The documentation has [not deployed for nearly 3 years](https://github.com/JuliaTime/TimeZones.jl/tree/gh-pages). It looks like it tries to use some `TRAVIS_*` variables, because the ancient Documenter version that it tries to use probably predated full GitHub Actions support. I hope that bumping Documenter is enough (fix #397).

I checked the built docs manually, and I didn't notice any major issues with the upgrade.

The only thing is that it looks like the doctests are pretty broken in general. I don't quite understand why the current CI is passing, but I assume that it will start failing now. But there have also been changes in Documenter, so it's probably a multitude of factors here.

A few changes in general:

* Decouples the documentation CI from doctests. I'd say you don't want the docs for a tag to fail to deploy just because you happen to have one doctest that is outdated.
* Just doctest on 1.9, because printing is not stable across Julia versions anyway, so doctesting with different Julia versions is not really recommended. You should target a specific version. Also, since we're committing the manifest here, we don't run into any strange problems with that -- the manifest is from 1.9 now.
* Removes the `Pkg.develop(PackageSpec(path=pwd()))` lines (from CI), since those are not necessary when you have the Manifest anyway.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -89,46 +89,32 @@ jobs:
   doctest:
     name: Documentation - Julia ${{ matrix.version }}
     runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: true
-      matrix:
-        version:
-          - "1.6"
-          - nightly
-        os:
-          - ubuntu-latest
-        arch:
-          - x64
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
-          version: ${{ matrix.version }}
-          arch: ${{ matrix.arch }}
+          version: "1.9"
       - uses: julia-actions/julia-buildpkg@latest
       - run: |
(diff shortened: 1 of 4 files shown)


mortenpi:
I would appreciate if someone could approve the CI run so that I could see if the changes are working or not.

omus:
Thanks @mortenpi! I appreciate the PR

PR #438, opened 2023-07-26
make latest unicode-cldr a non-lazy artifact

KristofferC:


omus:
Thanks. I'll be incorporating this change during the JuliaCon week. At the moment the Artifact.toml is generated by the `dev/generate_artifacts.jl` script so that code will also need to be updated (I can take care of that). Additionally, as you pointed out in person we don't need to keep each version of unicode-cldr as artifacts. I believe this is just a mistake caused from the artifact script which doesn't clean up old versions.

For those interested. Lazy artifacts can be problematic when using PackageCompiler.jl as all lazy artifacts can be downloaded to ensure no artifacts have to be retrieved at runtime. The unicode-cldr artifact is decently large which is why this is a problem.

omus:
Superseded by #439

PR #439, opened 2023-07-27
Use non-lazy platform specific artifacts for unicode-cldr

omus:
As brought up in #438 we were previously using lazy artifacts for the multiple unicode-cldr artifacts which are only needed on Windows for translating the current OS time zone name into the IANA time zone name. This approach was problematic because PackageCompiler 1.0 would always download all of the lazy artifacts and PackageCompiler 2.0 would not download the lazy artifacts (unless you use `include_lazy_artifacts=true`) resulting in the used unicode-cldr artifact being downloaded at runtime. Additionally, we only require the latest unicode-cldr artifact and the only reason we included multiple of these artifacts was due to us failing to clean them up in the `dev/generate_artifacts.jl` script which generates this file.

This PR changes the `dev/generate_artifacts.jl` to update the `Artifacts.toml` to only include the latest unicode-cldr. We now specify the platform for this architecture so that the artifact is only used on Windows which also allows us to specify it as non-lazy. Finally, there was a newer unicode-cldr release so I've also updated to use release-43-1 here.

Supercedes #438

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,8 +36,6 @@ jobs:
           # Test 32-bit only on Linux
           - os: macOS-latest
             arch: x86
-          - os: windows-latest
-            arch: x86
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,6 +36,8 @@ jobs:
           # Test 32-bit only on Linux
           - os: macOS-latest
             arch: x86
+          - os: windows-latest
+            arch: x86
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -2101,57 +2101,139 @@ lazy = true
     sha256 = "02f8321a3236105338a2a071f9f9ef98dadeb8f939cbe3dc2dd53e1931015604"
     url = "https://data.iana.org/time-zones/releases/tzdata96k.tar.gz"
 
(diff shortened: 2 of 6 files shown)


omus:
There is an argument to be made for testing on Windows 32-bit to ensure this artifact is present and working. I'll temporarily enable this environment as part of this PR but I don't believe it's necessary to test against this environment for other PRs. The exception being additional changes to the unicode-cldr artifact generation code.

DilumAluthge:
> The exception being additional changes to the unicode-cldr artifact generation code.

We could configure it so that the Windows 32-bit CI job runs on PRs that touch said code, but otherwise is automatically skipped.

omus:
An additional future improvement that can be made would be to create a new artifact to replace our use of unicode-cldr such that we can use a tiny XML file as the artifact instead of the 50MB artifact we currently fetch.

omus:
> > The exception being additional changes to the unicode-cldr artifact generation code.
> 
> We could configure it so that the Windows 32-bit CI job runs on PRs that touch said code, but otherwise is automatically skipped.

The path filtering with GHA workflows happens at a workflow level so we'd either need a separate workflow, [generate our matrix](https://github.blog/changelog/2020-04-15-github-actions-new-workflow-features/#new-fromjson-method-in-expressions), or introduce conditions into our steps.

If I end up making the reduced sized artifact I'll probably use a separate data pkg and remove the Artifacts.toml here so I'll put on the CI changes for now.

omus:
Validated that Windows 32-bit works in https://github.com/JuliaTime/TimeZones.jl/pull/439/commits/dee8df5c23fcdd903ab875717b66514fffed425c. Reverted the commit and will merge after CI completes

Issue #440, opened 2023-07-31
Failed to create ZonedDateTime range in Julia 1.9.2


JakubPorubcansky:
In Julia 1.7.3, the following code was working

```julia
julia> using Dates, TimeZones

julia> range(ZonedDateTime(2007, 03, 24, tz"Europe/Budapest"), length=6, step=Hour(1))
ZonedDateTime(2007, 3, 24, tz"Europe/Budapest"):Hour(1):ZonedDateTime(2007, 3, 24, 5, tz"Europe/Budapest")

julia> versioninfo()
Julia Version 1.7.3
Commit 742b9abb4d (2022-05-06 12:58 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)

(@v1.7) pkg> status TimeZones
      Status `C:\Users\jakub\.julia\environments\v1.7\Project.toml`
  [f269a46b] TimeZones v1.11.0
```

but after upgrading to Julia 1.9.2, it does not work anymore

```julia
julia> using Dates, TimeZones

julia> range(ZonedDateTime(2007, 03, 24, tz"Europe/Budapest"), length=6, step=Hour(1))
Error showing value of type StepRangeLen{ZonedDateTime, ZonedDateTime, Hour, Int64}:
ERROR: MethodError: no method matching ZonedDateTime(::ZonedDateTime)

Closest candidates are:
  ZonedDateTime(::Union{Period, TimeZone}...)
   @ TimeZones C:\Users\jakub\.julia\packages\TimeZones\miNCa\src\types\zoneddatetime.jl:136
  ZonedDateTime(::AbstractString)
   @ TimeZones C:\Users\jakub\.julia\packages\TimeZones\miNCa\src\parse.jl:86
  ZonedDateTime(::AbstractString, ::AbstractString; locale)
   @ TimeZones C:\Users\jakub\.julia\packages\TimeZones\miNCa\src\parse.jl:100
  ...

Stacktrace:
  [1] unsafe_getindex(r::StepRangeLen{ZonedDateTime, ZonedDateTime, Hour, Int64}, i::Int64)
    @ Base .\range.jl:948
[...]

Issue #399, opened 2022-08-03
Error after compilation - Unable to find time zone "Europe/Budapest". Try running `TimeZones.build()`


JakubPorubcansky:
I am trying to compile this minimal example package to an app

<details>
<summary>TestPackage.jl</summary>

```julia
module TestPackage

using TimeZones

function julia_main()::Cint
  try
    @info "Start"
    sleep(3)
    timeZone = TimeZone("Europe/Budapest")
    @info "timeZone=$timeZone"
    sleep(3)
    @info "Finish"
  catch exc
    @error "Something went wrong" exception=(exc, catch_backtrace())
    return 1
  end
  return 0
end

end # module

```
</details>

<details>
<summary>Project.toml</summary>

```
name = "TestPackage"
uuid = "d2663e91-0ef1-41e5-8c08-2797f1a8a949"
authors = ["Jakub Porubcansky <jakub.porubcansky@gmail.com"]
version = "0.1.0"

[deps]
TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
```
</details>

<details>
<summary>Manifest.toml</summary>

```
# This file is machine-generated - editing it directly is not advised

julia_version = "1.7.3"
manifest_format = "2.0"

[[deps.ArgTools]]
uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"

[[deps.Artifacts]]
uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"

[[deps.Base64]]
uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"

[[deps.Compat]]
deps = ["Base64", "Dates", "DelimitedFiles", "Distributed", "InteractiveUtils", "LibGit2", "Libdl", "LinearAlgebra", "Markdown", "Mmap", "Pkg", "Printf", "REPL", "Random", "SHA", "Serialization", "SharedArrays", "Sockets", "SparseArrays", "Statistics", "Test", "UUIDs", "Unicode"]
git-tree-sha1 = "9be8be1d8a6f44b96482c8af52238ea7987da3e3"
uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
version = "3.45.0"

[[deps.CompilerSupportLibraries_jll]]
deps = ["Artifacts", "Libdl"]
uuid = "e66e0078-7015-5450-92f7-15fbd957f2ae"

[[deps.Dates]]
deps = ["Printf"]
[...]

Issue #373, opened 2022-02-23
Reduce Unicode CLDR artifact size


jaakkor2:
On my local drive size of
.julia\artifacts\8d7201f06ff8061c1b5bff0f661b5297ec1d3158\cldr-release-40
is 318 MB.

If I searched code correctly, only one 49 kB file is actually used
https://github.com/unicode-org/cldr/blob/release-40/common/supplemental/windowsZones.xml
here
https://github.com/JuliaTime/TimeZones.jl/blob/master/src/winzone/WindowsTimeZoneIDs.jl

Could the artifact point directly to the `windowsZones.xml` file?

jaakkor2:
`..\.julia\artifacts\768b8b82edd878449348f31bf989867bb13e42a9\cldr-release-43`
folder size is  335 MB, and this package is using one file of it (IIUC) 
`..\.julia\artifacts\768b8b82edd878449348f31bf989867bb13e42a9\cldr-release-43\common\supplemental\windowsZones.xml`
has the size of 49 kB.

omus:
You are correct that TimeZones.jl only requires the `windowsZones.xml` file from the CLDR artifacts. This is an unfortunate side effect of this package switching to use Julia artifacts which requires an archive to be referenced and not just a flat file.

In order to address this problem with the artifacts system we'd need to generate our own permanent artifact archive for each CLDR release. Probably the right approach would be to have these artifacts as GitHub release assets which follows what JLLs do. It also probably makes sense to have this not be a part of the TimeZones.jl repo but probably a CLDRWindowsZones_jll repo so that the tags/releases make more sense. Additionally, it may make more sense to make a CLDR_jll.jl and have just the `windowsZones.xml` included in a special release asset so we can retrieve it separately.

The end result is having a much smaller artifact footprint as well as being able to update CLDR versions without having to make a new TimeZones.jl release. Unfortunately, this is a bunch of work but I do want to tackle this before or around JuliaCon 2023.

omus:
The PR https://github.com/JuliaTime/TimeZones.jl/pull/439 changed the Unicode CLDR artifacts to no longer be non-lazy and platform specific. Although this doesn't impact the download size on Windows machines it does limit the downloads to only be on Windows.

It still makes sense to make separate data package with this information akin to what was done in #441. I do want to have #441 be out in the wild for a little while to ensure this artifact model works well.

visr:
[...]

PR #441, opened 2023-08-07
Use precompiled time zone information from TZJData.jl

omus:
Updates the TimeZones to use the TZJData package as the default source of compiled time zones information. This eliminates the need to build the time zone information at package initialization time and avoids download issues with communicating with the IANA servers eliminating quite a few of the problems we've been having.

Users may choose to upgrade or downgrade the TZJData package they use to modify the version of the tzdata they use with the TimeZones package. Alternately, users with special use cases can still compile tzdata rules into time zone information which is stored in a scratch directory. Most users won't require this but it's useful when updating to using the latest tzdata version or exploratory research.

Fixes: #359, #357, #399, #343, #331, #325, #440, #300, #321, #302, #279

Future work: The Unicode CLDR artifacts are around ~50MB and we only require a small XML file located within them. We should probably take a similar approach to the TZJData.jl package and only host the data we require. (https://github.com/JuliaTime/TimeZones.jl/issues/373)


omus:
There's a bit more work to do here yet. I've got to:
- [ ] Fix the benchmark tests
- [ ] Deprecate the old build methods
- [ ] Update TZJData to have a 1.0 release
- [ ] Additional refactoring

I'm aiming to get this done by the end of the week

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
 version = "1.11.0"
 
 [deps]
+Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
```
omus, line 7: Note we can remove the dependency on LazyArtifacts now but doing so interferes with running PkgBenchmarks. I'll leave in this dependency for now and remove it in a follow up PR where benchmarks don't matter

test/tzdata/build.jl
```diff
@@ -0,0 +1,55 @@
...
+            @test_throws TimeZones.UnhandledTimeError ZonedDateTime(2100, warsaw)
+
+            # Using the tz string macro which runs at parse time means that the resulting TimeZone
+            # will not reflect changes made from compile or new builds during runtime.
+            @test tz"Africa/Windhoek" != TimeZone("Africa/Windhoek")
+            @test tz"Africa/Windhoek".cutoff != DateTime(2201, 4, 5)
+            @test TimeZone("Africa/Windhoek").cutoff == DateTime(2201, 4, 5)
+        end
+    finally
+        TimeZones._reload_cache(TimeZones._COMPILED_DIR[])
```
omus, line 53: Would be better if these tests could just compile the extended time zones in memory and avoid writing these to disk. Will try to do as a follow up

omus:
PR is only waiting on a 1.0 release of TZJData. Anyone wanting to review this PR should do so in the next couple of days.

omus:
Took a bit longer to get the TZJData v1.0.0 release merged into the registry than I anticipated. I'll let CI run and make a new TimeZones.jl release tonight and we'll finally be able to close all those issues

Project.toml
omus: https://github.com/JuliaTime/TimeZones.jl/issues/443

test/tzdata/build.jl
omus: https://github.com/JuliaTime/TimeZones.jl/issues/444

PR #446, opened 2023-08-22
move RecipesBase support to an extension on 1.9+

palday:
This uses [the method in the Pkg docs](https://pkgdocs.julialang.org/v1/creating-packages/#Transition-from-normal-dependency-to-extension) to continue support for Julia < 1.9.

On 1.9, this shaves about 100ms off of `using TimeZones` for me.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.12.0"
+version = "1.13.0"
 
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
@@ -26,8 +26,15 @@ TZJData = "1"
 julia = "1.6"
 p7zip_jll = "17.4"
 
+[extensions]
+TimeZonesRecipesBaseExt = "RecipesBase"
+
 [extras]
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test"]
+test = ["Test", "RecipesBase"]
+
+[weakdeps]
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
--- a/ext/TimeZonesRecipesBaseExt.jl
+++ b/ext/TimeZonesRecipesBaseExt.jl
@@ -28,4 +28,4 @@ for details on the options and their tradeoffs.
     end
 end
 
-end #module
+end # module
--- a/ext/TimeZonesRecipesBaseExt.jl
+++ b/ext/TimeZonesRecipesBaseExt.jl
@@ -1,3 +1,8 @@
+module TimeZonesRecipesBaseExt
+
+using TimeZones
+using RecipesBase: RecipesBase, @recipe
+
 #==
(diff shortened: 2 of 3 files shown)


palday:
For the Benchmark CI -- it looks like the environment isn't re-resolved when comparing to `orign/HEAD` and this causes errors.

src/TimeZones.jl
```diff
@@ -92,7 +91,10 @@ include("ranges.jl")
 include("discovery.jl")
 include("rounding.jl")
 include("parse.jl")
-include("plotting.jl")
 include("deprecated.jl")
 
+if !isdefined(Base, :get_extension)
```
omus, line 96:
```suggestion
# Required to support Julia `VERSION < v"1.9"`
if !isdefined(Base, :get_extension)
```

ext/TimeZonesRecipesBaseExt.jl
```diff
@@ -22,3 +27,5 @@ for details on the options and their tradeoffs.
         [], ys
     end
 end
+
+end #module
```
omus, line 31:
```suggestion
end # module
```

ext/TimeZonesRecipesBaseExt.jl
```diff
@@ -1,3 +1,8 @@
+module TimeZonesRecipesBaseExt
```
omus, line 1: Although the Pkg.jl documentation use this convention in their docs I think having "TimeZones" named in the extension is a little too verbose for my liking.

```suggestion
module RecipesBaseExt
```
Will need to also rename the file.

Project.toml
```diff
@@ -26,8 +26,15 @@ TZJData = "1"
 julia = "1.6"
 p7zip_jll = "17.4"
 
+[extensions]
+TimeZonesRecipesBaseExt = "RecipesBase"
```
omus, line 30:
```suggestion
RecipesBaseExt = "RecipesBase"
```

Project.toml
```diff
@@ -26,8 +26,15 @@ TZJData = "1"
...
+
 [extras]
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
```
omus, line 33: This is now required as on Julia 1.9 the `[weakdeps]` take precedence over `[deps]` so in order to test with this we need to now add `[extras]` right?

That explains the PkgBenchmark failure better too. You could work around that failure by setting the Julia version for the benchmark CI tests to 1.8 so that the `[weakdeps]` are ignored and `[deps]` are preferred. I'm fine with ignoring the benchmark tests for this PR so don't feel the need to go through this change.

ext/TimeZonesRecipesBaseExt.jl
palday: It's not documented, but apparently the convention is to have the 'host' package's name first to avoid name collisions, which are problematic. 

(pointed out to be by @ararslan in an NLopt PR: https://github.com/JuliaOpt/NLopt.jl/pull/204#discussion_r1293669487)

Project.toml
palday: > This is now required as on Julia 1.9 the `[weakdeps]` take precedence over `[deps]` so in order to test with this we need to now add `[extras]` right?

Correct.

I would also prefer to just ignore the failure for this one PR so that benchmarking remains on the current release moving forward.

ext/TimeZonesRecipesBaseExt.jl
omus: Thanks. I would speculate this is probably due to how `.ji` files currently work. Okay, let's leave this then.

omus (approved):

PR #450, opened 2023-10-24
move from CLDR artifact to code generation

### LINKED ISSUE #373: Reduce Unicode CLDR artifact size

visr:
Fixes #373, based on the proposal outlined in https://github.com/JuliaTime/TimeZones.jl/issues/373#issuecomment-1777825007. The main benefit is to remove a large artifact download on Windows, which has long paths, leading to issues with PackageCompiler.

This removes the `WindowsTimeZoneIDs module`. Even though the generated dict is only used in Windows, I decided to remove the conditional `Sys.iswindows` includes since these translations could perhaps be useful on other platforms as well.


visr:
@omus I would by happy to hear if you think this is a good way forward.

dev/generate_windows_zones.jl
```diff
@@ -0,0 +1,68 @@
...
+
+"Extract the mapping from Windows tzid to Olsen zone ID"
+function create_mapping(doc::Node)::Vector{Pair{String,String}}
+    timezones = doc[end][end][end]
+    mapping = Pair{String,String}[]
+
+    for node in children(timezones)
+        if tag(node) == "mapZone"
+            attr = attributes(node)
+            if attr["territory"] == "001"
```
omus, line 28:
```suggestion
            # Territory "001" is the global default
            if attr["territory"] == "001"
```

visr:
Thanks. Indeed I can see a benefit of having a separately upgradable package for this. I went for this change within the package to keep the PR somewhat small. I figured that when switching to a separate package we'd want to generate the code like this as well, so it is a good step to take regardless.

If you prefer I'd be happy to create a separate package.

I looked at some past cldr releases and saw that the Windows time zone IDs only change every few cldr releases.

visr:
Would love to hear what I can do to land this.

visr:
Bump :)

test/windows_zones.jl
```diff
@@ -0,0 +1,3 @@
+using TimeZones: WINDOWS_TRANSLATION
+
+@test WINDOWS_TRANSLATION["Central European Standard Time"] == "Europe/Warsaw"
```
omus, line 3:
```suggestion
@test occursin(r"^release-\d+(-\d+)$", UNICODE_CLDR_VERSION)
@test WINDOWS_TRANSLATION["Central European Standard Time"] == "Europe/Warsaw"
```

test/windows_zones.jl
```diff
@@ -0,0 +1,3 @@
+using TimeZones: WINDOWS_TRANSLATION
```
omus, line 1:
```suggestion
using TimeZones: UNICODE_CLDR_VERSION, WINDOWS_TRANSLATION
```

omus (commented):
Thanks for this. I'm good with proceeding with this change once we complete a round of refactoring. It would be good to make the separate repo but I'm good to adopt this here first.

dev/generate_windows_zones.jl
```diff
@@ -0,0 +1,69 @@
...
+end
+
+"""
+Generate code mapping the Windows tzid to the Olsen zone ID.
+The mapping is defined by the Unicode Common Locale Data Repository (CLDR).
+
+Details on the mapping can be found at:
+https://cldr.unicode.org/development/development-process/design-proposals/extended-windows-olson-zid-mapping
+"""
+function main(ARGS)
```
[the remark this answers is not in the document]
visr: Would be happy to remove ARGS here, but the reason I went for this signature is to already align with the standardized entry point coming in 1.11: https://docs.julialang.org/en/v1.11-dev/NEWS/#Command-line-option-changes.

This way we don't need to modify the signature later.

test/windows_zones.jl
visr: Thanks, I needed to add the `?` after the last group since a second number is optional; `release-43-1` and `release-45` are both valid.

visr:
Thanks for the review! I completed the refactoring and regenerated using the latest CLRD release. Only left `main(ARGS)` in, let me know if you prefer to change that as well.

dev/generate_windows_zones.jl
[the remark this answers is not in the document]
omus: There may be a change to use `Main.main(args)`: https://github.com/JuliaLang/julia/issues/54252. We can leave this as is but it would be a good idea to leave a link reference as you're intending on using a feature which isn't actually used here.

dev/generate_windows_zones.jl
```diff
@@ -0,0 +1,78 @@
...
+    open(path; write=true) do io
+        println(io, "# This file is generated by `dev/generate_windows_zones.jl`, do not edit.\n")
+        println(io, "const UNICODE_CLDR_VERSION = ", repr(latest_cldr_version), '\n')
+        println(io, "const WINDOWS_TRANSLATION = Dict{String, String}(")
+        for zone_pair in mapping
+            println(io, "    ", repr(zone_pair), ',')
+        end
+        println(io, ')')
+    end
+    return nothing
```
omus, line 53: The function body needed to be updated to make use of the `io::IO` change:
```suggestion
    println(io, "# This file is generated by `dev/generate_windows_zones.jl`, do not edit.\n")
    println(io, "const UNICODE_CLDR_VERSION = ", repr(latest_cldr_version), '\n')
    println(io, "const WINDOWS_TRANSLATION = Dict{String, String}(")
    for zone_pair in mapping
        println(io, "    ", repr(zone_pair), ',')
    end
    println(io, ')')
    return nothing
```

test/windows_zones.jl
omus: My bad. I added the grouping to make it optional but just forgot to add the `?` modifier

omus (commented):
Just needs the `io::IO` change and it looks like the benchmark `Project.toml` may need some updating. We'll see what CI says about that on this run.

visr:
> looks like the benchmark Project.toml may need some updating

Yes, I see "Package TimeZones does not have Artifacts in its dependencies". This looks like it is due to `origin/HEAD` needing Artifacts whereas this removes it. Same happened in https://github.com/JuliaTime/TimeZones.jl/pull/446#issuecomment-1688605979.

omus (approved):

PR #455, opened 2024-04-11
remove LazyArtifacts as a dependency. There are no lazy artifacts in this repo anymore.

KristofferC:
(at least from what I can see...)


visr (approved):
Correct, not lazy and Windows-only since #439, which superceded your #438.
#450 additionally gets rid of the Artifacts dependency.

PR #456, opened 2024-04-11
Make `ISOZonedDateTimeFormat` a const

KristofferC:
A bit of latency save and should allow it to get constant propagated etc.

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -39,32 +39,6 @@ const _COMPILED_DIR = Ref{String}()
 # abstract type UTC <: TimeZone end  # Already defined in the Dates stdlib
 abstract type Local <: TimeZone end
 
-function __init__()
-    # Write out our compiled tzdata representations into a scratchspace
-    desired_version = TZData.tzdata_version()
-
-    _COMPILED_DIR[] = if desired_version == TZJData.TZDATA_VERSION
-        TZJData.ARTIFACT_DIR
-    else
-        @info "Loading tzdata $desired_version"
-        TZData.build(desired_version, _scratch_dir())
-    end
-
-    # Load the pre-computed TZData into memory. Skip pre-fetching the first time
-    # TimeZones.jl is loaded by `deps/build.jl` as we have yet to compile the tzdata.
-    isdir(_COMPILED_DIR[]) && _reload_cache(_COMPILED_DIR[])
-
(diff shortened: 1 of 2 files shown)


src/TimeZones.jl
```diff
@@ -90,12 +64,44 @@ include("local.jl")
...
 include("rounding.jl")
-include("parse.jl")
-include("deprecated.jl")
 
 # Required to support Julia `VERSION < v"1.9"`
 if !isdefined(Base, :get_extension)
     include("../ext/TimeZonesRecipesBaseExt.jl")
 end
 
+function update_dates_extension()
```
omus, line 73: Can you move this function into "parse.jl"?
```suggestion
function _apply_parsing_extensions()
```

src/TimeZones.jl
```diff
@@ -90,12 +64,44 @@ include("local.jl")
...
+   # Load the pre-computed TZData into memory. Skip pre-fetching the first time
+   # TimeZones.jl is loaded by `deps/build.jl` as we have yet to compile the tzdata.
+   isdir(_COMPILED_DIR[]) && _reload_cache(_COMPILED_DIR[])
+
+   # Base extension needs to happen everytime the module is loaded (issue #24)
+   update_dates_extension()
+end
+
+update_dates_extension()
+const ISOZonedDateTimeFormat = DateFormat("yyyy-mm-ddTHH:MM:SS.ssszzz")
```
omus, line 102: Should be able to move this inside of "parse.jl" to keep things clean.

src/TimeZones.jl
```diff
@@ -90,12 +64,44 @@ include("local.jl")
...
+function update_dates_extension()
+    Dates.CONVERSION_SPECIFIERS['z'] = TimeZone
+    Dates.CONVERSION_SPECIFIERS['Z'] = TimeZone
+    Dates.CONVERSION_DEFAULTS[TimeZone] = ""
+    Dates.CONVERSION_TRANSLATIONS[ZonedDateTime] = (
+        Year, Month, Day, Hour, Minute, Second, Millisecond, TimeZone,
+    )
+end
+
+function __init__()
```
omus, line 82: I'm pretty sure the `__init__` didn't have to move but I'm fine with it being after the includes

omus (commented):
Some duplication with #457. Overall I'm happy with this change but would like to see some minor reorganization

PR #459, opened 2024-05-17
Enable Aqua tests

omus:


ericphanson:
This seems to have broken some CI builds:

```julia
ERROR: LoadError: LoadError: ArgumentError: Package TimeZones does not have RecipesBase in its dependencies:
- If you have TimeZones checked out for development and have
  added RecipesBase as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with TimeZones
```

https://github.com/jump-dev/Convex.jl/actions/runs/9180785040/job/25247128212?pr=684

PR #461, opened 2024-05-21
Restore RecipesBase dep to fix 1.6

ericphanson:
xref https://github.com/JuliaTime/TimeZones.jl/pull/459#issuecomment-2123418212

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TimeZones"
 uuid = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
-version = "1.16.0"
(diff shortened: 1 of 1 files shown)


odow (approved):
It can probably be removed from `extras` and `targets`?

ericphanson:
I'm not 100% sure, but it seems to have been there pre-#459 so I think it would make sense to leave it to keep this as a minimal revert of the broken bit

omus:
My bad. It's way to easy to miss this? Why can't I have the CI fail for TimeZones itself?

omus:
I've made another PR to add a CI check to ensure this doesn't happen again: https://github.com/JuliaTime/TimeZones.jl/pull/462

omus (approved):

PR #462, opened 2024-05-22
Ensure we don't break weakdeps

omus:
None of the CI checks currently in place ensure that we [maintain compatibility for older versions of Julia for `[weakdeps]`](https://pkgdocs.julialang.org/v1/creating-packages/#Transition-from-normal-dependency-to-extension). This adds a new GHA job to ensure these problems are caught before they make it into a release.

This should prevent this mistake from happening again: https://github.com/JuliaTime/TimeZones.jl/pull/459#issuecomment-2123418212

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -74,7 +74,7 @@ jobs:
       - shell: julia --color=yes --project=docs {0}
         run: |
           using Pkg
-          Pkg.add(PackageSpec(name="TimeZones", rev=ENV["SHA"])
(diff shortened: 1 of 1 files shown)


omus:
Appears that I need to use `Pkg.add` and cannot use `Pkg.develop` in order to catch this failure

omus:
Example of the job reproducing the failure: https://github.com/JuliaTime/TimeZones.jl/actions/runs/9184914310/job/25258000384

omus:
Looks like this check is working

PR #463, opened 2024-05-22
Defer loading the cache for constructing fixed time zones

omus:
Attempting to address this problem: https://github.com/apache/arrow-julia/pull/482#issuecomment-2123406667


omus:
I have some concerns about the lazy loading now. We seem to just be passing the buck on performing the cache initialization. I especially don't like it now that packages which use time zones may see bad timings with `@time_imports` from using `@tz_str`

PR #464, opened 2024-05-23
CI runner tweaks

omus:
- Avoid testing against Julia 1.6 on Windows as it is particularly slow
- Switch from testing against Intel based macOS to Apple Silicon macOS as the former is becoming less common (https://github.com/orgs/community/discussions/102846)
- Use CodeCov token to avoid rate limiting: https://docs.codecov.com/docs/codecov-uploader#supporting-token-less-uploads-for-forks-of-open-source-repos-using-codecov

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,6 +34,7 @@ jobs:
           - "1.6"  # LTS / Oldest supported version
           - "1"    # Latest release
           - nightly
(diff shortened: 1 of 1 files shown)


omus:
Noticed some codecov warnings: https://github.com/JuliaTime/TimeZones.jl/actions/runs/9209405477. I'll try to take care of those too

PR #457, opened 2024-04-11
Lazily initialize the time zone cache

KristofferC:
This avoids having to compile a bunch of code in `__init__` and improves the load time of the package (on Julia 1.11) from:

```julia
julia> @time_imports import TimeZones
...
              ┌ 2.6 ms TimeZones.TZData.__init__() 92.83% compilation time
               ├ 237.9 ms TimeZones.__init__() 94.11% compilation time (89% recompilation)
    276.3 ms  TimeZones 81.90% compilation time (88% recompilation)
```

```julia
julia> @time_imports import TimeZones
...
     17.5 ms  TimeZones 16.03% compilation time
```

The change in performance of the functions is (in my opinion) acceptable.

After:

```julia
julia> using BenchmarkTools, TimeZones

julia> @btime istimezone("Europe/Warsaw")
  48.068 ns (1 allocation: 16 bytes)
true
```

Before:

```julia
julia> @btime istimezone("Europe/Warsaw")
  45.356 ns (1 allocation: 16 bytes)
true
```

PR #479, opened 2024-10-29
Make `TimeZones.jl` relocatable

### LINKED ISSUE #467: Relocatability no longer given due to link to TZJData.ARTIFACT_DIR and storing path in const string reference
We have encountered a relocatability issue when creating a multi-stage docker container where we copy an app created by `PackageCompiler` to a clean image after having built it.

It now looks for the artifact in the old directory which was valid at build time and is no longer available in the final container.

The hash of the artifact is `520ce3f83be7fbb002cca87993a1b1f71fe10912` and I have verified that this is the hash of the artifact created when imporing `TZJData`.

Looking for this problem I have found a [post](https://github.com/JuliaLang/PackageCompiler.jl/issues/362) with a similar problem with a solution in the following [commit](https://github.com/JuliaLang/PackageCompiler.jl/pull/363/commits/649912a289e65705e83960aadba6ac56503f8ed7).

The problem seems to be the definition of a constant in `TZJData.jl`:

```julia
const ARTIFACT_DIR = artifact"tzjdata" 
```

which is referenced in `TimeZones.jl` (again, with the `const` keyword, but this time on the reference, so the value is very much mutable):

```julia
const _COMPILED_DIR = Ref{String}(TZJData.ARTIFACT_DIR)
```

According to the comment in the post mentioned above, "Is it because the const makes the path fixed at compilation, whereas we need it to load and find the artifact each time?", this seems to be a bad idea when trying to keep a package portable.

The solution seems to make both `const` definitions functions instead, i.e., in `TZJData.jl`:
```julia
function artifact_dir()
    return artifact"tzjdata" 
end
```
(not sure about the necessity to put this into a function in the first place), and in `TimeZones.jl`:
```julia
function compiled_dir()
    return TZJData.artifact_dir()
end
```

**BUT** I am not really sure how this works when building with a different `tzdata` version: In `TimeZones.jl`, it says:
```julia
function _build()
    desired_version = TZData.tzdata_version()
    if desired_version != TZJData.TZDATA_VERSION
        _COMPILED_DIR[] = TZData.build(desired_version, _scratch_dir())
    end

    return nothing
end
```

I have absolutely no clue how/whether this changed content of the referenced string ends up in my compiled app... and if it does (which I assume) how portability could still be ensured, because now it is definitely an incorrect path on my final docker image.

But I guess for the vanilla user not requesting a special version of `tzdata`, my solution would work, but I would need an extra string reference to the new compile dir if the user chooses a different version. That would be quite ugly and still not portable for special `tzdata` requirements...
```julia
const _CUSTOM_TZDATA_DIR = Ref{String}("")
function compiled_dir()
    return if isempty(_CUSTOM_TZDATA_DIR[])
        TZJData.artifact_dir()
    else
        _CUSTOM_TZDATA_DIR[]
    end
end
```

I assume / hope more experienced Julia programmers have better ideas?

lcontento:
Fixes #467 

Half of a possible solution to https://github.com/JuliaTime/TimeZones.jl/issues/467. The other half is at https://github.com/JuliaTime/TZJData.jl/pull/32. 

As discussed in the issue, there are many possible ways to fix this. I have tested that this approach works when including `TimeZones.jl` in a system image and relocating the depot. I think it should also be 100% equivalent to the old code in a normal setup. I cannot exclude that there are other relocatability issues hidden away, but even if this will not be the long term solution to the problem I think it is a decent fix for most people's use-cases.


omus:
Would be good to re-run and post the benchmarks from https://github.com/JuliaTime/TimeZones.jl/pull/457 around importing to see if this has an impact on load time.

src/TimeZones.jl
```diff
@@ -33,13 +33,16 @@ export TimeZone, @tz_str, istimezone, FixedTimeZone, VariableTimeZone, ZonedDate
...
-const _COMPILED_DIR = Ref{String}(TZJData.ARTIFACT_DIR)
+const _COMPILED_DIR = Ref{String}()
 
 # TimeZone types used to disambiguate the context of a DateTime
 # abstract type UTC <: TimeZone end  # Already defined in the Dates stdlib
 abstract type Local <: TimeZone end
 
 function __init__()
+    # Must be set at runtime to ensure relocatability 
+    _COMPILED_DIR[] = TZJData.artifact_dir()
```
omus, line 44: A option for backwards compat:
```suggestion
    # Backwards compatibility with TZJData v1.3.0 and below. Using older versions means TimeZones.jl is not relocatable
    _COMPILED_DIR[] = isdefined(TZJData, :artifact_dir) ? TZJData.artifact_dir() : TZJData.ARTIFACT_DIR
```

lcontento:
> Would be good to re-run and post the benchmarks from #457 around importing to see if this has an impact on load time.

Sure, will do.

src/TimeZones.jl
lcontento: If we want to ensure backwards compatibility then relocatability can only be enforced by adding a lower bound to the `TZJData` version. I would see adding a lower bound to the version of `TimeZones` as a more natural way to do it, since that's the package I expect people usually add to their `Project.toml`.
Unless there is sometimes the need for keeping old versions of `TZJData` pinned, in which case backwards compatibility becomes really necessary. I don't know the package well enough to say if that's the case.

DilumAluthge: Yeah, I think most downstream users are depending on TimeZones.jl, not TZJData.jl.

Is there a pressing need for TimeZones.jl to support old versions of TZJData.jl? If so, could we just backport https://github.com/JuliaTime/TZJData.jl/pull/32 to older versions of TZJData.jl?

omus: Part of the reason for the TimeZones.jl and TZJData.jl package separation is to allow for changing tzdata independently from the TimeZones.jl code. Setting a lower bound to TZJData.jl means we can't use those old versions. That's not the end of the world but the suggestion I had allows for relocatability and compatibility.

Backporting https://github.com/JuliaTime/TZJData.jl/pull/32 is possible but we would still need to update the compat in TimeZones.jl to allow for each of those patch releases.

lcontento: I have added a fallback for older versions by hardcoding the artifact hashes and querying them using `TZDATA_VERSION`. This is not very elegant but 1) there is no 100% foolproof solution to retrieve the hash AFAIK 2) there are not too many versions. In this way we can ensure relocatability even with the old TZJDatas.

omus: Do you care about relocatability for old TZJData versions? I'm fine with using `ARTIFACT_DIR` as the fallback. I'd rather break relocatability for old versions than break compat support unnecessarily (if that is a problem).

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Curtis Vogt <curtis.vogt@gmail.com>"]
 version = "1.19.0"
 
 [deps]
+Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
```
omus, line 7: Need to update the `docs/Manifest.toml`

src/TimeZones.jl
lcontento: I think the current fallback works with all old versions of TZJData, so there is no need to change the `[compat]` bounds.

lcontento: I have tested it for the last released version of TZJData on a system image and it works, even without the relocatability PR for TZJData.

Project.toml
lcontento: It works without a Manifest there. It seems reasonable to me but again I am not familiar with Documenter.

lcontento:
I tried to avoid the code coverage errors by excluding the new code, but the project level check still fails. The errors on Julia nightly were not there last time I looked at this, but they do not seem related to this PR.
@omus Is there anything else you would like to see improved here?

omus:
I'll try to get this PR moving today

omus:
We are good to move forward here. I'll need to take care of the failing `nightly` tests but I'll do that in another PR first

src/TimeZones.jl
```diff
@@ -42,23 +42,13 @@ abstract type Local <: TimeZone end
...
-            Base.SHA1("520ce3f83be7fbb002cca87993a1b1f71fe10912")
-        elseif TZJData.TZDATA_VERSION == "2023d"
-            Base.SHA1("b071cffdb310f5d3ca640c09cfa3dc3f23d450ad")
-        elseif TZJData.TZDATA_VERSION == "2023c"
-            Base.SHA1("52e48e96c4df04eeebc6ece0d9f1c3b545f0544c")
-        else
-            error("TZJData.jl with TZDATA_VERSION = $(TZJData.TZDATA_VERSION) is supposed to be relocatable!")
-        end
+        # Backwards compatibility for TZJData versions below v1.3.1
+        artifact_dict = Artifacts.parse_toml(joinpath(pkgdir(TZJData), "Artifacts.toml"))
```
lcontento, line 49: This will not work with a system image because `pkgdir(TZJData)` is the path of the TZJData package on the machine that created the system image.

lcontento: One could try to take the last part of the path and look into all depots until one finds the new package path. However, I am not 100% sure that having the source of the package in the depot is a requirement; I think you can probably run the system image without downloading the source. Hence, my original, but defintely not elegant, approach of hard-coding the hashes...

omus: Really? That seems like a bug as we're calling `pkgdir` at package initialization time so this should be looked up on the system which is running the sysimage

omus:
I think I would prefer to take the CodeCov hit and try to address the coverage drop later

omus:
I re-ran the benchmark on Julia 1.11.2:

With this PR:
```
julia> @time_imports import TimeZones
      0.7 ms  Printf
     17.2 ms  Dates
      0.4 ms  Scratch
      3.7 ms  InlineStrings
      0.5 ms  TZJData
               ┌ 0.1 ms p7zip_jll.__init__()
      3.7 ms  p7zip_jll
               ┌ 0.0 ms NetworkOptions.__init__()
      2.3 ms  NetworkOptions
      8.0 ms  ArgTools
               ┌ 0.3 ms nghttp2_jll.__init__()
      2.3 ms  nghttp2_jll
               ┌ 1.6 ms LibCURL_jll.__init__()
      3.4 ms  LibCURL_jll
               ┌ 0.0 ms MozillaCACerts_jll.__init__()
      2.0 ms  MozillaCACerts_jll
               ┌ 0.0 ms LibCURL.__init__()
      1.1 ms  LibCURL
               ┌ 2.9 ms Downloads.Curl.__init__()
     15.7 ms  Downloads
      0.3 ms  UUIDs
      0.4 ms  Compat
      0.3 ms  Compat → CompatLinearAlgebraExt
      0.4 ms  ExprTools
      0.5 ms  Mocking
               ┌ 3.6 ms TimeZones.TZData.__init__() 62.16% compilation time
               ├ 0.1 ms TimeZones.__init__()
     24.9 ms  TimeZones 8.91% compilation time

julia> using BenchmarkTools, TimeZones

julia> @btime istimezone("Europe/Warsaw");
  73.451 ns (1 allocation: 48 bytes)
```

Before this PR (with updated TZJData):
```
julia> @time_imports import TimeZones
      0.6 ms  Printf
     14.4 ms  Dates
      0.5 ms  Scratch
      2.9 ms  InlineStrings
      0.4 ms  TZJData
               ┌ 0.0 ms p7zip_jll.__init__()
      3.1 ms  p7zip_jll
               ┌ 0.0 ms NetworkOptions.__init__()
      1.8 ms  NetworkOptions
      7.2 ms  ArgTools
               ┌ 0.4 ms nghttp2_jll.__init__()
      2.4 ms  nghttp2_jll
               ┌ 1.6 ms LibCURL_jll.__init__()
      3.4 ms  LibCURL_jll
               ┌ 0.0 ms MozillaCACerts_jll.__init__()
      2.3 ms  MozillaCACerts_jll
               ┌ 0.0 ms LibCURL.__init__()
      1.4 ms  LibCURL
               ┌ 2.8 ms Downloads.Curl.__init__()
     16.3 ms  Downloads
      0.4 ms  UUIDs
      0.4 ms  Compat
      0.3 ms  Compat → CompatLinearAlgebraExt
      0.4 ms  ExprTools
      0.5 ms  Mocking
               ┌ 2.7 ms TimeZones.TZData.__init__() 79.32% compilation time
               ├ 0.0 ms TimeZones.__init__()
     24.2 ms  TimeZones 8.97% compilation time

julia> using BenchmarkTools, TimeZones

julia> @btime istimezone("Europe/Warsaw");
  73.484 ns (1 allocation: 48 bytes)
```

src/TimeZones.jl
lcontento: I guess it's a situation similar to `@__DIR__` which is also non-relocatable.

omus: Appears to be working as intended:
```sh
❯ JULIA_DEPOT_PATH=/tmp/depot julia -E 'using Pkg; Pkg.add(PackageSpec(name="TimeZones", version="1.19.0")); using TimeZones; TimeZones._COMPILED_DIR[]'
  Installing known registries into `/tmp/depot`
       Added `General` registry to /tmp/depot/registries
    Updating registry at `/tmp/depot/registries/General.toml`
   Resolving package versions...
   Installed Mocking ─────── v0.8.1
   Installed InlineStrings ─ v1.4.2
   Installed Compat ──────── v4.16.0
   Installed TZJData ─────── v1.3.0+2024b
   Installed ExprTools ───── v0.1.10
   Installed Scratch ─────── v1.2.1
   Installed TimeZones ───── v1.19.0
  Downloaded artifact: tzjdata
    Updating `/private/tmp/depot/environments/v1.11/Project.toml`
  [f269a46b] + TimeZones v1.19.0
    Updating `/private/tmp/depot/environments/v1.11/Manifest.toml`
  [34da2185] + Compat v4.16.0
  [e2ba6199] + ExprTools v0.1.10
  [842dd82b] + InlineStrings v1.4.2
  [78c3b35d] + Mocking v0.8.1
  [6c6a2e73] + Scratch v1.2.1
  [dc5dba14] + TZJData v1.3.0+2024b
  [f269a46b] + TimeZones v1.19.0
  [0dad84c5] + ArgTools v1.1.2
  [56f22d72] + Artifacts v1.11.0
  [ade2ca70] + Dates v1.11.0
  [f43a241f] + Downloads v1.6.0
  [7b1f6079] + FileWatching v1.11.0
  [b27032c2] + LibCURL v0.6.4
  [8f399da3] + Libdl v1.11.0
  [ca575930] + NetworkOptions v1.2.0
  [de0858da] + Printf v1.11.0
  [9a3f8284] + Random v1.11.0
  [ea8e919c] + SHA v0.7.0
  [fa267f1f] + TOML v1.0.3
  [cf7118a7] + UUIDs v1.11.0
  [4ec0a83e] + Unicode v1.11.0
  [deac9b47] + LibCURL_jll v8.6.0+0
  [29816b5a] + LibSSH2_jll v1.11.0+1
  [c8ffd9c3] + MbedTLS_jll v2.28.6+0
  [14a3606d] + MozillaCACerts_jll v2023.12.12
  [83775a58] + Zlib_jll v1.2.13+1
  [8e850ede] + nghttp2_jll v1.59.0+0
  [3f19e933] + p7zip_jll v17.4.0+2
Precompiling project...
  9 dependencies successfully precompiled in 7 seconds. 14 already precompiled.
"/tmp/depot/artifacts/7fdea2a12522469ca39925546d1fd93c10748180"

❯ mv /tmp/depot /tmp/depot2

❯ JULIA_DEPOT_PATH=/tmp/depot2 julia -E 'using TimeZones; TimeZones._COMPILED_DIR[]'
"/tmp/depot/artifacts/7fdea2a12522469ca39925546d1fd93c10748180"

❯ rm -rf /tmp/depot2

❯ JULIA_DEPOT_PATH=/tmp/depot julia -E 'using Pkg; Pkg.add(PackageSpec(url="https://github.com/lcontento/TimeZones.jl", rev="lc/relocatable")); using TimeZones; TimeZones._COMPILED_DIR[]'
  Installing known registries into `/tmp/depot`
       Added `General` registry to /tmp/depot/registries
     Cloning git-repo `https://github.com/lcontento/TimeZones.jl`
    Updating git-repo `https://github.com/lcontento/TimeZones.jl`
    Updating registry at `/tmp/depot/registries/General.toml`
   Resolving package versions...
   Installed Scratch ─────── v1.2.1
   Installed Compat ──────── v4.16.0
   Installed InlineStrings ─ v1.4.2
   Installed TZJData ─────── v1.3.0+2024b
   Installed Mocking ─────── v0.8.1
   Installed ExprTools ───── v0.1.10
  Downloaded artifact: tzjdata
    Updating `/private/tmp/depot/environments/v1.11/Project.toml`
  [f269a46b] + TimeZones v1.19.0 `https://github.com/lcontento/TimeZones.jl#lc/relocatable`
    Updating `/private/tmp/depot/environments/v1.11/Manifest.toml`
  [34da2185] + Compat v4.16.0
  [e2ba6199] + ExprTools v0.1.10
  [842dd82b] + InlineStrings v1.4.2
  [78c3b35d] + Mocking v0.8.1
  [6c6a2e73] + Scratch v1.2.1
  [dc5dba14] + TZJData v1.3.0+2024b
  [f269a46b] + TimeZones v1.19.0 `https://github.com/lcontento/TimeZones.jl#lc/relocatable`
  [0dad84c5] + ArgTools v1.1.2
  [56f22d72] + Artifacts v1.11.0
  [ade2ca70] + Dates v1.11.0
  [f43a241f] + Downloads v1.6.0
  [7b1f6079] + FileWatching v1.11.0
  [b27032c2] + LibCURL v0.6.4
  [8f399da3] + Libdl v1.11.0
  [ca575930] + NetworkOptions v1.2.0
  [de0858da] + Printf v1.11.0
  [9a3f8284] + Random v1.11.0
  [ea8e919c] + SHA v0.7.0
  [fa267f1f] + TOML v1.0.3
  [cf7118a7] + UUIDs v1.11.0
  [4ec0a83e] + Unicode v1.11.0
  [deac9b47] + LibCURL_jll v8.6.0+0
  [29816b5a] + LibSSH2_jll v1.11.0+1
  [c8ffd9c3] + MbedTLS_jll v2.28.6+0
  [14a3606d] + MozillaCACerts_jll v2023.12.12
  [83775a58] + Zlib_jll v1.2.13+1
  [8e850ede] + nghttp2_jll v1.59.0+0
  [3f19e933] + p7zip_jll v17.4.0+2
Precompiling project...
  9 dependencies successfully precompiled in 6 seconds. 14 already precompiled.
"/tmp/depot/artifacts/7fdea2a12522469ca39925546d1fd93c10748180"

❯ mv /tmp/depot /tmp/depot2

❯ JULIA_DEPOT_PATH=/tmp/depot2 julia -E 'using TimeZones; TimeZones._COMPILED_DIR[]'
"/tmp/depot2/artifacts/7fdea2a12522469ca39925546d1fd93c10748180"
```
Maybe there is something I'm forgetting about sysimages. Can you verify this works for your use case?

Project.toml
omus: Let's drop it. It is still needed for local dev but this means less to maintain

src/TimeZones.jl
lcontento: It does not work with system images. Here is a MWE
```bash
mkdir build_depot
mkdir sysimage_env
cd sysimage_env
JULIA_DEPOT_PATH=../build_depot julia -e 'using Pkg; Pkg.activate("."); Pkg.add(PackageSpec(name="TZJData", version=v"1.3.0")); Pkg.add(PackageSpec(url="https://github.com/lcontento/TimeZones.jl", rev="lc/relocatable")); Pkg.precompile()'
cd ..
mkdir build_env
cd build_env
JULIA_DEPOT_PATH=../build_depot julia -e 'using Pkg; Pkg.activate("."); Pkg.add("PackageCompiler"); Pkg.precompile()'
JULIA_DEPOT_PATH=../build_depot julia --project -e 'using PackageCompiler; create_sysimage(; project="../sysimage_env", sysimage_path="../sysimage.so")'
cd ..
mv build_depot sysimage_depot
JULIA_DEPOT_PATH=sysimage_depot julia -Jsysimage.so
```
The last command will fail because it will try accessing the `build_depot` folder which is no longer available.

lcontento: Unfortunately when using system images many things that one would expect to work no longer do...

omus: Thanks. I was also working on an example:
```
docker run -it --entrypoint=/bin/bash julia:1.11.2

apt-get update -qq
apt-get install -qq build-essential

cat >demo.jl <<EOF
using TimeZones
TimeZone("America/Winnipeg")
EOF

julia -e '
    using Pkg
    Pkg.add([
        PackageSpec(name="PackageCompiler", version="1"),
        PackageSpec(url="https://github.com/lcontento/TimeZones.jl", rev="lc/relocatable"), 
        PackageSpec(name="TZJData", version="1.3.0"),
    ])
    '

julia -e '
    using PackageCompiler
    PackageCompiler.create_sysimage(["TimeZones"]; sysimage_path="demo-sysimage.so", precompile_execution_file="demo.jl")
    '

echo "With sysimage"
julia -Jdemo-sysimage.so --trace-compile=stderr demo.jl

rm -rf /root/.julia/packages /root/.julia/compiled
echo "With sysimage after delete"
julia -Jdemo-sysimage.so --trace-compile=stderr demo.jl
```
The system image no longer requires the package contents or the `.ji` to be present in the depot. However, the artifact must still exist. The solution I have in place with `pkgdir` does allow `pkgdir` to point to the location where the `Artifact.toml` would be but will fail when trying to load the file.

omus: I suspect I'll roll back to using the hardcoded hashes but I'll do a little more experimentation first

omus: I came up with an improved MWE:
```
rm -rf build_depot build_env sysimage_depot sysimage_env
mkdir build_depot build_env sysimage_env
JULIA_DEPOT_PATH=build_depot julia --project=sysimage_env -e '
    using Pkg
    Pkg.add([
        PackageSpec(name="TZJData", version=v"1.3.0"),
        PackageSpec(url="https://github.com/JuliaTime/TimeZones.jl", rev="cv/relocatable"),
    ])
    Pkg.precompile()'
JULIA_DEPOT_PATH=build_depot julia --project=build_env -e '
    using Pkg
    Pkg.add(PackageSpec(name="PackageCompiler", version="2"))
    Pkg.precompile()
    using PackageCompiler
    create_sysimage(; project="sysimage_env", sysimage_path="sysimage.so")'
JULIA_DEPOT_PATH=build_depot julia -Jsysimage.so --project=sysimage_env -E 'using TimeZones; TimeZones._COMPILED_DIR[]'
mv build_depot sysimage_depot
JULIA_DEPOT_PATH=sysimage_depot julia -Jsysimage.so --project=sysimage_env -E 'using TimeZones; TimeZones._COMPILED_DIR[]'
```
It looks like we can use `Base.identify_package` and `Base.locate_package` to get the package location with the activated sysimage.

DilumAluthge: How long does it take to build a sysimage that contains TimeZones.jl? If it just takes a few minutes, could we maybe add a sysimage test to CI, so that we can catch any relocatability regressions in the future?

We could check for `ENV["CI"]`, so that people who run `Pkg.test("TimeZones")` locally don't have to run the test if they don't want to.

omus:
Hmmm, the CI sysimage test isn't failing when I expected it to

omus:
I've verified that the tests are failing when using `pkgdir` and older versions of TZJData: https://github.com/JuliaTime/TimeZones.jl/pull/479/commits/73da68afd6d40266b3d56a1df8f71c038a8ff9c5. Should be good to merge this after the CI wraps up. Thanks everyone!

omus (approved):

src/TimeZones.jl
omus: I ended up writing these tests directly in GHA. If I find this to be problematic I'll pull them out into the TimeZones tests directly

Issue #470, opened 2024-09-04
Enable ZonedDateTime to parse its own string representation


GHTaarn:
When a `ZonedDateTime` with a zero millisecond component is represented as a string, then the millisecond component is not shown in the `String` leading to `parse` errors if the `String` is parsed, e.g.:

```julia-repl
julia> x=ZonedDateTime("2024-09-03T18:42:59.000+00:00")
2024-09-03T18:42:59+00:00

julia> parse(ZonedDateTime, string(x))
ERROR: ArgumentError: Unable to parse date time. Expected directive Delim(.) at char 20
Stacktrace:
 [1] macro expansion
   @ /snap/julia/104/share/julia/stdlib/v1.10/Dates/src/parse.jl:104 [inlined]
 [2] tryparsenext_core(str::String, pos::Int64, len::Int64, df::DateFormat{Symbol("yyyy-mm-ddTHH:MM:SS.ssszzz"), Tuple{…}}, raise::Bool)
   @ Dates /snap/julia/104/share/julia/stdlib/v1.10/Dates/src/parse.jl:38
 [3] macro expansion
   @ /snap/julia/104/share/julia/stdlib/v1.10/Dates/src/parse.jl:150 [inlined]
 [4] tryparsenext_internal
   @ /snap/julia/104/share/julia/stdlib/v1.10/Dates/src/parse.jl:125 [inlined]
 [5] parse(::Type{ZonedDateTime}, str::String, df::DateFormat{Symbol("yyyy-mm-ddTHH:MM:SS.ssszzz"), Tuple{…}})
   @ Dates /snap/julia/104/share/julia/stdlib/v1.10/Dates/src/parse.jl:284
 [6] parse(::Type{ZonedDateTime}, str::String)
   @ Dates /snap/julia/104/share/julia/stdlib/v1.10/Dates/src/parse.jl:282
 [7] top-level scope
   @ REPL[228]:1
Some type information was truncated. Use `show(err)` to see complete types.

julia> 
```
As far as I can tell, `DateTime` from the `Dates` standard library is always able to parse its own output and in my opinion, it would be nice if `ZonedDateTime` could do this as well.

omus:
> As far as I can tell, `DateTime` from the `Dates` standard library is always able to parse its own output and in my opinion, it would be nice if `ZonedDateTime` could do this as well.

[...]

PR #483, opened 2025-01-25
ZonedDateTime can parse its own string representation 

GHTaarn:
This PR solves issue [#470](https://github.com/JuliaTime/TimeZones.jl/issues/470) in a way that formally preserves backwards compatibility of `TimeZones`.
The first commit is what solves the issue. The second commit is optional, but given that `Dates.default_format(::Type{ZonedDateTime})` is no longer used in the `TimeZones` package, not `public` and not documented anywhere, I feel that it would be most correct to delete it as this would still formally preserve backwards compatibility.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209258842306688

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -17,7 +17,7 @@ begin
     # Follows the ISO 8601 standard for date and time with an offset. See
     # `Dates.ISODateTimeFormat` for the `DateTime` equivalent.
     const ISOZonedDateTimeFormat = DateFormat("yyyy-mm-dd\\THH:MM:SS.ssszzz")
-    const NoMillisecondFormat = DateFormat("yyyy-mm-dd\\THH:MM:SSzzz")
+    const ISOZonedDateTimeNoMillisecondFormat = DateFormat("yyyy-mm-dd\\THH:MM:SSzzz")
 end
 
 Base.parse(::Type{ZonedDateTime}, str::AbstractString) = ZonedDateTime(str)
@@ -104,7 +104,7 @@ end
 
 function ZonedDateTime(str::AbstractString)
     res = tryparse(ZonedDateTime, str, ISOZonedDateTimeFormat)
-    isnothing(res) ?  ZonedDateTime(str, NoMillisecondFormat) : res
+    return isnothing(res) ? ZonedDateTime(str, ISOZonedDateTimeNoMillisecondFormat) : res
 end
 
 function ZonedDateTime(str::AbstractString, df::DateFormat)
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -20,7 +20,25 @@ begin
     const ISOZonedDateTimeNoMillisecondFormat = DateFormat("yyyy-mm-dd\\THH:MM:SSzzz")
 end
 
-Base.parse(::Type{ZonedDateTime}, str::AbstractString) = ZonedDateTime(str)
+function Base.parse(::Type{ZonedDateTime}, str::AbstractString)
+    res = tryparse(ZonedDateTime, str, ISOZonedDateTimeFormat)
+    return isnothing(res) ? parse(ZonedDateTime, str, ISOZonedDateTimeNoMillisecondFormat) : res
+end
+
+function Base.parse(::Type{ZonedDateTime}, str::AbstractString, df::DateFormat)
+    argtypes = Tuple{Type{<:TimeType},AbstractString,DateFormat}
+    try
+        invoke(parse, argtypes, ZonedDateTime, str, df)
+    catch e
+        if e isa ArgumentError
+            rethrow(ArgumentError(
+                "Unable to parse string \"$str\" using format $df. $(e.msg)"
+            ))
+        else
+            rethrow()
+        end
+    end
+end
 
 function tryparsenext_fixedtz(str, i, len, min_width::Int=1, max_width::Int=0)
     i == len && str[i] === 'Z' && return ("Z", i+1)
@@ -101,30 +119,6 @@ end
 function Dates.format(io::IO, d::DatePart{'Z'}, zdt, locale)
     write(io, string(zdt.zone))  # In most cases will be an abbreviation.
 end
-
-function ZonedDateTime(str::AbstractString)
-    res = tryparse(ZonedDateTime, str, ISOZonedDateTimeFormat)
-    return isnothing(res) ? ZonedDateTime(str, ISOZonedDateTimeNoMillisecondFormat) : res
-end
-
-function ZonedDateTime(str::AbstractString, df::DateFormat)
-    try
-        parse(ZonedDateTime, str, df)
-    catch e
-        if e isa ArgumentError
-            rethrow(ArgumentError(
-                "Unable to parse string \"$str\" using format $df. $(e.msg)"
-            ))
-        else
-            rethrow()
-        end
-    end
-end
-
-function ZonedDateTime(str::AbstractString, format::AbstractString; locale::AbstractString="english")
-    ZonedDateTime(str, DateFormat(format, locale))
-end
-
 
 """
     _parsesub_tzabbr(str, [i, len]) -> Union{Tuple{AbstractString, Integer}, Exception}
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -26,10 +26,10 @@ end
 When the `TimeZones` package is loaded, 2 extra character codes are available
 for constructing the `format` string:
 
-| Code       | Matches   | Comment                                                       |
-|:-----------|:----------|:--------------------------------------------------------------|
-| `z`        | +02:00    | Numeric UTC offset                                            |
-| `Z`        | GST, UTC  | Time zone abbreviation                                        |
+| Code       | Matches            | Comment                                               |
+|:-----------|:-------------------|:------------------------------------------------------|
+| `z`        | +02:00, -0100, +14 | Parsing matches a fixed numeric UTC offset `±hh:mm`, `±hhmm`, or `±hh`. Formatting outputs `±hh:mm` |
+| `Z`        | UTC, GMT, America/New_York | Name of a time zone as specified in the IANA tz database |
 """ DateFormat
 
 function Base.parse(::Type{ZonedDateTime}, str::AbstractString)
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -16,8 +16,8 @@ begin
 
     # Follows the ISO 8601 standard for date and time with an offset. See
     # `Dates.ISODateTimeFormat` for the `DateTime` equivalent.
-    const ISOZonedDateTimeFormat = DateFormat("yyyy-mm-ddTHH:MM:SS.ssszzz")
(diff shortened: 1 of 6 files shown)


omus:
I took a quick look and I think this is a good approach to move things in the right direction. I'll do a full review this week.

iamed2:
Is it possible to have a solution that removes the need for [this](https://github.com/JuliaDatabases/LibPQ.jl/blob/ae7e231fe9f2db88cf731de84a471fac8e168436/src/parsing.jl#L270-L287) (allow parsing variable-length fractional settings like parsing DateTime does)?

Also, keep in mind years that are not four digits long will change the length of the input.

GHTaarn:
Thank you for your feedback @iamed2 , very good point about non 4 digit years, I have fixed this so that this aspect of the ISO standard is now supported. I felt that the code looked nicer if I used `tryparse`. The 3 argument version that I used is defined in `Dates` and is undocumented, I hope that this is okay since `TimeZones` already uses other undocumented features of the `Dates` package. I am considering submitting a PR suggesting some documentation for this method.

I cannot see an easy solution to allow the fractional seconds to have variable length like in `Dates.DateTime` and `Dates.Time`, the best solutions that I can see would be either for Julia to enhance `Dates.DateFormat` to be able to handle this or for `TimeZones` to use a custom format type for `ZonedDateTime`.

omus:
After really getting into the code I noticed a couple of things which I went ahead and changed here:

1. The `parse` function is intended to be the primary way for parsing a string into a `ZonedDateTime`. I've done some refactoring to reflect that.
2. The default format now enforces a literal `T` is found. This is mainly precautionary in case another package registers `T` as a character to parse.
3. I've updated the heuristic determining which format to use to be based upon if a period can be found in the string. That should be faster than parsing the string twice.
4. I've added more exhaustive tests here to ensure the underlying Dates parsing machinery doesn't silently break this functionality.

src/parse.jl
```diff
@@ -10,12 +10,39 @@ function init_dates_extension()
...
+function Base.parse(::Type{ZonedDateTime}, str::AbstractString)
+    # Works as the format should only contain a period when milliseconds are included
+    return if contains(str, '.')
+        parse(ZonedDateTime, str, ISOZonedDateTimeFormat)
+    else
+        parse(ZonedDateTime, str, ISOZonedDateTimeNoMillisecondFormat)
+    end
 end
 
-Dates.default_format(::Type{ZonedDateTime}) = ISOZonedDateTimeFormat
```
omus, line 18: As you said this isn't documented anywhere so it's technically not public. I think it's safest to deprecate this in the off chance anyone is using this.

GHTaarn:
@omus : It is late now, I will have a look a your changes tomorrow

omus:
> @omus : It is late now, I will have a look a your changes tomorrow

Sounds good. Feel free to review next week instead if you prefer. I plan on merging this on Monday and making a new release.

omus:
> Is it possible to have a solution that removes the need for [this](https://github.com/JuliaDatabases/LibPQ.jl/blob/ae7e231fe9f2db88cf731de84a471fac8e168436/src/parsing.jl#L270-L287) (allow parsing variable-length fractional settings like parsing DateTime does)?

IIRC the variable-length fractional parsing used by DateTime requires that it the variable portion occurs at the end of the string. We would need to make modifications to Dates.jl for that or roll our own solution here. I'm not opposed to doing that but it's a larger lift that what we have here.

GHTaarn:
>     1. The `parse` function is intended to be the primary way for parsing a string into a `ZonedDateTime`. I've done some refactoring to reflect that.
> 
>     2. The default format now enforces a literal `T` is found. This is mainly precautionary in case another package registers `T` as a character to parse. This also reflects what is used for `Dates.ISODateTimeFormat`.

I am perfectly fine with these two.
 
>     3. I've updated the heuristic determining which format to use to be based upon if a period can be found in the string. That should be faster than parsing the string twice.

I feel that this change makes the code more vulnerable to changes in `ISOZonedDateTimeFormat`, e.g. if the ISO standard were to change in the future and allow '.' characters in other places like allowing milliseconds in the offset. I ran some benchmarks and did not see any performance improvement from this change. I also searched the Julia documentation and found the 3 argument `tryparse` mentioned [here](https://docs.julialang.org/en/v1/stdlib/Dates/#Constructors), so we should be able to rely on it being present in future versions of Julia. For these reasons, I prefer the `tryparse` based solution that I proposed.

>     4. I've added more exhaustive tests here to ensure the underlying Dates parsing machinery doesn't silently break this functionality.

Yes, this is certainly an improvement. I felt that it would be nice to have a year where the ISO standard really applies, a millisecond value ending with a `0` and at least one time zone that is not UTC, so I added these to the testset.

Finally, I added some docstrings for the `ZonedDateTime` parse constructors and an extra `TimeZones` specific docstring for `Dates.DateFormat`. The latter is of course not really in the scope of this PR and I don't know if this is an acceptable way to do this, I just felt that it would be nice to have this info in a docstring.

Your decision to deprecate rather than remove `Dates.default_format` is also fine with me.

omus:
> > ```
> > 3. I've updated the heuristic determining which format to use to be based upon if a period can be found in the string. That should be faster than parsing the string twice.
> > ```
> 
> I feel that this change makes the code more vulnerable to changes in `ISOZonedDateTimeFormat`, e.g. if the ISO standard were to change in the future and allow '.' characters in other places like allowing milliseconds in the offset. I ran some benchmarks and did not see any performance improvement from this change. I also searched the Julia documentation and found the 3 argument `tryparse` mentioned [here](https://docs.julialang.org/en/v1/stdlib/Dates/#Constructors), so we should be able to rely on it being present in future versions of Julia. For these reasons, I prefer the `tryparse` based solution that I proposed.

There is a noticeable performance difference although it's not huge:

```julia
julia> using TimeZones, BenchmarkTools

julia> using TimeZones: ISOZonedDateTimeFormat, ISOZonedDateTimeNoMillisecondFormat

julia> function p1(str)
           return if contains(str, '.')
               parse(ZonedDateTime, str, ISOZonedDateTimeFormat)
           else
               parse(ZonedDateTime, str, ISOZonedDateTimeNoMillisecondFormat)
           end
       end
p1 (generic function with 1 method)

julia> function p2(str)
           res = tryparse(ZonedDateTime, str, ISOZonedDateTimeFormat)
           return res !== nothing ? res : parse(ZonedDateTime, str, ISOZonedDateTimeNoMillisecondFormat)
       end
p2 (generic function with 1 method)

julia> @btime p1("2000-01-02T03:04:05-07:00")
  938.524 ns (26 allocations: 1.07 KiB)
2000-01-02T03:04:05-07:00

julia> @btime p1("2000-01-02T03:04:05-07:00")
  932.524 ns (26 allocations: 1.07 KiB)
2000-01-02T03:04:05-07:00

julia> @btime p2("2000-01-02T03:04:05-07:00")
  966.650 ns (26 allocations: 1.07 KiB)
2000-01-02T03:04:05-07:00

julia> @btime p1.(fill("2000-01-02T03:04:05-07:00",1000));
  925.000 μs (26006 allocations: 1.10 MiB)

julia> @btime p2.(fill("2000-01-02T03:04:05-07:00",1000));
  958.959 μs (26006 allocations: 1.10 MiB)
```

You are correct that the `tryparse` approach is more adaptable to changes to `ISOZonedDateTimeFormat` but as the  original format is based off of an ISO standard we are fairly safe to say it's not going to substantially change. One aspect of the `tryparse` approach that concerns me is that if we did change the format constant and it would start failing by using `tryparse` we wouldn't see any error message when failing to parse the first format which could lead to confusion for end users. 

No matter the approach we take to parsing this dual format the test suite that is in place should ensure reliable parsing so changes to the `ISOZonedDateTimeFormat` which break parsing or the `.` heuristic will not go unnoticed.

src/types/zoneddatetime.jl
```diff
@@ -166,6 +166,29 @@ function ZonedDateTime(date::Date, args...; kwargs...)
...
 end
 
+# Parsing constructors
+
+"""
+    ZonedDateTime(str::AbstractString)
+
+Construct a `ZonedDateTime` by parsing `str`. This method is designed so that
+`zdt == ZonedDateTime(string(zdt))` where `zdt` can be any `ZonedDateTime`
+object.
```
omus, line 176:
````suggestion
Construct a `ZonedDateTime` by parsing `str`. This method is designed so that
`zdt == ZonedDateTime(string(zdt))` where `zdt` can be any `ZonedDateTime`
object. Take note that this method will always result creating a `ZonedDateTime`
with a `FixedTimeZone` which may can result in different results with date/time
arithmetic.

## Examples
```jltest
julia> zdt = ZonedDateTime(2025, 3, 8, 9, tz"America/New_York")
2025-03-08T09:00:00-05:00

julia> timezone(zdt)
America/New_York (UTC-5/UTC-4)

julia> zdt + Day(1)
2025-03-09T09:00:00-04:00

julia> pzdt = ZonedDateTime(string(zdt))
2025-03-08T09:00:00-05:00

julia> timezone(pzdt)
UTC-05:00

julia> pzdt + Day(1)
2025-03-09T09:00:00-05:00
```
````

File: src/parse.jl [hunk 3]
```diff
@@ -10,12 +10,51 @@ function init_dates_extension()
...
+@doc """
+    DateFormat(format::AbstractString, locale="english") --> DateFormat
+
+When the `TimeZones` package is loaded, 2 extra character codes are available
+for constructing the `format` string:
+
+| Code       | Matches   | Comment                                                       |
+|:-----------|:----------|:--------------------------------------------------------------|
+| `z`        | +02:00    | Numeric UTC offset                                            |
+| `Z`        | GST, UTC  | Time zone abbreviation                                        |
```
omus, line 32:
```suggestion
| Code       | Matches            | Comment                                               |
|:-----------|:-------------------|:------------------------------------------------------|
| `z`        | +02:00, -0100, +14 | Parsing matches a fixed numeric UTC offset `±hh:mm`, `±hhmm`, or `±hh`. Formatting outputs `±hh:mm` |
| `Z`        | UTC, GMT, America/New_York | Name of a time zone as specified in the IANA tz database |
```

omus (approved):

src/types/zoneddatetime.jl
GHTaarn: I don't know how to do this in Github, I just wanted to correct the two typos in lines
176 and 177 e.g. with
```markdown
object. Take note that this method will always create a `ZonedDateTime`
with a `FixedTimeZone` which can result in different results with date/time
```
Other than that, these changes are fine with me.

src/parse.jl
GHTaarn: Fine with me

GHTaarn:
@omus  Regarding `Base.parse(::Type{ZonedDateTime}, ::AbstractString)`: Let us keep your version with `contains`.

omus:
Will merge after CI completes

PR #484, opened 2025-01-31
Make CodeCov project changes informational

omus:
I don't think we should see failed PR status checks when code coverage decreases. We still get good visibility into lines where coverage is missed as those still show up as annotations when viewing the diff


omus:
Updated this org setting to allow tokenless upload (tokens can still be used): https://app.codecov.io/account/gh/JuliaTime/org-upload-token

omus:
It appears that this documentation isn't quite right as a `.yaml` extension does work:

> Can I name the file .codecov.yml?
>
> Yes, you can name the file `codecov.yml` or `.codecov.yml`. However, the file must still be located in the repository root, `dev/`, or `.github/` directories
– https://docs.codecov.com/docs/codecov-yaml#can-i-name-the-file-codecovyml

omus:
I've made a suggestion to the codecov documentation.

omus:
I've confirmed that both `.codecov.yaml` and `codecov.yaml` are supported.

.github/workflows/CI.yml
```diff
@@ -78,10 +78,9 @@ jobs:
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
-      - uses: codecov/codecov-action@v4
+      - uses: codecov/codecov-action@v5
         with:
           files: lcov.info
-          token: ${{ secrets.CODECOV_TOKEN }}
```
omus, line 84: I've deleted the secret from the repository

PR #497, opened 2025-10-01
Restore `Dates.default_format(::Type{ZonedDateTime})`

omus:
Was removed in https://github.com/JuliaTime/TimeZones.jl/pull/483 as we believed this was only used with parsing. We did discover that there were some legitimate use cases of this with `Dates.format`: https://github.com/quinnj/JSON3.jl/issues/313. I thought maybe we could modify JSON3.jl to work around this but while implementing that change I changed my mind on what was the right approach.

As of now I think that:
- Most users should avoid using `Dates.default_format` for parsing as they won't be able to parse `ZonedDateTime` strings without milliseconds. They should use `parse(ZonedDateTime, str)` instead.
- When using `Date.format` it is reasonable to use `Dates.default_format(::Type{ZonedDateTime})`.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -21,14 +21,4 @@ const TransitionTimeInfo = TZFile.TransitionTimeInfo
 
 @deprecate build(; force=false) build(TZJData.TZDATA_VERSION; force)
 
-function Dates.default_format(::Type{ZonedDateTime})
-    depwarn(
-        "`Dates.default_format(ZonedDateTime)` is deprecated and has no direct " *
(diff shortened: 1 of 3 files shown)


omus:
Nightly failure for x86 is unrelated:
```
allocations: Test Failed at /home/runner/work/TimeZones.jl/TimeZones.jl/test/types/timezone.jl:45
  Expression: #= /home/runner/work/TimeZones.jl/TimeZones.jl/test/types/timezone.jl:45 =# @allocations(istimezone("America/Winnipeg")) == 1
   Evaluated: 0 == 1
```

PR #499, opened 2025-11-26
Defer creating the scratch directory until needed

### LINKED ISSUE #498: TZData always writes (empty) scratchspace folder
When one uses TimeZones.jl it seems that a scratchfolder is always created due to the use of `_scratch_dir()` in `init` at [`src\tzdata\TZData.jl:19`](https://github.com/JuliaTime/TimeZones.jl/blob/4d5f314ce6b01d7c0681b2c915282263f9989e44/src/tzdata/TZData.jl#L19).

Always needing to write to the filesystem seems unnecessary as the scratch folder stays empty, and it is a bad practice for packages in general. Furthermore, this makes it impossible to use TimeZones in an environment without write permissions, as encountered in https://github.com/Deltares/Ribasim/pull/2710. In that PackageCompiler setting, given that all required timezone data is already in the `share/julia/artefacts` folder, it should be able to skip the scratchfolder completely.

There are other places that call `_scratch_dir`, but I'm not sure if these are as problematic.

omus:
Fixes https://github.com/JuliaTime/TimeZones.jl/issues/498. Should improve compatibility when installed into immutable environments when the package has never been initialized.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -148,6 +148,10 @@ jobs:
         run: |
           chmod -R ugo-w build-depot
           find build-depot -perm /u+w,g+w,o+w  # List directories/files with any of these permission bits
+          if ! touch build-depot/write-test 2>/dev/null; then
+              echo "Depot is not immutable" >&2
+              exit 1
+          fi
       - name: Validate sysimage works
         shell: julia --color=yes --project=sysimage -Jsysimage.so {0}
         run: |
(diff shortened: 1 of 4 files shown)


omus:
Testing this is turning out to be quite annoying. I'm not sure why `DEPOT/artifacts/Overrides.toml` seems to be needed. I'll grant access and figure things out from there

evetion:
Wow, thanks for the very quick fix and extensive test effort that went into it! 👍🏻

PR #502, opened 2025-11-27
Eliminate allocation with `ZonedDateTime` constructor using `is_dst::Bool`

### LINKED ISSUE #495: `ZonedDateTime` allocates memory when using disambiguating method with `Bool`s (but not with `Int`s)
```julia-repl
julia> using TimeZones

julia> @time ZonedDateTime(2025, 7, 18, 22, 5, tz"Europe/Berlin")
  0.000002 seconds
2025-07-18T22:05:00+02:00

julia> @time ZonedDateTime(2025, 7, 18, 22, 5, tz"Europe/Berlin", 1)
  0.000002 seconds
2025-07-18T22:05:00+02:00

julia> @time ZonedDateTime(2025, 7, 18, 22, 5, tz"Europe/Berlin", true)
  0.000001 seconds (1 allocation: 48 bytes)
2025-07-18T22:05:00+02:00
```
The last method allocates memory.  It'd be nice to remove the memory allocation, since the other ones don't have it

omus:
Fixes #495. We now no longer see allocations when using `ZonedDateTime(dt, tz, is_dst::Bool)` when the local date/time is not ambiguous. We still see allocations when it is ambiguous though. 

```julia
julia> VERSION
v"1.12.2"

julia> @allocated ZonedDateTime(DateTime(2025, 7, 18, 22, 5), tz"Europe/Berlin", true)  # unambiguous
0

julia> @allocated ZonedDateTime(DateTime(2025, 10, 26, 2), tz"Europe/Berlin", 1)  # ambiguous
0

julia> @allocated ZonedDateTime(DateTime(2025, 10, 26, 2), tz"Europe/Berlin", 2) # ambiguous
0

julia> @allocated ZonedDateTime(DateTime(2025, 10, 26, 2), tz"Europe/Berlin", true) # ambiguous
64

julia> @allocated ZonedDateTime(DateTime(2025, 10, 26, 2), tz"Europe/Berlin", false) # ambiguous
64
```

### DIFF
--- a/src/types/zoneddatetime.jl
+++ b/src/types/zoneddatetime.jl
@@ -92,12 +92,12 @@ ambiguous within the given time zone you can set `is_dst` to resolve the ambigui
     elseif num == 0
         throw(NonExistentTimeError(dt, tz))
     elseif num == 2
-        possible = [first(possible), last(possible)]
         mask = [isdst(zdt.zone.offset) for zdt in possible]
 
         # Mask is expected to be unambiguous.
         !xor(mask...) && throw(AmbiguousTimeError(dt, tz))
 
+        # Since the mask is unambiguous we can pick the ZonedDateTime which matches `is_dst`
         return mask[1] == is_dst ? possible[1] : possible[2]
     else
         throw(AmbiguousTimeError(dt, tz))
--- a/src/types/zoneddatetime.jl
+++ b/src/types/zoneddatetime.jl
@@ -83,7 +83,7 @@ end
 Construct a `ZonedDateTime` by applying a `TimeZone` to a `DateTime`. If the `DateTime` is
 ambiguous within the given time zone you can set `is_dst` to resolve the ambiguity.
 """
-function ZonedDateTime(dt::DateTime, tz::VariableTimeZone, is_dst::Bool)
+@inline function ZonedDateTime(dt::DateTime, tz::VariableTimeZone, is_dst::Bool)
     possible = interpret(dt, tz, Local)
 
     num = length(possible)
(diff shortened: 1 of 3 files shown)


omus:
Julia 1.10.10, 1.11.7, 1.12.2, and 1.14.0-DEV.1310 all report this on the REPL:
```julia
julia> @allocated ZonedDateTime(DateTime(2025, 7, 18, 22, 5), tz"Europe/Berlin", true)
0
```
I'm not sure why I see allocations in the tests on Julia 1.12.2 and 1.14.0-DEV.1310 but this overall seems like an improvement.

src/types/zoneddatetime.jl
```diff
@@ -92,12 +92,12 @@ ambiguous within the given time zone you can set `is_dst` to resolve the ambigui
     elseif num == 0
         throw(NonExistentTimeError(dt, tz))
     elseif num == 2
-        possible = [first(possible), last(possible)]
```
omus: Turns out I didn't need this after all

giordano:
> I'm not sure why I see allocations in the tests on Julia 1.12.2 and 1.14.0-DEV.1310 but this overall seems like an improvement.

Code coverage is effectful and in v1.12+ it tends to cause extra allocations.

omus:
> > I'm not sure why I see allocations in the tests on Julia 1.12.2 and 1.14.0-DEV.1310 but this overall seems like an improvement.
> 
> Code coverage is effectful and in v1.12+ it tends to cause extra allocations.

Good to know. That may not be the root cause here though since `Pkg.test(coverage=true)` and `Pkg.test(coverage=false)` have the same allocations.

Issue #469, opened 2024-09-04
Provide user friendly exceptions with deprecated time zones


omus:
As discussed in #419 when a time zone becomes deprecated as "Europe/Amsterdam" was as of tzdata release 2022f they see the following error:

```julia
julia> using TimeZones

julia> tz"Europe/Amsterdam"
ERROR: LoadError: ArgumentError: The time zone "Europe/Amsterdam" is of class `TimeZones.Class(:LEGACY)` which is currently not allowed by the mask: `TimeZones.Class(:FIXED) | TimeZones.Class(:STANDARD)`
Stacktrace:
 [1] TimeZone(str::String, mask::TimeZones.Class)
   @ TimeZones ~/.julia/dev/TimeZones/src/types/timezone.jl:51
 [2] TimeZone(str::String)
   @ TimeZones ~/.julia/dev/TimeZones/src/types/timezone.jl:42
 [3] var"@tz_str"(__source__::LineNumberNode, __module__::Module, str::Any)
   @ TimeZones ~/.julia/dev/TimeZones/src/types/timezone.jl:72
in expression starting at REPL[17]:1
```

This error does not provide the user with the details of any possible replacement for the deprecated time zone. The `backward` region file contains this information and for [2022f](https://data.iana.org/time-zones/tzdb-2022f/backward) it states that "Europe/Brussels" is the replacement for "Europe/Amsterdam". Providing this information to users would allow an easy upgrade path instead of having them rely on a legacy time zone via `TimeZones("Europe/Amsterdam", TimeZones.Class(:LEGACY))` or choose an [alternate time zone which may be incorrect](https://github.com/JuliaTime/TimeZones.jl/issues/419#issuecomment-2111818142).

omus:
> Deprecations are commonly included in minor version bumps for Julia package. The difference here is that TimeZones.jl doesn't allow the use of legacy/deprecated time zones by default. I could see an argument to be made for tying the default TimeZone.Class with the `--depwarn` flag and having `--depwarn=error` be the only mode where TimeZone.Class(:LEGACY) is disallowed.
– https://github.com/JuliaTime/TZJData.jl/issues/31#issuecomment-2341738128

PR #504, opened 2026-01-07
[RFC] `:LEGACY` typezones with deprecation warnings.

rofinn:
Related to #469 and #419.

## Background

We already have access to links when compiling tzjfiles, but we currently don't retain that information for TimeZones.jl to use. The proposed solution is to:

1. Optionally include timezone link targets in our TZJFile tuples
2. Update the `TimeZone` constructor to auto-redirect legacy timezones with the link info and throw a `depwarn`, so errors are limited to running with `--depwarn=error`.

## Usage

On Julia v1.12.3

Basic:
```sh
> julia --project=.

julia> using TimeZones

julia> tz"Europe/Amsterdam"
Europe/Brussels (UTC+1/UTC+2)
```

Error:
```sh
> julia --depwarn=error --project=.

julia> using TimeZones

julia> tz"Europe/Amsterdam"
ERROR: LoadError: The time zone "Europe/Amsterdam" is deprecated, using "Europe/Brussels" instead.
Stacktrace:
 [1] _depwarn(msg::Any, funcsym::Any, force::Bool)
   @ Base ./deprecated.jl:262
 [2] #invokelatest_gr#232
   @ ./reflection.jl:1295 [inlined]
 [3] invokelatest_gr
   @ ./reflection.jl:1289 [inlined]
 [4] #depwarn#953
   @ ./deprecated.jl:256 [inlined]
 [5] depwarn
   @ ./deprecated.jl:251 [inlined]
 [6] TimeZone(str::String, mask::TimeZones.Class)
   @ TimeZones ~/repos/invenia/TimeZones.jl/src/types/timezone.jl:56
 [7] TimeZone(str::String)
   @ TimeZones ~/repos/invenia/TimeZones.jl/src/types/timezone.jl:42
 [8] var"@tz_str"(__source__::LineNumberNode, __module__::Module, str::Any)
   @ TimeZones ~/repos/invenia/TimeZones.jl/src/types/timezone.jl:85
in expression starting at REPL[2]:1
```

V1 (same behaviour):
```sh
> JULIA_TZJ_VERSION=1 julia --project=.

julia> using TimeZones, Dates

julia> tz"Europe/Amsterdam"
ERROR: LoadError: ArgumentError: The time zone "Europe/Amsterdam" is of class `TimeZones.Class(:LEGACY)` which is currently not allowed by the mask: `TimeZones.Class(:FIXED) | TimeZones.Class(:STANDARD)`
Stacktrace:
 [1] TimeZone(str::String, mask::TimeZones.Class)
   @ TimeZones ~/repos/invenia/TimeZones.jl/src/types/timezone.jl:64
 [2] TimeZone(str::String)
   @ TimeZones ~/repos/invenia/TimeZones.jl/src/types/timezone.jl:42
 [3] var"@tz_str"(__source__::LineNumberNode, __module__::Module, str::Any)
   @ TimeZones ~/repos/invenia/TimeZones.jl/src/types/timezone.jl:85
in expression starting at REPL[2]:1
```

## Decisions Summary

- `Union{String, Nothing}` in the compiled files for simplicity.
- We are **not handling the recursive case for links** (a -> b -> c) and assuming (a -> c, b -> c) as stated in the `backward` file header.
- To keep the cache tuples `isbits`, I'm reusing InlineStrings.jl with an empty string instead of `nothing`. Looks like `InlineString31` should fit all the names, but it does increase memory usage per cache entry.
- The JULIA_TZJ_VERSION feature flag and compiled path changes are just some hacks to support switching between formats for testing the behaviour. It has been handy for testing, but could be reverted. Also, maybe there's an easier way I'm unfamiliar with?

### DIFF
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -5,7 +5,7 @@ using Dates
 using Printf
 using Scratch: @get_scratch!
 using Unicode
-using InlineStrings: InlineString15
+using InlineStrings: InlineString15, InlineString31
 using TZJData: TZJData
 
 import Dates: TimeZone, UTC
--- a/src/TimeZones.jl
+++ b/src/TimeZones.jl
@@ -42,7 +42,12 @@ abstract type Local <: TimeZone end
 
 function __init__()
     # Set at runtime to ensure relocatability
-    _COMPILED_DIR[] = @static if isdefined(TZJData, :artifact_dir)
+    # Prefer scratch-compiled directory with matching versions, fall back to artifact
+    expected_dir = TZData.compiled_dir()
+
+    _COMPILED_DIR[] = if isdir(expected_dir)
+        expected_dir
+    elseif isdefined(TZJData, :artifact_dir)
         TZJData.artifact_dir()
     else
         # Backwards compatibility for TZJData versions below v1.3.1. The portion of the
--- a/src/types/timezone.jl
+++ b/src/types/timezone.jl
@@ -39,9 +39,9 @@ US/Pacific (UTC-8/UTC-7)
 TimeZone(::AbstractString, ::Class)
 
 function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
-    tz, class = get(_TZ_CACHE, str) do
+    tz, class, link = get(_TZ_CACHE, str) do
         if occursin(FIXED_TIME_ZONE_REGEX, str)
-            FixedTimeZone(str), Class(:FIXED)
+            FixedTimeZone(str), Class(:FIXED), InlineString31("")
         else
             throw(ArgumentError("Unknown time zone \"$str\""))
         end
@@ -83,7 +83,7 @@ function istimezone(str::AbstractString, mask::Class=Class(:DEFAULT))
         return true
     end
 
-    # Checks against pre-compiled time zones
-    class = get(() -> (UTC_ZERO, Class(:NONE)), _TZ_CACHE, str)[2]
+    # Checks against pre-compiled time zones (3-tuple now: tz, class, link)
+    class = get(() -> (UTC_ZERO, Class(:NONE), InlineString31("")), _TZ_CACHE, str)[2]
     return mask & class != Class(:NONE)
 end
--- a/src/types/timezone.jl
+++ b/src/types/timezone.jl
@@ -47,6 +47,19 @@ function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
         end
     end
 
+    # Auto-redirect LEGACY timezones to their modern equivalents
+    # Only when user hasn't explicitly opted in to LEGACY class
+    if !isempty(link) && class == Class(:LEGACY) && mask & Class(:LEGACY) == Class(:NONE)
+        # Note: Using depwarn here allows users to control behavior via --depwarn flag.
+        # With --depwarn=error, this becomes an error (strict mode).
+        # This matches the behavior requested in issue #469 https://github.com/JuliaTime/TimeZones.jl/issues/469#issuecomment-2341741754.
+        Base.depwarn(
+            "The time zone \"$str\" is deprecated, using \"$link\" instead.",
+            :TimeZone
+        )
+        return TimeZone(String(link), mask)
+    end
+
     if mask & class == Class(:NONE)
         throw(ArgumentError(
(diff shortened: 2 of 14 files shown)


src/TimeZones.jl
```diff
@@ -42,7 +42,12 @@ abstract type Local <: TimeZone end
 
 function __init__()
     # Set at runtime to ensure relocatability
-    _COMPILED_DIR[] = @static if isdefined(TZJData, :artifact_dir)
+    # Prefer scratch-compiled directory with matching versions, fall back to artifact
```
rofinn, line 45: Part of the version feature flag and path hacks for testing both v1 and v2 behaviour.

src/types/timezone.jl
```diff
@@ -39,14 +39,27 @@ US/Pacific (UTC-8/UTC-7)
 TimeZone(::AbstractString, ::Class)
 
 function TimeZone(str::AbstractString, mask::Class=Class(:DEFAULT))
-    tz, class = get(_TZ_CACHE, str) do
+    tz, class, link = get(_TZ_CACHE, str) do
         if occursin(FIXED_TIME_ZONE_REGEX, str)
-            FixedTimeZone(str), Class(:FIXED)
+            FixedTimeZone(str), Class(:FIXED), InlineString31("")
```
rofinn, line 44: Unfortunately, a `Union` with `nothing` is not `isbits`, hence the empty `InlineString31`.

src/types/timezone.jl
```diff
@@ -39,14 +39,27 @@ US/Pacific (UTC-8/UTC-7)
...
+    # Only when user hasn't explicitly opted in to LEGACY class
+    if !isempty(link) && class == Class(:LEGACY) && mask & Class(:LEGACY) == Class(:NONE)
+        # Note: Using depwarn here allows users to control behavior via --depwarn flag.
+        # With --depwarn=error, this becomes an error (strict mode).
+        # This matches the behavior requested in issue #469 https://github.com/JuliaTime/TimeZones.jl/issues/469#issuecomment-2341741754.
+        Base.depwarn(
+            "The time zone \"$str\" is deprecated, using \"$link\" instead.",
+            :TimeZone
+        )
+        return TimeZone(String(link), mask)
```
rofinn, line 60: NOTE: If the `backward` file had a recursive case of a -> b -> c, then this would still error. The header comments say that's been avoided due to other parsers not handling it.

src/types/timezonecache.jl
```diff
@@ -63,10 +69,17 @@ function Base.get(body::Function, cache::TimeZoneCache, name::AbstractString)
 end
 
 # Build specific tzdata version if specified by `JULIA_TZ_VERSION`
-function _build()
-    desired_version = TZData.tzdata_version()
-    if desired_version != TZJData.TZDATA_VERSION
-        _COMPILED_DIR[] = TZData.build(desired_version, _scratch_dir())
+# Also rebuilds if the TZJFile format version doesn't match the expected version
+function _build(tzjf_version::Integer=TZJFile.tzjfile_version())
```
rofinn, line 73: Another main hack is for switching compiled file formats.

src/tzjfile/TZJFile.jl
```diff
@@ -4,7 +4,36 @@ using Dates: Dates, DateTime, Second, datetime2unix, unix2datetime
...
+julia> TZJFile.tzjfile_version()
+2
+
+# Use version 1 via environment variable
+julia> ENV["JULIA_TZJ_VERSION"] = "1"
+julia> TZJFile.tzjfile_version()
+1
+```
+"""
+function tzjfile_version()
```
rofinn, line 30: Again, a handy feature for switching compiled file versions, but it could be removed.

rofinn:
I'm not surprised about the TZJData system image errors, but the allocation test differences on 1.6 and nightly seem odd.

src/types/timezone.jl
omus: My initial thought is to encode the link information into the `Class`

rofinn: Yeah, that's where I looked first, but I found a few issues with that:

1. We'd still have the same `isbits` issue, but just moving to a nested type.
2. We'd be adding bloat to the very small `Class` type. It just wraps a `UInt8` right now.
3. I don't see the current bitwise operations really making sense with it.

PR #509, opened 2026-04-13
Ignore dotfiles when walking tz directories

jd-foster:
Encountered the following error:
```
LoadError: ArgumentError: Magic file identifier "TZjf" not found.
  Stacktrace:
    [1] read_signature
      @ ~/.julia/dev/TimeZones/src/tzjfile/read.jl:15 [inlined]
    [2] read(io::IOStream)
      @ TimeZones.TZJFile ~/.julia/dev/TimeZones/src/tzjfile/read.jl:8
    [3] open(::typeof(TimeZones.TZJFile.read), ::String, ::Vararg{String}; kwargs::@Kwargs{})
      @ Base ./io.jl:410
    [4] open
      @ ./io.jl:407 [inlined]
    [5] #reload!##0
      @ ~/.julia/dev/TimeZones/src/types/timezonecache.jl:31 [inlined]
    [6] walk_tz_dir(f::TimeZones.var"#reload!##0#reload!##1"{TimeZones.TimeZoneCache}, dir::String)
      @ TimeZones ~/.julia/dev/TimeZones/src/utils.jl:126
```

[This observation](https://github.com/apache/arrow-julia/issues/510#issuecomment-2488695003) yielded the answer, that typically macOS users see the error from .DS_Store files being included in the caching.

Issue #473, opened 2024-09-06
`TimeZones.build` doesn't rebuild compiled files upon failure


omus:
While working on #471 I noticed at after I introduced the fix in #472 I couldn't use `TimeZones.build("2024b")` to re-build without first deleting the `compiled/tzjf/v1/2024b` scratch folder first:
```julia
julia> TimeZones.build("2024b")
ERROR: Cache remains empty after loading
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] reload!
   @ ~/.julia/dev/TimeZones/src/types/timezonecache.jl:40 [inlined]
 [3] _reload_tz_cache
   @ ~/.julia/dev/TimeZones/src/types/timezonecache.jl:73 [inlined]
 [4] build(version::String; force::Bool)
   @ TimeZones ~/.julia/dev/TimeZones/src/build.jl:16
 [5] build(version::String)
   @ TimeZones ~/.julia/dev/TimeZones/src/build.jl:11
 [6] top-level scope
   @ REPL[2]:1
```
Seems like we should check that the destination directory both exists and is not empty when skipping compilation.

omus:
Was fixed by https://github.com/JuliaTime/TimeZones.jl/pull/474

PR #510, opened 2026-09-15
Ignore dotfiles when deciding whether tzdata is already compiled

youdie006:
#509 taught `walk_tz_dir` to skip dotfiles, but `build`'s "is this already compiled?" check reads the same directory with a bare `readdir`:

```julia
# src/tzdata/build.jl:56
if !isdir(compiled_dir) || isempty(readdir(compiled_dir))
```

So a macOS `.DS_Store` in the compiled directory counts as compiled data. Measured on `master`, with only `.DS_Store` in `compiled/tzjf/v1/2016j`:

```
pristine : build() logs nothing, directory afterwards is [".DS_Store"]   - zero tz files
fixed    : "Compiling tzdata 2016j region data", directory is [".DS_Store", "GMT", "UTC"]
```

With no tz files, `reload!` throws `Cache remains empty after loading` - the error #473 reported, and the one #474's emptiness check was added to prevent. Re-running `TimeZones.build()` cannot repair it, because the dotfile keeps the directory non-empty forever; #473's manual workaround of deleting the scratch folder is the only way out.

#509 made this more reachable rather than less: before it, the walker threw `Magic file identifier "TZjf" not found` on the dotfile, so the problem announced itself.

`build` already filters the tz **source** directory against `REGIONS` twelve lines earlier (`:53`), and `deprecated.jl:37` does the same. The compiled directory was the one place a raw `readdir` result was trusted as tz data.

The second hunk applies the same guard to `compile(tz_source_dir, dest_dir)` at `compile.jl:716`, the only other bare `readdir` of a tz directory. That path is not covered by the new test: reaching it needs a dotfile with real binary content in a source directory passed directly to `TZData.compile`, and how it fails then depends on the bytes (I saw both an `Unhandled line found with type: ... Bud1` warning and a `BoundsError` from the `split(line, r"\s+"; limit=3)` destructure). Say the word if you would rather I split that hunk out.

<details>
<summary>Verification</summary>

`Pkg.test()` on `julia:1.11` with `JULIA_TZ_VERSION=2016j`, as `CI.yml` sets it:

| | Pass | Fail | Broken | Total |
|---|---|---|---|---|
| this branch | 32795 | 0 | 4 | 32799 |
| `build.jl` reverted, test kept | 32794 | 1 | 4 | 32799 |
| bound widened to "always recompile" | 32794 | 1 | 4 | 32799 |

The revert fails the first row (`"UTC" in [".DS_Store"]`); always-recompiling fails the second (`!("UTC" in [".DS_Store", "GMT", "SENTINEL", "UTC"])`), which is in the test specifically to keep #474's caching behaviour pinned. Different rows in each direction.

The existing `build` testset passes on `master` because it runs `TZData.cleanup` first, so `compiled_dir` does not exist and the `!isdir` branch short-circuits - the `readdir` half is never evaluated. That is why this was not caught.

The test needs no network: the archive file is touched and the source directory is seeded from the repo's own `deps/tzsource_custom/utc`.

</details>

Prepared with AI assistance (Claude); I reproduced the skipped build and the empty compiled directory myself, and verified both mutation directions.


youdie006:
CI note: the two red legs on this PR are pre-existing on master, not caused by this change.

`Julia nightly - ubuntu-latest - x64` and `Julia 1.13.0 - ubuntu-latest - x86` are the same two jobs that fail in the scheduled master runs at `d743cd6a` with no PR involved, and they have failed there every day for at least five consecutive runs (34941760535, 34819149247, 34744910397, 34679552964, 34572959811). Every other leg on this PR is green.

I have not investigated the cause of those two, since it is outside the scope of this change.

